PR #154, opened 2021-04-07
Disable cardfield autofocus

### LINKED ISSUE #135: [Android] CardField focuses itself when view is loaded
**Describe the bug**
On Android, when you load a view that has the CardField component in it, it focuses itself, which is not desired. 

**To Reproduce**
1. Go to "Card payments using webhooks" screen
2. See the cardfield being focused

**Expected behavior**
CardField should not automatically focus itself

cc @michelleb-stripe @michaelbayday

arekkubaczkowski:
resolves #135


babyrusa:
Hi @thorsten-stripe , I have stripe-react-native 0.1.5 but this is still happening. Once I touched the CardField, it would never go blur which resulted in me not able to click on anything else on in the View. Is there anyway I could resolve
react-native : 0.64.2

PR #197, opened 2021-05-11
fix: account for library rename in plugin

cruzach:
renaming the library from `stripe-react-native` to `@stripe/stripe-react-native` broke the config plugin's behavior

tested these changes locally, and both `AndroidManifest.xml` and `.entitlements` files are updated when running `expo prebuild` with the new app.json configuration

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,17 +38,19 @@ npm install @stripe/stripe-react-native
 If you're using Expo, add:
 
 ```json
-expo: {
-  ...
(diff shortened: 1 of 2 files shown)


thorsten-stripe:
Thanks @cruzach. How quickly do you need this to be deployed? Can it wait a couple of days?

cruzach:
Yeah it can wait a couple days, no rush

PR #207, opened 2021-05-13
fix: handle different pjson output

### LINKED ISSUE #200: 'JSON value of type NSMutableDictionary cannot be converted to NSString' related to appInfo
**Describe the bug**
When trying to use the SDK in an iOS app I ran into this error : 
```
JSON value '{
    type = git;
    url = "git+https://github.com/stripe/stripe-react-native.git";
}' of type NSMutableDictionary cannot be converted to NSString
 
+[RCTConvert NSString:]
    RCTConvert.m:59
$s19stripe_react_native9StripeSdkC10initialise6params8resolver8rejecterySo12NSDictionaryC_yypSgcySSSg_AKs5Error_pSgtctF
```

Tracing through this, what happens is : 
we try to set `appInfo` to [identify the plugin](https://stripe.com/docs/building-plugins#setappinfo) in 
https://github.com/stripe/stripe-react-native/blob/b831a18d9ccc5f2b68db2d7c7d94bcf1b1af0f86/ios/StripeSdk.swift#L52-L55 

that code expects `appInfo["url"]` to be a String but it's actually an object as seen in the error message.

The value for this comes from here : 
https://github.com/stripe/stripe-react-native/blob/b831a18d9ccc5f2b68db2d7c7d94bcf1b1af0f86/src/components/StripeProvider.tsx#L28-L33 
The URL field for the appInfo (` url: pjson.repository,` ) is coming from the package.json file. 

Although in the Github project [it looks like that URL is a string](https://github.com/stripe/stripe-react-native/blob/b831a18d9ccc5f2b68db2d7c7d94bcf1b1af0f86/package.json#L43), what I find is that when I actually install the SDK through npm, the package.json that is served in the dependency package has this object instead.
It's a valid format for that file type : https://docs.npmjs.com/cli/v7/configuring-npm/package-json#repository 

```
~  npm install @stripe/stripe-react-native
....
~  cat  node_modules/@stripe/stripe-react-native/package.json | grep -A 4  repository
  "repository": {
    "type": "git",
    "url": "git+https://github.com/stripe/stripe-react-native.git"
  },
  "scripts": {
```

I imagine that the file gets rebuilt on `npm`s side or something similiar.

We should be more robust in how we parse that file in StripeProvider.tsx  — at the very least we should probably do `repository.url || repository` but we might want to make that code overall more aware of the package.json format than it is at the moment.

arekkubaczkowski:
resolves #200

### DIFF
--- a/src/components/StripeProvider.tsx
+++ b/src/components/StripeProvider.tsx
@@ -25,9 +25,14 @@ export interface Props {
   setUrlSchemeOnAndroid?: boolean;
 }
 
+const repository: any = pjson.repository;
+
 const appInfo: AppInfo = {
   name: shouldAttributeExpo() ? `${pjson.name}/expo` : pjson.name,
-  url: pjson.repository,
+  // package.json output installed via npm is a bit different than from yarn
+  // the repository field can be an object or string
+  // for more context: https://github.com/stripe/stripe-react-native/issues/200
+  url: repository.url || repository,
   version: pjson.version,
   partnerId: shouldAttributeExpo() ? EXPO_PARTNER_ID : undefined,
 };


thorsten-stripe (approved):
LGTM, thanks

PR #208, opened 2021-05-13
fix: use npmignore

### LINKED ISSUE #204: Remove example from published package
**Describe the bug**
I don't think the example should be included within the node_module. Perhaps this should be excluded when publishing? 

**Expected behavior**
Examples should be tested when cloning the repo not via the package installed within node_modules.

**Screenshots**
image

**Additional context**

Context is I am trying to patch the package with patch-package this folder is giving me issues.

arekkubaczkowski:
resolves #204 
as yarn doesn't support negations in package.json we need to use .npmignore for excluding certain files

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,4 @@
+lib/typescript/example
+**/__tests__
+**/__fixtures__
+**/__mocks__
--- a/package.json
+++ b/package.json
@@ -14,11 +14,7 @@
     "android",
     "ios",
     "cpp",
-    "stripe-react-native.podspec",
-    "!lib/typescript/example",
-    "!**/__tests__",
-    "!**/__fixtures__",
-    "!**/__mocks__"
+    "stripe-react-native.podspec"
   ],
   "scripts": {
     "test": "jest",


.npmignore
```diff
@@ -0,0 +1,4 @@
+lib/typescript/example
+**/__tests__
+**/__fixtures__
+**/__mocks__
```
michaelbayday, line 4: I think **/example should be npm ignored as well.

in the package's current state it is publishing the example file and all of the pod dependencies etc.

thorsten-stripe, line 4:
```suggestion
**/__mocks__
**/example
**/.circleci
**/.github
**/scripts
```

thorsten-stripe (changes requested):
@arekkubaczkowski should we add the appium files also?

And do we need to add everything from the `.gitignore` file also, or will that be added automatically?

.npmignore
arekkubaczkowski: I agree, and this is why we have defined whitelist of files inside of package.json
https://github.com/stripe/stripe-react-native/blob/fix/matching-files-pattern/package.json#L11
combining this with couple of exceptions from .npmignore w get exactly what we need.
@thorsten-stripe @michaelbayday you can run `npm pack` and check what is inside of package.

thorsten-stripe (approved):
Okay, I now understand that it didn't work because we had negations in the allow list. By moving the negations to the disallow list (`.npmignore`) the allow list now works as expected 👍

.npmignore
thorsten-stripe: Okay, I now understand that it didn't work because we had negations in the allow list. By moving the negations to the disallow list (.npmignore) the allow list now works as expected 👍

PR #213, opened 2021-05-13
fix: expose config-plugin and add blank swift file

cruzach:
When I tested this library via installing by `npm`, I realized the `app.plugin.js` was not being distributed with the package 😱 

Also added another plugin that will add a blank, no-op Swift file to the xcode project to avoid the "undefined symbols for architecture..." errors described here- https://github.com/stripe/stripe-react-native#troubleshooting - although it is not clear to me yet what the best solution for `LIBRARY_SEARCH_PATHS` is. 

However, it doesn't look like you need to clear out **all** LIBRARY_SEARCH_PATHS. In particular, it looks like `"\"$(TOOLCHAIN_DIR)/usr/lib/swift-5.0/$(PLATFORM_NAME)\""` causes the problem. (I'm not sure what side effects removing that will cause, so I'm not doing that in the plugin yet)


src/plugin/withStripe.ts
```diff
@@ -72,6 +74,22 @@ export function setApplePayEntitlement(
   return entitlements;
 }
 
+/**
+ * Add a blank Swift file to the Xcode project for Swift compatibility.
+ */
+export const withNoopSwiftFile: ConfigPlugin = (config) => {
```
thorsten-stripe, line 80: @cruzach does this add the needed bridging header?

cruzach: the bridging header is added by default by Expo 👍

thorsten-stripe (approved):

PR #226, opened 2021-05-17
card field autofocus

### LINKED ISSUE #199: AutoFocus/focus function on CardField
An AutoFocus/focus function on the CardField component could be really cool.

It will save some clicks to my user and by abtesting we saw that the conversion rate was better in the past

arekkubaczkowski:
resolves #199


thorsten-stripe:
`autofocus={true}` does not seem to bring up the keyboard though, at least on the simulators for me both on Android and iOS. @arekkubaczkowski any idea how we can force the keyboard to open?

thorsten-stripe (approved):
LGTM, thanks

PR #228, opened 2021-05-17
handle apple pay errors

### LINKED ISSUE #148: Displaying errors in apple pay sheet.
I would like to show errors if payment parameters are not valid in the apple pay sheet.

Would be good to display a general error message and disallow the pay button if the shipping address isn't valid for example, or if possible show validation for particular shipping address fields, like zipcode or state.

**Additional context**
When "onShippingContact" event is triggered you usually need to run some logic where you calculate taxes or shipping method prices. There can be a situation where the user inputs an invalid shipping address and I don't see a way to point out that error to the user in the apple pay sheet.

arekkubaczkowski:
resolve #148


thorsten-stripe:
@arekkubaczkowski the approach looks good to me, and I works fine for me within the `onShippingContactSelected` handler, however it does not work for me in the `onShippingMethodSelected` handler.

ios/StripeSdk.swift
```diff
@@ -256,8 +256,12 @@ class StripeSdk: RCTEventEmitter, STPApplePayContextDelegate, STPBankSelectionVi
                 paymentSummaryItems.append(PKPaymentSummaryItem(label: label, amount: amount, type: type))
             }
         }
+        let shippingAddressErrors = Mappers.mapAddressFields(errorAddressFields).map {
+            PKPaymentRequest.paymentShippingAddressInvalidError(withKey: $0, localizedDescription: $0 + " field error")
```
thorsten-stripe, line 260: I think we should allow folks to overwrite with their own error message. Maybe we can do

```js
handler: (
      summaryItems: ApplePay.CartSummaryItem[],
      errorAddressFields?: Array<{ field: ApplePay.AddressFields; message?: string; }>
    ) => Promise<{
      error?: StripeError<ApplePayError>;
    }>
```

Wdyt?

thorsten-stripe (changes requested):

ios/StripeSdk.swift
arekkubaczkowski: done :+1:

thorsten-stripe (approved):
LGTM, thanks

PR #232, opened 2021-05-18
Add example folder to .npmignore

gran33:
Hey, 
Thanks for this awesome wrapping library for Stripe. And welcome to React Native world.
I just started using your lib and figure out that you publish to NPM also your `example` folder, means we ll get in when we do `npm install`, and It includes big `Pod` library.

So, I added to your `.npmignore` to ignore this unnecessary example folder.

Thanks,
Ran

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -2,3 +2,4 @@ lib/typescript/example
 **/__tests__
 **/__fixtures__
 **/__mocks__
+example/


arekkubaczkowski:
@gran33 besides .npmignore we also use white list in package.json 
https://github.com/stripe/stripe-react-native/blob/master/package.json#L11 so we don't need to specify this in npmignore. I cannot reproduce this issue, it has been fixed here:

https://github.com/stripe/stripe-react-native/issues/204

I am closing this PR

PR #245, opened 2021-05-20
fix: clear card params when are not valid

### LINKED ISSUE #237: [Android] With `CardField` when the card is completed and then uncomplete the flag does not change
**Describe the bug**
With `CardField` I enter a valid card and then remove some data or change the card number to something invalid. In the card details returned by the `CardField` component

**To Reproduce**
- Use the sample app
- Go to the WebhookPayment screen
- Enter valid data
- remove some data or change the card number to something invalid
- validate the form, the payment is ok

**Expected behavior**
- When we validate the form with incorrect data the payment should not be ok

- Android simulator

arekkubaczkowski:
resolves #237


thorsten-stripe (changes requested):
@arekkubaczkowski could you include the fix for Android in this PR also?

arekkubaczkowski:
> @arekkubaczkowski could you include the fix for Android in this PR also?

Oh right, I thought it happens only on iOS for some reason

thorsten-stripe (approved):
LGTM, thanks

PR #252, opened 2021-05-21
fix: make data param nullable for onActivityResult on android

experiment322:
It seems that the code using this param handled the possibility for `data` to be `null`, but it wasn't declared as a nullable type.

Encountered a crash while playing with another library which also spawned some activity which can return no data (ie `data = null`), but even so, it seemed that this param should've been declared with a nullable type.

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
+++ b/android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
@@ -43,7 +43,7 @@ class StripeSdkModule(reactContext: ReactApplicationContext, cardFieldManager: S
   private var confirmPaymentClientSecret: String? = null
 
   private val mActivityEventListener = object : BaseActivityEventListener() {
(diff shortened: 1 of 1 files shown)


thorsten-stripe:
I think this might have been resolved by https://github.com/stripe/stripe-react-native/pull/255 @arekkubaczkowski can you take a look?

arekkubaczkowski:
> I think this might have been resolved by #255 @arekkubaczkowski can you take a look?

I think it's unrelated, I am fine to merge this PR

thorsten-stripe:
Thanks @experiment322 . Could you please sign the CLA?

arekkubaczkowski (approved):

experiment322:
@thorsten-stripe, I've already signed it. I don't know why it's reported that it isn't signed. Does it need some kind of approval from someone?

thorsten-stripe:
@experiment322 looks like maybe an issue with email address difference, see message above: https://github.com/stripe/stripe-react-native/pull/252#issuecomment-845939162

experiment322:
Fixed

PR #255, opened 2021-05-22
fix: crash on onActivityResult if Stripe isn't initalized

cruzach:
Apps can crash if they haven't initialized stripe, and run code that triggers `BaseActivityEventListener.onActivityResult`, so the listener's content should be wrapped inside of `if (::stripe.isInitialized) {`


thorsten-stripe (approved):
LGTM, thanks

PR #260, opened 2021-05-24
dismissing keyboard on tap outside CardField

### LINKED ISSUE #222: Unable to dismiss the keyboard using `CardField`
**Describe the bug**
There is no way to dismiss the keyboard (by pressing "Done", or taping away) using the `CardField` component from what I can tell. This is the same behavior in the example project as well.

**To Reproduce**
1. Use the `CardField` component on a screen
2. Tap into the field to open the keyboard

**Expected behavior**
A user can easily dismiss the keyboard.

arekkubaczkowski:
resolves #222


example/src/components/PaymentScreen.tsx
```diff
@@ -26,18 +27,21 @@ const PaymentScreen: React.FC<Props> = ({ paymentMethod, children }) => {
...
-    <ScrollView
-      accessibilityLabel="payment-screen"
-      keyboardShouldPersistTaps="always"
-      style={styles.container}
-    >
-      {children}
-      {/* eslint-disable-next-line react-native/no-inline-styles */}
-      <Text style={{ opacity: 0 }}>appium fix</Text>
-    </ScrollView>
+    <StripeContainer keyboardShouldPersistTaps={false}>
```
thorsten-stripe, line 34: `false, deprecated, use 'never' instead` - https://reactnative.dev/docs/scrollview#keyboardshouldpersisttaps

thorsten-stripe (changes requested):
I think `false` has been deprecated as a value for `keyboardShouldPersistTaps`

thorsten-stripe:
I found there's a bit of a glitch on Android when switching between the email and card input fields:
https://user-images.githubusercontent.com/23213994/119335309-c9c6f700-bcbe-11eb-99bd-631ec771b416.mp4

example/src/components/PaymentScreen.tsx
arekkubaczkowski: it's a custom property for StripeContainer, just the same prop name as for ScrollView to be more consistent but we don't need to handle the same cases as scrollView does, thats why I set boolean instead of `always | never | handled`

thorsten-stripe: Okay, sounds good, thanks!

thorsten-stripe (approved):
LGTM, thanks

PR #262, opened 2021-05-24
filter activity results by requests codes

arekkubaczkowski:


android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
```diff
@@ -21,6 +21,21 @@ import kotlinx.coroutines.runBlocking
...
 
+  companion object {
+    private const val PAYMENT_REQUEST_CODE = 50000
+    private const val SETUP_REQUEST_CODE = 50001
+    private const val SOURCE_REQUEST_CODE = 50002
+    private const val ADD_PAYMENT_METHOD_REQUEST_CODE = 6001
+    private const val PAYMENT_SHEET_REQUEST_CODE = 6003
+    private const val PAYMENT_SHEET_OPTIONS_REQUEST_CODE = 6004
+
+    private val VALID_REQUEST_CODES = setOf(
```
thorsten-stripe, line 32: Are we concerned about these ever changing, no, right?

arekkubaczkowski: ideally would be make them public in android-sdk to be able use it but for now it's marked as a private

cruzach (approved):

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
thorsten-stripe: cc @michelleb-stripe ^

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
```diff
@@ -39,12 +52,16 @@ class StripeSdkModule(reactContext: ReactApplicationContext, cardFieldManager: S
...
 
   private val mActivityEventListener = object : BaseActivityEventListener() {
     override fun onActivityResult(activity: Activity, requestCode: Int, resultCode: Int, data: Intent?) {
-      if (::stripe.isInitialized) {
+      if (!::stripe.isInitialized) {
+        return
+      }
+      paymentSheetFragment?.activity?.activityResultRegistry?.dispatchResult(requestCode, resultCode, data)
+
+      if (VALID_REQUEST_CODES.contains(requestCode)) {
```
michelleb-stripe, line 64: Why do you need to filter?

cruzach: to make sure that the relevant `startActivityForResult` came from the expected source (stripe), and not a separate library that also uses activity triggers

michelleb-stripe: What prevents you from forwarding the onActivityResult to more than one function?  This function doesn't return if it has handled the activity result, and it doesn't include an else statement in the case where it is not one of the expected values.

thorsten-stripe (changes requested):
https://github.com/stripe/stripe-react-native/pull/262#discussion_r638280477

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
arekkubaczkowski: @michelleb-stripe sorry but I don't get what u mean, why should we return something there?

michelleb-stripe: You don't need to return something.  

I am not understanding why you need to filter at this point.  It is my understanding that the functions onActivityResult do their own filtering based on the result code, because it is each of the listeners that know the value of the result code they started.

If you call onActivityResult and the result is not for that handler it should ignore it.   In this way you could call multiple onActivityResult, and only the one listening for the result code would do the work.

thorsten-stripe:
Closing this out as discussed as `stripe-android` already takes care of this.

PR #284, opened 2021-05-28
remove build tools version

dulmandakh:
Setting buildToolsVersion is no longer recommended, it'll default to Android Gradle Plugin defaults. It makes DX worse by requiring multiple versions of build tools by various dependencies require various versions.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -28,7 +28,6 @@ def getExtOrIntegerDefault(name) {
 
 android {
   compileSdkVersion getExtOrIntegerDefault('compileSdkVersion')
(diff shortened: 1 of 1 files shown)


itsBhuviRana:
in my build gradle file there is no line like this.

thorsten-stripe:
Thanks for flagging this @dulmandakh. Any chance you have a link where I can read up on this? I've been struggling to find any docs.

dulmandakh:
https://developer.android.com/studio/releases/gradle-plugin?hl=tr&buildsystem=cmake#3-1-0

thorsten-stripe (approved):
Brilliant, thanks a mill!

PR #288, opened 2021-05-31
Added Expo Go update to troubleshooting

paulasjes-stripe:
Been seeing a lot of confusion around the error `TypeError: null is not an object (evaluating '_NativeStripeSdk.default.initialise')` on Android when using Expo. Adding the solution to the troubleshooting section as it's not an obvious error.


thorsten-stripe (approved):
LGTM, thanks!

PR #305, opened 2021-06-02
Fix UPI payment method on iOS

jamesblasco:
When creating a payment method, the `vpa` parameter from the `Upi` payment methd is missing in iOS
 [Mappers.swift#L548](https://github.com/stripe/stripe-react-native/blob/967bdbe78abe46c772e0ad68d084d83769075f6b/ios/Mappers.swift#L548)

React Native: 
https://github.com/stripe/stripe-react-native/blob/967bdbe78abe46c772e0ad68d084d83769075f6b/src/types/PaymentMethods.ts#L199-L201

Android:
https://github.com/stripe/stripe-react-native/blob/967bdbe78abe46c772e0ad68d084d83769075f6b/android/src/main/java/com/reactnativestripesdk/Mappers.kt#L298

### DIFF
--- a/ios/Mappers.swift
+++ b/ios/Mappers.swift
@@ -546,6 +546,9 @@ class Mappers {
             "Sofort": [
                 "country": paymentMethod.sofort?.country
             ],
+            "Upi": [
+                "vpa": paymentMethod.upi?.vpa
+            ],
         ]
         return method
     }


thorsten-stripe (approved):
LGTM, thanks

PR #313, opened 2021-06-06
docs: Added steps to IOS troubleshooting

TwistedTamarin:
Added extra steps to resolve IOS swift linker errors


Aawaraishq00007 (approved):
Mohammed

arekkubaczkowski:
@Aawaraishq00007 what is behind this change? I don't think it's relevant whether user has configured bridging header before

PR #320, opened 2021-06-08
chore: export confirmPayment directly

### LINKED ISSUE #318: How to accept on-session payments with saved payment method.
Hi, I am migrating my App from Charges API to PaymentIntents and I think I am missing something very basic here but Is there any documentation about how to accept on-session payments with saved cards/payment methods. I checked Stripe's documentation about saving and reusing a payment method but it only explains about off-session payments and steps one should follow to start a recovery flow in case the off-session payment gets declined.

In my use case, and I think this is a very common use case, customer adds a card and pay for their order. I am saving this card for future use with **setup_future_usage : on-session**. Everything working perfectly fine when I try to add a new card and pay. The payment is processed perfectly, the authentication screen pops up wherever required, the payment method gets saved to the customer and so on.. But I can't understand how should I use this saved payment method later.

Any help will be highly appreciated.

thorsten-stripe:
Resolves #318

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -263,7 +263,7 @@ PODS:
     - React-Core
   - Stripe (21.6.0):
     - Stripe/Stripe3DS2 (= 21.6.0)
-  - stripe-react-native (0.1.2):
+  - stripe-react-native (0.1.4):
     - React
     - Stripe (~> 21.6.0)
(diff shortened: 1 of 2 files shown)


thorsten-stripe:
Closing this in favor of https://github.com/stripe/stripe-react-native/pull/354

PR #337, opened 2021-06-14
add imperative cardField methods (focus, blur, clear)

arekkubaczkowski:
- expose CardField methods (focus, blur, clear)
- inject CardField reference to react-native InputState so the state of card field input can be handled via RN components like ScrollView

There is no need to use `StripeContainer` anymore since it's registered to RN InputState.

auto keyboard state example:
```
<ScrollView
      keyboardShouldPersistTaps="never" // <-- add this line 
>
      {children}
 </ScrollView>
```

example of methods usage:

```
// App.js

const ref = useRef(null)

const handleFocus = () => {
    ref.current.focus()
}

const handleBlur = () => {
    ref.current.blur()
}

const handleClear = () => {
    ref.current.clear()
}

return (
 <CardField
        ref={ref}
  />
)
```

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/StripeSdkCardView.kt
+++ b/android/src/main/java/com/reactnativestripesdk/StripeSdkCardView.kt
@@ -52,6 +52,28 @@ class StripeSdkCardView(private val context: ThemedReactContext) : FrameLayout(c
     }
   }
 
+  fun requestFocusFromJS() {
+    val binding = CardInputWidgetBinding.bind(mCardWidget)
+    binding.cardNumberEditText.requestFocus()
+    binding.cardNumberEditText.showSoftKeyboard()
+  }
+
+  fun requestBlurFromJS() {
+    val binding = CardInputWidgetBinding.bind(mCardWidget)
+    binding.cardNumberEditText.hideSoftKeyboard()
+    binding.cardNumberEditText.clearFocus()
+    binding.container.requestFocus()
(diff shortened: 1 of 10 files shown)


thorsten-stripe (approved):
LGTM, thanks

aganov:
@arekkubaczkowski not sure if you spotted but I've added `onBlur` support from your prev. PR to the patch. Which as expected only works on iOS. Just FIY in case you've missed it and apparently landed here too...

```swift
func paymentCardTextFieldDidEndEditing(_ textField: STPPaymentCardTextField) {
    onFocusChange?(["focusedField": NSNull()])
}
```

apperside:
If someone needs this features in the meanwhile that stripe makes a new release, you can install the library from master

`yarn add stripe/stripe-react-native#master`

or 

`npm i stripe/stripe-react-native#master`

after installing it, go inside node_modules/@stripe/stripe-react-native , delete the example folder and run `yarn` or `npm i`, it should build the library automatically.

At the end of the process, you should have a new lib folder and you should be able to use the library (it worked for me, at least)

thanhcuong1990:
Please release this PR soon. Thanks!

thorsten-stripe:
We'll be releasing a new version later this week. If you need it earlier you can build and install the master branch locally: https://github.com/stripe/stripe-react-native/blob/master/CONTRIBUTING.md#install-library-as-local-repository

PR #354, opened 2021-06-18
chore: rename confirmPaymentMethod to confirmPayment

thorsten-stripe:
It's creating some confusion that `functions` is exporting `confirmPaymentMethod` whereas `useStripe` is exporting `confirmPayment`: https://github.com/stripe/stripe-react-native/issues/318

This PR is unifying it and getting rid of `confirmPaymentMethod` across the board as it also doesn't make too much sense since one does confirm the payment intent, not the method.


arekkubaczkowski:
LGTM

arekkubaczkowski (approved):

Issue #315, opened 2021-06-07
Issues with Payment Sheet not appearing


trex-quo:
**Describe the bug**
Calling `initPaymentSheet()` and `presentPaymentSheet()` in immediate succession does not bring up the Payment Sheet. Adding a short sleep() statement (~1000ms) in between the two calls resolves the issue.  

Additionally, calling `presentPaymentSheet()` from a modal crashes the app. 

**To Reproduce**
```
  const onSubscribePress = async () => {
    let {error} = await initPaymentSheet({
      paymentIntentClientSecret: clientSecret,
      googlePay: true,
      merchantCountryCode: 'US',
      testEnv: true,
    });

    if (error) {
        //handle error
    }

    //without this line, the payment sheet will not open
    await sleep(1000);

    try {
      console.log('presenting payment sheet');
      ({error} = await presentPaymentSheet({
        clientSecret: clientSecret,
      }));
    } catch (e) {
      console.log({e});
      setLoading(false);
      return;
    }
  };
```

**Expected behavior**
It seems logical to choose not make a backend call until the client taps the payment button, meaning that the `init` and `present` call would fall under the same function. Sleeping is not an ideal solution to this problem, so it would be nice if the payment sheet was presented properly without having to do this. 

It would also be helpful if we could present the payment sheet from within a modal. As is, we need to redesign the app specifically for Android in order to present the payment sheet outside of our subscribe screen, which is contained in a modal. 

**Desktop (please complete the following information):**
 - OS: macOS Big Sur
 - Version 11.2.3

**Smartphone (please complete the following information):**
 - Device: iPhone 11, Galaxy S10
 - @stripe/stripe-react-native v0.1.4

tayzlor:
Just a quick comment to mention I'm experiencing the same issue when you try and call both the initialise function and the present function off the back of a button press, and can confirm that a small delay does resolve the issue. 

Currently doing some further troubleshooting to understand why and will provide any more info when I have it!

tayzlor:
So looks like this may be occurring because initPaymentSheet uses fragment commit() [https://developer.android.com/reference/android/app/FragmentTransaction.html#commit()] to update the paymentSheetFragment var in  https://github.com/stripe/stripe-react-native/blob/89042f4d54a7b525cea4a245ce8a7fc8ebf8de03/android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt#L273-L275

[...]

Issue #290, opened 2021-05-31
[ios] `presentPaymentSheet` will crash the app once is being called. 


paschalidi:
**Describe the bug**
Have been following [this](https://stripe.com/docs/payments/accept-a-payment?platform=react-native&lang=node&ui=payment-sheet) tutorial and even though most of the things went pretty well, the app would crash when I am calling the `presentPaymentSheet` with the `clientSecret` returned from my backend. There isn't any log I could find that would help me debug the issue. 

Here is my component, I hope pasting the code helps a little bit more. I am using firebase functions as my backend.
```
import React, {useState, useEffect} from 'react';
import {StyleSheet, Alert} from 'react-native';
import {Text, useTheme} from 'react-native-paper';
import {
  useStripe,
} from '@stripe/stripe-react-native';
import {useProfile} from './ProfileProvider';
import {SafeAreaView} from 'react-native-safe-area-context';
import {FormButton} from '../../components/FormButton';

export default function CardDetailsScreen({navigation}) {
  const {styles} = useStyles();
  const {
    profile: {stripeId},
  } = useProfile();
  const {initPaymentSheet, presentPaymentSheet} = useStripe();
  const [paymentSheetEnabled, setPaymentSheetEnabled] = useState(false);
  const [clientSecret, setClientSecret] = useState();

  const [loading, setLoading] = useState(false);

  useEffect(() => {
    const initialisePaymentSheet = async () => {
      const fetchPaymentSheetParams = async () => {
        try {
          const response = await fetch(
            'https://MY_BACKEND_ENDPOINT.cloudfunctions.net/completePaymentWithStripe',
            {
              method: 'POST',
              headers: {
                'Content-Type': 'application/json',
              },
              body: JSON.stringify({
                amount: 55 * 100,
                customerStripeId: stripeId,
              }),
            },
          );

          const {paymentIntent, ephemeralKey} = await response.json();
          setClientSecret(paymentIntent);

          return {paymentIntent, ephemeralKey};
        } catch (e) {
          console.log(e);
          debugger;
        }
      };

      const {paymentIntent, ephemeralKey} = await fetchPaymentSheetParams();
      const {error} = await initPaymentSheet({
        customerId: stripeId,
        customerEphemeralKeySecret: ephemeralKey,
        paymentIntentClientSecret: paymentIntent,
      });

      if (!error) {
        setPaymentSheetEnabled(true);
      }
    };

    stripeId && initialisePaymentSheet();
[...]

PR #366, opened 2021-06-23
Allow payment sheets within modals

jfrolich:
Currently when you try to call `presentPaymentSheet` from a modal, the app will crash. This has been reported in a few issues (#315, #290).

I am not entirely sure if this fix doesn't breaks non-modal use-cases. But at least it worked correctly within a modal for our use-case with this modification.


dearlove0007 (commented):
Ok

arekkubaczkowski (approved):

Issue #371, opened 2021-06-25
confirmSetupIntent not working as expected


fdelacruzsoto:
**Describe the bug**

Currently when I try to call `confirmSetupIntent` I get the following error:

```
{
  "stripeErrorCode": null,
  "declineCode": null,
  "localizedMessage": "Card details not complete",
  "message": "Card details not complete",
  "type": null,
  "code": "Failed"
}
```
Card details are always complete

**To Reproduce**

This is my current code for the component 

```
    <StripeProvider publishableKey="my test key">
      <CardField
        postalCodeEnabled={true}
        placeholder={{
          number: '4242 4242 4242 4242',
          expiration: '10/24',
          cvc: '123',
          postalCode: 'C.P.',
        }}
        cardStyle={{
          backgroundColor: '#FFFFFF',
          textColor: '#000000',
          borderWidth: 1,
          borderColor: COLOR.gray_100,
        }}
        style={{
          width: '100%',
          height: 50,
          marginVertical: SIZES.MD,
          borderRadius: SIZES.MD,
          ...STYLE_SHADOW,
        }}
        onCardChange={inputCard => {
          setCardDetails(inputCard);
        }}
        onFocus={focusedField => {
          setInputField(
            focusedField ? INPUT_FIELD[focusedField.toString()] : '',
          );
        }}
      />
      <Text>{inputField}</Text>
      <Button
        disabled={!cardDetails?.complete}
        onPress={submit}
        style={{
          marginTop: SIZES.MD,
          borderRadius: SIZES.MD,
          ...STYLE_SHADOW,
        }}>
        Guardar
      </Button>
    </StripeProvider>
```

And this is the submit function 

```
  const submit = async () => {
    setLoading(true);
    try {
      await createOrUpdateClient();
      await updateProfile({
        variables: {
          input: {
            card: {
              last_digits: cardDetails?.last4,
              expiry_date: `${cardDetails?.expiryMonth}/${cardDetails?.expiryYear}`,
            },
          },
        },
      });
      const setupIntentResponse = await createSetupIntent();
      const billingDetails: PaymentMethods.BillingDetails = {
        email,
      };
      const { setupIntent, error } = await confirmSetupIntent(
        setupIntentResponse.data?.stripeSetUpIntent?.client_secret as string,
        {
          type: 'Card',
          billingDetails,
        },
      );
      console.log('LOG: error ', JSON.stringify(error, null, 2));
[...]

PR #379, opened 2021-06-30
fix: fix an issue where card field was set to null

fdelacruzsoto:
As described on issue #371, the card field component was returning null when doing the confirmSetup call, which caused this method to return the error saying that card details were missing, this PR aims to fix this issue.


arekkubaczkowski:
it's not a solution for your issue, this might cause unexpected behavior like access to the cardField reference once screen really has been destroyed.

PR #392, opened 2021-07-02
fix: dates discrepancies 

### LINKED ISSUE #368: Discrepancy between Date handling on iOS and Android
We noticed a discrepancy between iOS / Android regarding "created" attribute on PaymentIntent on the native side:

https://github.com/stripe/stripe-react-native/blob/master/android/src/main/java/com/reactnativestripesdk/Mappers.kt#L328 -> created is a string
https://github.com/stripe/stripe-react-native/blob/master/ios/Mappers.swift#L360 -> created is an int
https://github.com/stripe/stripe-react-native/blob/master/src/types/PaymentIntents.ts#L7 -> created is a string

I'm not sure if it has an effect on React Native itself but it does on Android. Adjusting convertToUnitTimestamp on Mappers.kt on Android to return a String would fix this issue.

arekkubaczkowski:
closes #368

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/Mappers.kt
+++ b/android/src/main/java/com/reactnativestripesdk/Mappers.kt
@@ -325,7 +325,7 @@ internal fun mapFromPaymentIntentResult(paymentIntent: PaymentIntent): WritableM
   map.putString("status", mapIntentStatus(paymentIntent.status))
   map.putString("description", paymentIntent.description)
   map.putString("receiptEmail", paymentIntent.receiptEmail)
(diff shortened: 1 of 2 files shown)


thorsten-stripe (approved):
LGTM, thanks

thorsten-stripe (changes requested):
Actually, in iOS I get `"created": "1625667070000.0",` whereas on Android I get `"created": "1625667014000",` we should align this. Actually, rather than doing the fancy math, let's just return the API value:

```
/**
       * Time at which the object was created. Measured in seconds since the Unix epoch.
       */
      created: number;
```

thorsten-stripe (approved):
LGTM, thanks

PR #394, opened 2021-07-02
fix: createToken response discrepancy

### LINKED ISSUE #344: createToken seems to return different object response according to Platform
**Describe the bug**
I'm using the `createToken` method introduced in 0.1.3 and I'm retrieving the token id (tok_xxxxx) in the response.

**To Reproduce**
Steps to reproduce the behaviour:

Hypothesis: `card` is valid Card object obtained from a stripe `CardField` component. I used a setState on the cardField handleChange to set this card object.

Using the following code :

```javascript
if (Platform.OS === 'android') {
    const { token } = await stripe.createToken(card)
} else {
    const { token: { token } } = await stripe.createToken(card)
}
```

I get the expected result I want but this implementation is surprising.

**Expected behaviour**
Response should be formatted the same on both platform.

Also, on the [documentation website](https://stripe.dev/stripe-react-native/api-reference/modules.html#createtoken), `createToken` seems to expect parameter [CreateTokenParams](https://stripe.dev/stripe-react-native/api-reference/interfaces/card.createtokenparams.html) but it does not make a lot of sense to me, I guess typing is wrong here too ? (I'm expecting the type to be like this one: https://stripe.dev/stripe-react-native/api-reference/interfaces/card.params.html)

Thanks in advance for your consideration.

@stripe/stripe-react-native : Version 0.1.3

arekkubaczkowski:
closes #344

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
+++ b/android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
@@ -350,14 +350,14 @@ class StripeSdkModule(reactContext: ReactApplicationContext, cardFieldManager: S
   fun createToken(params: ReadableMap, promise: Promise) {
     val type = getValOr(params, "type", null)?.let {
       if (it != "Card") {
-        promise.reject(CreateTokenErrorType.Failed.toString(), "$it type is not supported yet")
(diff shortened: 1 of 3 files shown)


thorsten-stripe (approved):
LGTM, thanks

PR #395, opened 2021-07-02
WIP:  resolving payment sheet initialization in proper time

### LINKED ISSUE #315: Issues with Payment Sheet not appearing

arekkubaczkowski:
closes #315

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/Constants.kt
+++ b/android/src/main/java/com/reactnativestripesdk/Constants.kt
@@ -3,4 +3,5 @@ package com.reactnativestripesdk
 var ON_PAYMENT_RESULT_ACTION = "com.reactnativestripesdk.PAYMENT_RESULT_ACTION"
 var ON_PAYMENT_OPTION_ACTION = "com.reactnativestripesdk.PAYMENT_OPTION_ACTION"
 var ON_CONFIGURE_FLOW_CONTROLLER = "com.reactnativestripesdk.CONFIGURE_FLOW_CONTROLLER_ACTION"
+var ON_INIT_PAYMENT_SHEET = "com.reactnativestripesdk.INIT_PAYMENT_SHEET"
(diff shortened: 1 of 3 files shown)


arekkubaczkowski:
@thorsten-stripe would you mind to test it on your end as well?

thorsten-stripe (approved):
Works well for me 👍

PR #403, opened 2021-07-06
Handle 'token' and 'paymentMethodId' in confirmSetupIntent

idcuesta:
## Description
This PR tries to fix issues we found when migrating from tipsi-stripe to official stripe react native library. Please see this https://github.com/stripe/stripe-react-native/issues/371#issuecomment-874022323 to get more context.

We have seen that there are these issues when confirming a setup intent:
- Typescript spec doesn't consider `token` or `paymentMethodId` as possible parameters in ConfirmSetupIntent.Params
- iOS react native implementation supports `token`, but completely ignores `paymentMethodId`
- Android react native implementation does not consider any of these parameters and crashes if called when there is no view instance available.

We have created a fork and got our flow working with small fixes to the above mentioned issues. I am creating this PR in case it is useful for anyone else having the same issues.

__Note:__ I am very new to Stripe and I am sure there will be cases I didn't considered or there might be better way of solving things, so please, feel free to comment or suggest any changes.

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/PaymentMethodCreateParamsFactory.kt
+++ b/android/src/main/java/com/reactnativestripesdk/PaymentMethodCreateParamsFactory.kt
@@ -157,9 +157,23 @@ class PaymentMethodCreateParamsFactory(private val clientSecret: String, private
 
   @Throws(PaymentMethodCreateParamsException::class)
   private fun createCardPaymentSetupParams(): ConfirmSetupIntentParams {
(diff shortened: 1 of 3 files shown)


thorsten-stripe:
@idcuesta I think there might be some confusion with how SetupIntents work. There really shouldn't be a need to use tokens or payment methods with SetupIntents. Please see https://github.com/stripe/stripe-react-native/issues/371#issuecomment-875674939

Issue #405, opened 2021-07-07
[Android] `createToken` not handling incorrect numbers


Sotatek-AnNguyen3:
**Describe the bug**
when i input bank number into cardfield number on Android it's pass , no error (ios show error number red color but not android) then when i create token with that number , my app got crash

**To Reproduce**
1. Go to number field in card field
2. input bank number
3. See error

**Expected behavior**
the bank number will change to red color - mean it's not card number

**Screenshots**
<img width="429" alt="Screen Shot 2021-07-07 at 22 40 59" src="https://user-images.githubusercontent.com/61266900/124789398-71993b00-df74-11eb-8623-b8faf78df66e.png">

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

Sotatek-AnNguyen3:
> Can you clarify the exact number you're inputting? On my end I get "Card details not complete" error when calling `createToken` with an incomplete number.

here is my test number 9704231300893972. I'm using CardField on android and input this number but got crash app event i put try catch when i call createToken 

![Screen Shot 2021-07-13 at 09 19 16](https://user-images.githubusercontent.com/61266900/125379814-66eb0580-e3bb-11eb-81c6-0f0664127528.png)

thorsten-stripe:
I'm able to reproduce this. Thanks for reporting @Sotatek-AnNguyen3

Sotatek-AnNguyen3:
@thorsten-stripe 
this problem also happen if you enter wrong year exp example : 99 , it's will make app crash . And when i input mounth and year exp then input cvc then i re-input mounth and year exp , CardField won't recognize the new value . thank you very much sir .

Sotatek-AnNguyen3:
> @thorsten-stripe
> this problem also happen if you enter wrong year exp example : 99 , it's will make app crash . other bug that when i input mounth and year exp then input cvc then i re-input mounth and year exp , CardField won't recognize the new value . thank you very much sir .

souhe:
Hi @Sotatek-AnNguyen3, I reproduced the issue and figured out what is wrong. We will have to fix the error handling for `createToken` function so it won't crash your app and you will be able to get the error in your code. After this fix you will be able to get the error message from `res.error.message` like with every other stripe function without the app crash.

Regarding the number validation (highlighting number in red) it has to be improved on the native Android SDK @thorsten-stripe

vivekmahato26:
[...]

PR #419, opened 2021-07-14
fix: better error handling for createToken - android

souhe:
Better error handling for createToken function. Now exceptions about not valid year or card number can be catched.

This is to stop issues like #405


arekkubaczkowski (approved):

arekkubaczkowski:
LGTM

PR #422, opened 2021-07-15
WIP: Wechat Pay

### LINKED ISSUE #52: WeChat Pay: Accept a payment (app to app flow)
WeChatPay is currently on private beta. Should be available for implementation by early March. Will update here with docs once ready.

arekkubaczkowski:
iOS and Android bindings implementation for WeChat Pay
closes #52 
- [x] iOS binding
- [x] android bindings

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -132,8 +132,6 @@ dependencies {
   api 'com.facebook.react:react-native:+'
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
   implementation 'com.stripe:stripe-android:17.1.0'
-  implementation 'com.stripe:stripe-wechatpay:17.1.0'
-  implementation 'com.tencent.mm.opensdk:wechat-sdk-android-without-mta:6.7.0'
   implementation 'com.google.android.material:material:1.3.0'
   implementation 'androidx.appcompat:appcompat:1.2.0'
   implementation 'androidx.legacy:legacy-support-v4:1.0.0'
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -8,7 +8,7 @@ buildscript {
   }
 
   dependencies {
-    classpath 'com.android.tools.build:gradle:4.1.2'
+    classpath 'com.android.tools.build:gradle:4.2.1'
     // noinspection DifferentKotlinGradleVersion
     classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
   }
@@ -131,7 +131,9 @@ dependencies {
   // noinspection GradleDynamicVersion
   api 'com.facebook.react:react-native:+'
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
-  implementation 'com.stripe:stripe-android:16.10.2'
+  implementation 'com.stripe:stripe-android:17.1.0'
+  implementation 'com.stripe:stripe-wechatpay:17.1.0'
+  implementation 'com.tencent.mm.opensdk:wechat-sdk-android-without-mta:6.7.0'
   implementation 'com.google.android.material:material:1.3.0'
   implementation 'androidx.appcompat:appcompat:1.2.0'
   implementation 'androidx.legacy:legacy-support-v4:1.0.0'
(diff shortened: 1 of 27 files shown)


android/build.gradle
```diff
@@ -131,7 +131,9 @@ dependencies {
   // noinspection GradleDynamicVersion
   api 'com.facebook.react:react-native:+'
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
-  implementation 'com.stripe:stripe-android:16.10.2'
+  implementation 'com.stripe:stripe-android:17.1.0'
+  implementation 'com.stripe:stripe-wechatpay:17.1.0'
+  implementation 'com.tencent.mm.opensdk:wechat-sdk-android-without-mta:6.7.0'
```
thorsten-stripe, line 136: The important thing is that `com.tencent.mm.opensdk:wechat-sdk-android-without-mta:6.7.0` is not included in the SDK gradle, only in the example project gradle. We need to make sure it is not a dependency for everyone using the SDK!

thorsten-stripe (changes requested):

android/build.gradle
ccen-stripe: as discussed offline, please move the following two dependencies to the actual example app
```
implementation 'com.stripe:stripe-wechatpay:17.1.0'
implementation 'com.tencent.mm.opensdk:wechat-sdk-android-without-mta:6.7.0'
```

android/src/main/java/com/reactnativestripesdk/PaymentMethodCreateParamsFactory.kt
```diff
@@ -99,14 +110,14 @@ class PaymentMethodCreateParamsFactory(private val clientSecret: String, private
 
     if (paymentMethodId != null) {
       val cvc = getValOr(params, "cvc", null)
-      val paymentMethodOptionParams = if (cvc != null) PaymentMethodOptionsParams.Card(cvc) else null
```
ccen-stripe, line 102: We should also pass appID to paymentOptions when it's wechat by
```
val paymentMethodOptionParams = PaymentMethodOptionsParams.WeChatPay(appId)
```
like what we are doing in iOS `PaymentMethodFactory.createWeChatPayPaymentMethodOptions`

ccen-stripe (dismissed):

example/server/index.ts
```diff
@@ -60,6 +60,10 @@ function getKeys(payment_method?: string) {
       publishable_key = process.env.STRIPE_PUBLISHABLE_KEY_MX;
       secret_key = process.env.STRIPE_SECRET_KEY_MX;
       break;
+    case 'wechat_pay':
+      publishable_key = process.env.STRIPE_PUBLISHABLE_KEY_WECHAT;
+      secret_key = process.env.STRIPE_SECRET_KEY_WECHAT;
```
thorsten-stripe, line 65: Can you add these to the `.env.example` file please?

ios/StripeSdk.swift
```diff
@@ -602,6 +602,10 @@ class StripeSdk: RCTEventEmitter, STPApplePayContextDelegate, STPBankSelectionVi
             paymentIntentParams.paymentMethodOptions = paymentMethodOptions
             paymentIntentParams.shipping = Mappers.mapToShippingDetails(shippingDetails: params["shippingDetails"] as? NSDictionary)
         }
+        
+        if (paymentMethodType == STPPaymentMethodType.weChatPay) {
+            STPPaymentHandler.shared().simulateAppToAppRedirect = true
```
thorsten-stripe, line 607: We need to make sure to only set this when when the publishable key includes `test`. Can you add a regex to the if statement to check for that?

example/ios/StripeSdkExample/Info.plist
```diff
@@ -60,6 +70,8 @@
 	<array>
 		<string>armv7</string>
 	</array>
+	<key>LSApplicationQueriesSchemes</key>
+	<string>weixin</string>
```
thorsten-stripe, line 74: Is `LSApplicationQueriesSchemes` only needed if we call `canOpenURL`? Since we're not doing that here, maybe we can omit?

thorsten-stripe (approved):
LGTM, thanks

example/ios/StripeSdkExample/Info.plist
arekkubaczkowski: yes you're right

PR #446, opened 2021-07-21
Unsupported text input method

### LINKED ISSUE #440: CardField error thrown on RN0.62 after upgrading to 0.1.5
**Describe the bug**
An error is thrown when the CardField component gets unmounted.

**To Reproduce**
1. Use RN version 0.62.3 or below
2. Use v0.1.5 of stripe-react-native
3. Use the CardField component
4. Make the CardField component unmount (in my case navigating to another screen for ex)
5. There should be an error thrown: `TextInputState.currentlyFocusedInput() is not a function` *(iOS)*, `undefined is not a function` *(Android)*.

**Expected behavior**
No error to be thrown.

**Screenshots**
~~If applicable, add screenshots to help explain your problem.~~

**Desktop:**
 - OS: MacOS
 - Version: Big Sur

**React Native:**
- Version: 0.62.3

**Smartphone:**
 - Device: iPhone6
 - OS: iOS
 - Version: 12

**Additional context**
The TextInputState **currentlyFocusedInput** method was introduced in RN0.63. Previously, the **currentlyFocusedField** method had to be used.

On my end I'm trying to patch this by using the old method. It would be nice if the library could support both though.

arekkubaczkowski:
resolves #440 

this fix resolves issue which throws exception if certain TextInputState methods are not available since it has been added from 
 react-native 0.63x.

### DIFF
--- a/src/components/CardField.tsx
+++ b/src/components/CardField.tsx
@@ -17,9 +17,8 @@ import {
 } from 'react-native';
 const TextInputState = require('react-native/Libraries/Components/TextInput/TextInputState');
 
-const CardFieldNative = requireNativeComponent<CardFieldInput.NativeProps>(
-  'CardField'
-);
+const CardFieldNative =
+  requireNativeComponent<CardFieldInput.NativeProps>('CardField');
 
 const unsupportedMethodMessage = (field: string) =>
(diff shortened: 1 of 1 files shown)


thorsten-stripe (approved):
LGTM, thanks

erikhu:
@arekkubaczkowski amazing! thank you so much

PR #454, opened 2021-07-22
[plugin] fix: never set an empty merchant identifier array with plugin

cruzach:
Previously, if you applied the plugin and didn't provide any merchant IDs, it would create an empty array of merchant IDs:

```
<key>com.apple.developer.in-app-payments</key>
<array/>
```

Also upgraded the config-plugins package (no longer includes things like jimp, so should be significantly smaller)

### DIFF
--- a/package.json
+++ b/package.json
@@ -160,6 +160,6 @@
     ]
   },
   "dependencies": {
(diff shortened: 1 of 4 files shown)


thorsten-stripe (approved):
LGTM, thanks

PR #487, opened 2021-08-04
docs workflow

### LINKED ISSUE #471: GitHub action for API ref auto gen
Create a GitHub action that will run `yarn docs` and copy the output and commit it to the `gh-pages` branch: https://github.com/stripe/stripe-react-native/tree/gh-pages

image

arekkubaczkowski:
Create a workflow for auto deploying api reference documentation into gh-pages branch
closes #471


.github/workflows/generate-docs.yml
```diff
@@ -0,0 +1,44 @@
...
+          yarn example && yarn
+
+      - name: Build docs
+        run: yarn docs
+
+      - name: Deploy
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./docs
```
thorsten-stripe, line 43: I think we should add `force_orphan: true` there isn't a need to keep a version history for docs.

thorsten-stripe (approved):
Just one nit, after that LGTM, thanks.

PR #491, opened 2021-08-04
upgrade RN in example to 0.64.2

arekkubaczkowski:


thorsten-stripe:
I think this would break expo, no https://blog.expo.dev/expo-sdk-42-579aee2348b6? @cruzach

cruzach:
Yeah, it would break running the example in Expo Go (although we do still have the snack example, but not ideal)

FYI we're looking into upgrading to RN 64 in the next SDK (43)

PR #493, opened 2021-08-05
fix: google pay type import

### LINKED ISSUE #492: [Bug] Incorrect typing import on useGooglePay hook, type information lost
**Describe the bug**
Type information is being lost on the `useGooglePay` hook, as it is incorrectly being importing `GooglePay` types.

The file `stripe-react-native/lib/typescript/src/hooks/useGooglePay.d.ts` imports  the `GooglePay` types as follows:
`import type { GooglePay } from './src/types';`
which is incorrect. The file should be imported as follows instead:
`import type { GooglePay } from '../types';`

**To Reproduce**
Go to file linked above in the node_modules folder to confirm

arekkubaczkowski:
Fixes #492

### DIFF
--- a/src/hooks/useGooglePay.tsx
+++ b/src/hooks/useGooglePay.tsx
@@ -1,5 +1,5 @@
 import { useCallback, useState } from 'react';
-import type { GooglePay } from 'src/types';
+import type { GooglePay } from '../types';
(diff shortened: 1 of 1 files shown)


thorsten-stripe (approved):
LGTM, thanks

PR #500, opened 2021-08-09
Fix custom flow.

jamesblasco:
While payment sheet is in custom flow `presentPaymentSheetPromise` is called twice, on `ON_PAYMENT_OPTION_ACTION` and `ON_PAYMENT_RESULT_ACTION` intents.

This breaks the flutter plugin, as the promised is expected to be returned only once.


arekkubaczkowski (approved):

arekkubaczkowski:
LGTM

PR #501, opened 2021-08-09
Remove params for presentPaymentSheet

jamesblasco:
Right now presentPaymentSheets has two params:

 - `clientSecret` not used anywhere
 - `confirmPayment`, used to differentiate between single-step and multipe-step payment sheets.

This last param is not very clear for developers as it always needs to be the opposite of the param `customFlow` from initPaymentSheet. This could be simplified to inherit the config from initPaymentSheet, or otherwise explain better and throw an error in the case  `customFlow` and `confirmPayment` are the same.

My proposal is to remove both params. This would look like the native implementations that don't need any params in the respective methods

### DIFF
--- a/ios/StripeSdk.swift
+++ b/ios/StripeSdk.swift
@@ -140,27 +140,32 @@ class StripeSdk: RCTEventEmitter, STPApplePayContextDelegate, STPBankSelectionVi
     func confirmPaymentSheetPayment(resolver resolve: @escaping RCTPromiseResolveBlock,
                                     rejecter reject: @escaping RCTPromiseRejectBlock) -> Void  {
         DispatchQueue.main.async {
-            self.paymentSheetFlowController?.confirm(from: UIApplication.shared.delegate?.window??.rootViewController ?? UIViewController()) { paymentResult in
-                switch paymentResult {
-                case .completed:
-                    resolve([])
-                case .canceled:
-                    resolve(Errors.createError(PaymentSheetErrorType.Canceled.rawValue, "The payment has been canceled"))
-                case .failed(let error):
-                    resolve(Errors.createError(PaymentSheetErrorType.Failed.rawValue, error.localizedDescription))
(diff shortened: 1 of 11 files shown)


arekkubaczkowski:
@jamesblasco  I think the approach is fine and reasonable but iOS doesn't compile at all, please fix this first.

ios/StripeSdk.swift
```diff
@@ -140,27 +140,32 @@ class StripeSdk: RCTEventEmitter, STPApplePayContextDelegate, STPBankSelectionVi
...
         DispatchQueue.main.async {
-            self.paymentSheetFlowController?.confirm(from: UIApplication.shared.delegate?.window??.rootViewController ?? UIViewController()) { paymentResult in
-                switch paymentResult {
-                case .completed:
-                    resolve([])
-                case .canceled:
-                    resolve(Errors.createError(PaymentSheetErrorType.Canceled.rawValue, "The payment has been canceled"))
-                case .failed(let error):
-                    resolve(Errors.createError(PaymentSheetErrorType.Failed.rawValue, error.localizedDescription))
+            if let paymentSheetFlowController = self.paymentSheetFlowController {
```
arekkubaczkowski, line 143: please fix the warning in this line

ios/StripeSdk.swift
```diff
@@ -140,27 +140,32 @@ class StripeSdk: RCTEventEmitter, STPApplePayContextDelegate, STPBankSelectionVi
...
+                    }
+                   
                 }
+            } else {
+                resolve(Errors.createError(PaymentSheetErrorType.Failed.rawValue, "No payment sheet has been initialized yet"))
             }
         }
     }
     
     @objc(presentPaymentSheet:resolver:rejecter:)
```
arekkubaczkowski, line 162: you have to update parameters here also, otherwise it won't compile

arekkubaczkowski:
there is create a new PR https://github.com/stripe/stripe-react-native/pull/514

PR #506, opened 2021-08-11
Enhance payment sheet DX

### LINKED ISSUE #505: Enhance the payment sheet DX
You can `initPaymentSheet` with bad params and it either crashes or does not validate and return an error. Validating the params passed and ensuring a developer-friendly error message is returned is vital as it helps users self-serve to correct themselves.

- Case A: pass `customerEphemeralKeySecret: “123”` on initPaymentSheet(), the SDK does not return an `error`, instead tries to present PaymentSheet then dismisses. Really hard for a user to debug and self help as there’s no error message. A support interaction of a user running into this issue zendesk link and second Zendesk link where the user passes a publishable key instead of EphemeralKey secret.
- Case B pass `paymentIntentClientSecret: “abc”`, no validation, instead app just crashes.
- Case c pass dummy customerId, above, app crashes.
- Case d) pass a valid ephemeral key, but no customer param, app crashes. 

@arekkubaczkowski please investigate whether the native SDKs can take care of any of this, and how we can prevent the crashes.

arekkubaczkowski:
closes #505

### DIFF
--- a/ios/Errors.swift
+++ b/ios/Errors.swift
@@ -36,7 +36,7 @@ class Errors {
     static internal let isPIClientSecretValidRegex: NSRegularExpression? = try? NSRegularExpression(
         pattern: "^pi_[^_]+_secret_[^_]+$", options: [])
 
-    static internal let isSIClientSecretValidRegex: NSRegularExpression? = try? NSRegularExpression(
+    static internal let isSetiClientSecretValidRegex: NSRegularExpression? = try? NSRegularExpression(
         pattern: "^seti_[^_]+_secret_[^_]+$", options: [])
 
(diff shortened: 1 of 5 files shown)


souhe (approved):

ios/Errors.swift
```diff
@@ -33,6 +33,25 @@ enum CreateTokenErrorType: String {
 }
 
 class Errors {
+    static internal let isPIClientSecretValidRegex: NSRegularExpression? = try? NSRegularExpression(
```
thorsten-stripe, line 36: I think it also makes sense to add one for `customerEphemeralKeySecret` and check that it starts with `ek_`?

thorsten-stripe (approved):
Just one nit, otherwise LGTM, thanks.

PR #512, opened 2021-08-12
chore: tipsi migration guide

### LINKED ISSUE #364: [Docs] tipsi-stripe migration guide
It's come up a couple of times that folks migrating from tipsi-stripe are struggling a bit, especially when they are using class components and can't use hooks. We should provide some migration docs for them.

arekkubaczkowski:
closes #364


thorsten-stripe (approved):
LGTM, thanks

PR #513, opened 2021-08-12
isApplePaySupported - null as a default value 

### LINKED ISSUE #496: iOS: `isApplePaySupported` initial state should be `null` when using `useApplePay` hook
As a developer, I would like to lay out my UI differently if I _know_ Apple Pay is not supported, as opposed to when I am _waiting_ to find out.

Ideally, the `isApplePaySupported` field would be `null`, and updated to `false` or `true` when support status is known.   Apple's HID guidelines state that 
1. you should not show the Apple Pay button if the device doesn't support the payment method
2. you should render the Apple Pay button first, before other payment methods (EG a button to use Stripe's `CardField`.)

If the initial state were `null`, it would be possible to render EG an activity indicator until the state is resolved, then rendering the Apple Pay button (or not) along with the others, instead of having the UI reflow.

Laying out space for the Apple Pay button that starts blank, and stays blank unless `isApplePaySupported` resolves to `true`?  But this sort of dictates a vertically stacked array of buttons, with unused whitespace, so not awesome.

**Additional context**
Thanks for reading!

arekkubaczkowski:
closes #496

### DIFF
--- a/src/hooks/useStripe.tsx
+++ b/src/hooks/useStripe.tsx
@@ -51,7 +51,9 @@ import {
  * useStripe hook
  */
 export function useStripe() {
-  const [isApplePaySupported, setApplePaySupported] = useState(false);
+  const [isApplePaySupported, setApplePaySupported] = useState<boolean | null>(
+    null
+  );
 
   useEffect(() => {
     async function checkApplePaySupport() {


thorsten-stripe (approved):
LGTM, thanks

PR #514, opened 2021-08-12
Simplify present payment sheet

arekkubaczkowski:
Right now presentPaymentSheets has two params:

 - `clientSecret` not used anywhere
 - `confirmPayment`, used to differentiate between single-step and multipe-step payment sheets.

This last param is not very clear for developers as it always needs to be the opposite of the param `customFlow` from initPaymentSheet. This could be simplified to inherit the config from initPaymentSheet, or otherwise explain better and throw an error in the case  `customFlow` and `confirmPayment` are the same.

My proposal is to remove both params. This would look like the native implementations that don't need any params in the respective methods


thorsten-stripe (approved):
LGTM, thanks

PR #525, opened 2021-08-18
add openApplePaySetup method

### LINKED ISSUE #400: Stripe equivalent for `openNativePaySetup` from `tipsi-stripe`
`tipsi-stripe`, from which we are migrating, had a method called `openNativePaySetup` which allowed us to open Apple Wallet from within our app, in case the user hadn't set it up (seen here: https://github.com/tipsi/tipsi-stripe/blob/2e777c704bd00833fa1642035b03886049563846/src/Stripe.js#L232).

We would like to have a method or way to do the same, as well as to be able to check if Apple Pay is support but **not** set up. Right now, if I'm not mistaken, `isApplePaySupported` checks for both, so there's no in-between.

arekkubaczkowski:
closes #400

### DIFF
--- a/ios/StripeSdk.m
+++ b/ios/StripeSdk.m
@@ -25,6 +25,10 @@ RCT_EXTERN_METHOD(
                   rejecter: (RCTPromiseRejectBlock)reject)
 
 RCT_EXTERN_METHOD(
+                  openApplePaySetup: (RCTPromiseResolveBlock)resolve
+                  rejecter: (RCTPromiseRejectBlock)reject)
+
+RCT_EXTERN_METHOD(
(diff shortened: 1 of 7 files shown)


thorsten-stripe:
Is this testable in the simulator? For me nothing happens upon calling `openApplePaySetup` but `error` is returned as `undefined`.

arekkubaczkowski:
> Is this testable in the simulator? For me nothing happens upon calling `openApplePaySetup` but `error` is returned as `undefined`.

It requires access to Apple wallet so it is testable only on real device

thorsten-stripe (approved):
LGTM, thanks

PR #545, opened 2021-08-30
feat: improve handling onGooglePayReady callback

### LINKED ISSUE #528: Google Pay ON_INIT_GOOGLE_PAY initGooglePayPromise called twice
Hi!

When integrating the new Google Pay functionality we noticed that this line is called twice:

https://github.com/stripe/stripe-react-native/blob/master/android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt#L177

caused by using the same callback twice:
https://github.com/stripe/stripe-react-native/blob/master/android/src/main/java/com/reactnativestripesdk/GooglePayFragment.kt#L55
https://github.com/stripe/stripe-react-native/blob/master/android/src/main/java/com/reactnativestripesdk/GooglePayFragment.kt#L70
and thus causing the broadcast intent to be fired twice.

Does react-native handle this issue gracefully? On Flutter the app will typically crash as the promise cannot be resolved a second time.

souhe:
Fixes #528 

Now we are waiting for both callbacks (from `GooglePayMethodLauncher` and `GooglePayLauncher`) to be called in order to send onReady event.

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/GooglePayFragment.kt
+++ b/android/src/main/java/com/reactnativestripesdk/GooglePayFragment.kt
@@ -16,6 +16,26 @@ import com.stripe.android.model.StripeIntent
 class GooglePayFragment : Fragment() {
   private var googlePayLauncher: GooglePayLauncher? = null
   private var googlePayMethodLauncher: GooglePayPaymentMethodLauncher? = null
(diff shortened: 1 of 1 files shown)


arekkubaczkowski (approved):

arekkubaczkowski:
LGTM, thanks

PR #548, opened 2021-08-31
split apple pay address by new line

### LINKED ISSUE #531: Apple Pay 'street' field with two lines not being split
When Apple Pay provides the street field along with city, state, etc., both in billing address and in shipping address and, the address has two lines (e.g., number and street and, suite number, etc.), street is given by Apple Pay as 'line1'\n'line2'. It must be split manually by the developer.
This is not being done in stripe-react-native, instead, all the content of 'street' is loaded in 'line1' field.

1. Call presentApplePay requiring 'postalAddress' for billing or shipping address.
2. In Apple Pay sheet select an address (billing or shipping) with two street lines in it.
3. Check out billingDetails/address/line1 in the payment method returned and both lines will be in line1.
4. Check out billingDetails/address/line2 in the payment method returned and it will be empty.
5. Continue to complete the payment and the same will happen with the shipping address node in the payment intent.

Expected behavior
When a contact's address has two lines, the 'street' field provided by Apple Pay must be split (using '\n' as the separator) and the resulting components must be loaded in the corresponding fields (line1 and line2).

 - Device: iPhone 11 simulator
 - OS: iOS. 14.5
 - Version:  0.2.0

NOTE: Street is not provided by Apple Pay until the purchase is completed, only the rest of the address fields are provided through useApplePay (onShippingContactSelected).

Workaround
Once the payment method is attached to a customer, split the street and update it to save the changes. Once the payment intent is processed successfully, update its address as well.

arekkubaczkowski:
closes #531

### DIFF
--- a/ios/StripeSdk.swift
+++ b/ios/StripeSdk.swift
@@ -336,6 +336,15 @@ class StripeSdk: RCTEventEmitter, STPApplePayContextDelegate, STPBankSelectionVi
     
     func applePayContext(_ context: STPApplePayContext, didCreatePaymentMethod paymentMethod: STPPaymentMethod, paymentInformation: PKPayment, completion: @escaping STPIntentClientSecretCompletionBlock) {
         self.applePayCompletionCallback = completion
+        
+        let address = paymentMethod.billingDetails?.address?.line1?.split(whereSeparator: \.isNewline)
+        if (address?.indices.contains(0) == true) {
+            paymentMethod.billingDetails?.address?.line1 = String(address?[0] ?? "")
+        }
+        if (address?.indices.contains(1) == true) {
+            paymentMethod.billingDetails?.address?.line2 = String(address?[1] ?? "")
+        }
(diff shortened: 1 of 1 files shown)


YoQuieroDistraerlo:
Hey @arekkubaczkowski, thank you for this!
I've tested your code and what I see is as follows: 'paymentMethod' returned from 'presentApplePay' has the street correctly split in line1, line2.

However, the payment method in backend is still being created with the original street from Apple Pay (i.e., 'line1'\n'line2') and the payment intent is also still being created with the same problem.

Please note that this is important not only for logistics but also for the address checks to work properly.
Thanks again!

arekkubaczkowski:
@YoQuieroDistraerlo can you open another issue and elaborate more about this case?

YoQuieroDistraerlo:
> @YoQuieroDistraerlo can you open another issue and elaborate more about this case?

Added #553

PR #587, opened 2021-09-15
fix: update Podfile.lock stripe-react-native version

ivankoop:
Update Podfile.lock stripe-react-native version on example app.


arekkubaczkowski:
lgtm, thanks

arekkubaczkowski (approved):

PR #601, opened 2021-09-20
Add button color, return URL and billing details to JS and iOS #1

porter-stripe:
- Adds button color, return URL and billing details in JS and iOS, Android will come later once we have a snapshot of the GA candidate
- This is going into a feature branch in my fork, then eventually into master. For now I don't have permissions to the react native repo.
- TODO: Android stuff, `allowsDelayedPaymentMethods`

### DIFF
--- a/example/ios/Podfile
+++ b/example/ios/Podfile
@@ -6,6 +6,10 @@ target 'StripeSdkExample' do
   config = use_native_modules!
 
   use_react_native!(:path => config["reactNativePath"])
+pod 'Stripe', :git => 'https://github.com/stripe-ios/stripe-ios.git', :branch => 'private'
+pod 'StripeUICore', :git => 'https://github.com/stripe-ios/stripe-ios.git', :branch => 'private'
+pod 'StripeCore', :git => 'https://github.com/stripe-ios/stripe-ios.git', :branch => 'private'
+
 
   pod 'stripe-react-native', :path => '../..'
   # Enables Flipper.
--- a/src/types/PaymentSheet.ts
+++ b/src/types/PaymentSheet.ts
@@ -7,6 +7,9 @@ export declare namespace PaymentSheet {
       customFlow?: boolean;
       merchantDisplayName?: string;
       style?: 'alwaysLight' | 'alwaysDark' | 'automatic';
+      returnURL?: string;
+      primaryButtonColor?: string;
+      defaultBillingDetails?: BillingDetails;
     };
 
   type ClientSecretParams =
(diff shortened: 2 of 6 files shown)


porter-stripe:
CI is gonna fail since it can't pull in our internal pods.

example/ios/Podfile
```diff
@@ -6,6 +6,10 @@ target 'StripeSdkExample' do
   config = use_native_modules!
 
   use_react_native!(:path => config["reactNativePath"])
+pod 'Stripe', :git => 'https://github.com/stripe-ios/stripe-ios.git', :branch => 'private'
+pod 'StripeUICore', :git => 'https://github.com/stripe-ios/stripe-ios.git', :branch => 'private'
+pod 'StripeCore', :git => 'https://github.com/stripe-ios/stripe-ios.git', :branch => 'private'
```
yuki-stripe, line 11: Should this podfile work for public users? Not sure about permissions to the private github

src/types/PaymentSheet.ts
```diff
@@ -40,7 +44,20 @@ export declare namespace PaymentSheet {
         merchantCountryCode?: string;
         testEnv?: boolean;
       };
-
+  export interface Address {
```
yuki-stripe, line 47: how do users see documentation for these things? I didn't find it in `/docs`, do they use the native SDK docs?

src/types/PaymentSheet.ts
```diff
@@ -40,7 +44,20 @@ export declare namespace PaymentSheet {
...
+    city?: string;
+    country?: string;
+    line1?: string;
+    line2?: string;
+    postalCode?: string;
+    state?: string;
+  }
+  export interface BillingDetails {
+    address: Address;
```
yuki-stripe, line 56: I think I remember Android doing optionality differently here.  Will this interface work for Android, too?  @brnunes-stripe

example/ios/Podfile
porter-stripe: No this won't work, we'll need to update this back to the original once PaymentSheet is GA'd

src/types/PaymentSheet.ts
porter-stripe: That's a good question, this SDK (react native) isn't documented well. There will need to be some effort made to get that fixed. Not sure what that looks like in the short-term, there's many types and interfaces not documented at all....

src/types/PaymentSheet.ts
porter-stripe: I don't have a snapshot build of Android to really confirm whether or not this will work, so I can't say yet. @brnunes-stripe let me know if you think this will be compatible.

example/ios/Podfile
porter-stripe: I just did it this way so I could get the latest build of PaymentSheet so I could test E2E in iOS.

src/types/PaymentSheet.ts
brnunes-stripe: Only difference is that `address` is also optional. Can we make it optional on iOS too so the APIs match?
https://github.com/stripe/stripe-android/blob/master/paymentsheet/src/main/java/com/stripe/android/paymentsheet/PaymentSheet.kt#L129-L144

porter-stripe:
@yuki-stripe correct, once iOS and Android GA I'll throw up a quick change pointing to the latest versions, then we'll publish the RN package!

Merging this into a feature branch for now...

PR #629, opened 2021-10-07
feat: Add validation state to CardField (#423)

### LINKED ISSUE #423: CardField handle validation errors
I'm using `CardField` and `createToken` method in my application. Everything is working perfectly fine.
But I want to do customizations with `CardField`, and I'm unable to do that. 

`CardField` is handling all the validations which are really great, but I also want to display errors to users, that which specific fields have issues.
I was able to handle the error message for `ExpiryDate`, as shown in the below screenshots.

Screenshot 2021-07-15 at 2 40 03 PM

Screenshot 2021-07-15 at 2 40 10 PM

**How I handled this**
`CardField` has a method on `onCardChange`, in which I have access to `expiryMonth` and `expiryYear`. I can validate it, and show errors to the user. But I can't do it for the card number. Because when the card number is empty or is not valid, `onCardChange` gives `null` for `last4` digits of the card. Same for `CVC`. 

`CardField` has a method on `onCardChange`, which gives `cardDetails`. Addition of some bool attributes like `validCard`, `validExpiryDate`, and `validCVC`, so I can show & hide errors at my end. Just like we have `complete` in `cardDetails`, which is `false` when any of the fields are invalid, and `true` when all the fields are valid. 

N/A

**Additional context**
N/A

atiernan:
Closes #423 

This adds three new fields to the card details returned by `onCardChange`: `validCVC`, `validExpiryDate`, and `validNumber`. On iOS, each field has a possible value of "Valid", "Invalid", "Incomplete", and "Unknown". The Android SDK doesn't seem to expose whether a field is incomplete or not, so Android has "Valid", "Invalid", and "Unknown".

This is my first time working on a native module so I'm not 100% on the Swift and Kotlin syntax, let me know if there's a better way to write this and I'll update it

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/CardChangedEvent.kt
+++ b/android/src/main/java/com/reactnativestripesdk/CardChangedEvent.kt
@@ -36,6 +36,9 @@ internal class CardChangedEvent constructor(viewTag: Int, private val cardDetail
     }
 
     eventData.putBoolean("complete", complete)
+    eventData.putString("validNumber", cardDetails["validNumber"]?.toString())
+    eventData.putString("validCVC", cardDetails["validCVC"]?.toString())
+    eventData.putString("validExpiryDate", cardDetails["validExpiryDate"]?.toString())
 
     if (postalCodeEnabled) {
       eventData.putString("postalCode", cardDetails["postalCode"]?.toString())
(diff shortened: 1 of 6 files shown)


atiernan:
For my use case the Incomplete state is a necessity, as my apps styling would be out of sync with the SDKs error styling (showing the error state when the field is empty/incomplete)

Would it not be better to have the ideal situation on iOS and push for an update to the Android SDK, to align them, rather than settle for a subpar solution on both platforms.

I can see the want for consistency between platforms though, so I'm happy to update the PR either way

atiernan:
@csabol-stripe @michelleb-stripe Is there anything I can do to help progress this?

giorgiofellipe:
Thanks @atiernan for the great work.

@arekkubaczkowski would be really great to have it merged and published. 

Let me know if there is anything I can help.

acomley-stripe (approved):

Issue #185, opened 2021-05-06
Android: 3D secure popup styling changes


benomatis:
- [ ] 1. Android: 3DS progress spinner color
- [ ] 2. Android: The section titles under the 3DS2 complete authentication take the color of the main `backgroundColor` (which is supposed to be iOS only?) when opened

arekkubaczkowski:
there was also one thing on my side to fix: https://github.com/stripe/stripe-react-native/pull/234/files#diff-4c983f2aa3f1856047fffa58ecd9f6447b449cfa4f480e8eb159ac858758aa57R424

- the first thing is that `accentColor` now changes the section titles color, @michelleb-stripe is it intentional?
- `accentColor` probably should change the spinner color (I haven't found any different way to set it tho)

So I guess the rest of issues should be fixed on android side, @michelleb-stripe let me know if I am wrong

thorsten-stripe:
@michelleb-stripe can you help out with this?

michelleb-stripe:
@thorsten-stripe and @arekkubaczkowski Let me get back to you on this.

michelleb-stripe:
We have this on our list to look at, but I am not sure I will get to it this week.

michelleb-stripe:
> the first thing is that accentColor now changes the section titles color, @michelleb-stripe is it intentional?
This does intentionally change the section title colors

> accentColor probably should change the spinner color (I haven't found any different way to set it tho)
From the code it looks like setting the accentColor should set the spinner color.  

Let me know if you are seeing anything different @arekkubaczkowski, and if there are any other parts to this issue that I am missing.

arekkubaczkowski:
I've checked this again and it works properly within the latest library version

benomatis:
@arekkubaczkowski sorry, what is working properly exactly? The initial points were:

1. Android: 3DS progress spinner color
  **Question**: have we exposed a separate setting that would control the color of the spinner?

 2. Android: The section titles under the 3DS2 complete authentication take the color of the main backgroundColor (which is supposed to be iOS only?) when opened
  **2 Questions**: a) are the section titles supposed to take the color of the main backgroundColor at all? b) Has there been a documentation change added that corrected this value being iOS only?

arekkubaczkowski:
>Android: 3DS progress spinner color
Question: have we exposed a separate setting that would control the color of the spinner?

According to recent @michelleb-stripe answer it's intentional. @michelleb-stripe any chance to split it to two separate params?

[...]

PR #687, opened 2021-10-29
Fix BillingAddress properties being null

musthafa1996:
Fixed BillingAddress properties being null while creating a payment method on Android.
This issue has been addressed at #185.


bill-pairaktaridis:
Does this address optional fields being treated as mandatory on Android?

charliecruzan-stripe:
I think this is fixed by https://github.com/stripe/stripe-react-native/pull/751/, which will be released soon

PR #748, opened 2021-12-14
Add capabilityCredit and capabilityDebit merchantCapability options for Apple Pay

### LINKED ISSUE #728: Support merchantCapabilities for Apple Pay
We only support debit card payments, but our Apple Pay popup currently includes credit cards, and when users select a credit card, the transaction is declined, which is a bad user experience.

We want to only allow them to select debit cards from Apple Pay.

This should be possible by using `merchantCapabilities` on `PKPaymentRequest`:

https://developer.apple.com/documentation/passkit/pkmerchantcapability

We want to set `capabilityCredit` to `false`.

We can tell users not to select credit cards, or decline transactions after they select one, but both are worse user experiences than doing this with Apple Pay's filtering API.

brendanlong:
These allow developers to pass `capabilityCredit: true` or `capabilityDebit: true` when opening
an Apple Pay interface to filter to only credit or debit cards.

See: https://developer.apple.com/documentation/apple_pay_on_the_web/applepaypaymentrequest/1916123-merchantcapabilities

Fixes #728

### DIFF
--- a/ios/StripeSdk.swift
+++ b/ios/StripeSdk.swift
@@ -472,6 +472,13 @@ class StripeSdk: RCTEventEmitter, STPApplePayContextDelegate, STPBankSelectionVi
         let merchantIdentifier = self.merchantIdentifier ?? ""
         let paymentRequest = StripeAPI.paymentRequest(withMerchantIdentifier: merchantIdentifier, country: country, currency: currency)
         
+        if (params["capabilityCredit"] as? Bool ?? false) {
+          paymentRequest.merchantCapabilities.update(with: .capabilityCredit)
+        }
+        if (params["capabilityDebit"] as? Bool ?? false) {
+          paymentRequest.merchantCapabilities.update(with: .capabilityDebit)
+        }
+        
         let requiredShippingAddressFields = params["requiredShippingAddressFields"] as? NSArray ?? NSArray()
         let requiredBillingContactFields = params["requiredBillingContactFields"] as? NSArray ?? NSArray()
         let shippingMethods = params["shippingMethods"] as? NSArray ?? NSArray()
--- a/src/types/ApplePay.ts
+++ b/src/types/ApplePay.ts
@@ -59,6 +59,8 @@ export namespace ApplePay {
   }
 
   export interface PresentParams {
+    capabilityCredit?: Boolean;
+    capabilityDebit?: Boolean;
     cartItems: CartSummaryItem[];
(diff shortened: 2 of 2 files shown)


brendanlong:
I doubt this is exactly how you want to implement this, but I figured example code would make it easier to find this if you do want to do it.

ios/StripeSdk.swift
```diff
@@ -472,6 +472,13 @@ class StripeSdk: RCTEventEmitter, STPApplePayContextDelegate, STPBankSelectionVi
         let merchantIdentifier = self.merchantIdentifier ?? ""
         let paymentRequest = StripeAPI.paymentRequest(withMerchantIdentifier: merchantIdentifier, country: country, currency: currency)
         
+        if (params["capabilityCredit"] as? Bool ?? false) {
+          paymentRequest.merchantCapabilities.update(with: .capabilityCredit)
+        }
+        if (params["capabilityDebit"] as? Bool ?? false) {
+          paymentRequest.merchantCapabilities.update(with: .capabilityDebit)
+        }
+
```
charliecruzan-stripe, line 481: what do you think of something like

```
        if let cardType = params["supportedCardType"] as? NSString {
            if (cardType.isEqual(to: "Credit")) {
                paymentRequest.merchantCapabilities.update(with: .capabilityCredit)
            } else if (cardType.isEqual(to: "Debit")) {
                paymentRequest.merchantCapabilities.update(with: .capabilityDebit)
            }
        }
  ```

So we only add 1 param to handle this, and default to accepting both credit and debit?

charliecruzan-stripe:
It would also be good to see if Google Pay exposes a similar filter so we have cross-platform support for this feature: https://developers.google.com/pay/api/android/guides/tutorial#allowed-payment-methods

charliecruzan-stripe:
This is now released in v0.22.0 via https://github.com/stripe/stripe-react-native/pull/1164

jayfeldman12:
I might be missing something, but I don't see where the option to specify only credit or debit was done. In [presentParams](https://github.com/stripe/stripe-react-native/blob/master/src/types/ApplePay.ts#L102) I don't see a field relevant to this.

charliecruzan-stripe:
This is released as part of the new Platform Pay API, which replaced the individual Apple/Google Pay APIs: https://github.com/stripe/stripe-react-native/blob/master/src/types/PlatformPay.ts#L65

Issue #739, opened 2021-12-02
PostalCode not sent with CardField


bernardorubin:
**Describe the bug**
When using the CardField component the PostalCode isn't being sent. i can see the new Payment Method being correctly sent on the Stripe dashboard but the PostalCode is missing.

**To Reproduce**
1. Use the `CardField` component and set postalCodeEnabled to true.
2. Then send the data to stripe using 
```
const { error, paymentMethod } = await createPaymentMethod({
          type: "Card",
        });
```
3. Log into the stripe admin, filter the customer with the same email, notice how there is no Postal Code under the customer Payment methods details.

**Expected behavior**
Postal Code should be sent using CardField and should appear under the payment method details.

**Screenshots**
Screenshot of a payment method that was sent using `tipsi-stripe` with correct Postal Code Data (below) and the payment method being sent by `stripe-react-native` (above)
<img width="530" alt="Screen Shot 2021-12-01 at 20 14 33" src="https://user-images.githubusercontent.com/25016886/144509541-ec151229-fe1c-4820-92db-3b341f99ce9e.png">
<img width="378" alt="Screen Shot 2021-12-02 at 16 02 53" src="https://user-images.githubusercontent.com/25016886/144510406-43635124-4014-49dc-a5d0-3f4999ba4421.png">

**Smartphone (please complete the following information):**
 - OS: macOS
 - React Natie on both iOS and android

**Additional context**
I'm not getting any errors back from the response.

**Code:**
```
        const [card, setCard] = useState(null);

        const { error, paymentMethod } = await createPaymentMethod({
          type: "Card",
        });
        
       <CardField
        postalCodeEnabled={true}
        placeholder={{
          number: "1234 5678 1234 5678",
        }}
        cardStyle={{
          backgroundColor: theme.colors.basic200,
        }}
        style={{
          width: "100%",
          height: 50,
          marginVertical: theme.space.lg,
        }}
        onCardChange={(cardDetails) => {
          setCard(cardDetails);
        }}
      />
```

seanzhang-stripe:
Hi @bernardorubin thanks a lot for reporting this problem and I'm sorry for the inconvenience.

I'm able to reproduce the issue in Android, and I can see that the `post_code` is missing in the request and that is why you don't find it in the Dashboard. 

While waiting for the fix, here is a simple workaround that you can try.

All you need to do is create an empty `billingDetails` object and pass it to the `createPaymentMethod()`

```js
[...]

PR #751, opened 2021-12-16
fix: some updates to the mapToBillingDetails()

seanzhang-stripe:
Fix issue https://github.com/stripe/stripe-react-native/issues/739

1. Make `mapToBillingDetails()` return null if both `billingDetails` and `cardAddress` are both null.

2. We shouldn't ignore the `postalCode` and `country` from `cardAddress` when constructing the `PaymentMethod.BillingDetails`

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/Mappers.kt
+++ b/android/src/main/java/com/reactnativestripesdk/Mappers.kt
@@ -419,34 +419,39 @@ internal fun mapToAddress(addressMap: ReadableMap?, cardAddress: Address?): Addr
 }
 
 internal fun mapToBillingDetails(billingDetails: ReadableMap?, cardAddress: Address?): PaymentMethod.BillingDetails? {
-  if (billingDetails == null) {
-    return null
+  if (billingDetails == null && cardAddress == null) {
+    return null;
   }
   val address = Address.Builder()
-    .setPostalCode(getValOr(billingDetails, "addressPostalCode"))
-    .setCity(getValOr(billingDetails, "addressCity"))
-    .setCountry(getValOr(billingDetails, "addressCountry"))
-    .setLine1(getValOr(billingDetails, "addressLine1"))
-    .setLine2(getValOr(billingDetails, "addressLine2"))
-    .setState(getValOr(billingDetails, "addressState"))
-
-    cardAddress?.let { ca ->
-      ca.postalCode?.let {
-        if (it.isNotEmpty()) {
-          address.setPostalCode(it)
-        }
-      }
-      ca.country?.let {
-        address.setCountry(it)
+  if (billingDetails != null) {
+    address
+      .setPostalCode(getValOr(billingDetails, "addressPostalCode"))
+      .setCity(getValOr(billingDetails, "addressCity"))
+      .setCountry(getValOr(billingDetails, "addressCountry"))
+      .setLine1(getValOr(billingDetails, "addressLine1"))
+      .setLine2(getValOr(billingDetails, "addressLine2"))
+      .setState(getValOr(billingDetails, "addressState"))
+  }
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/reactnativestripesdk/Mappers.kt
```diff
@@ -419,34 +419,39 @@ internal fun mapToAddress(addressMap: ReadableMap?, cardAddress: Address?): Addr
...
-    cardAddress?.let { ca ->
-      ca.postalCode?.let {
-        if (it.isNotEmpty()) {
-          address.setPostalCode(it)
-        }
-      }
-      ca.country?.let {
-        address.setCountry(it)
+  if (billingDetails != null) {
+    address
```
charliecruzan-stripe, line 427: if we move 
```
  val billingDetailsBuilder =  PaymentMethod.BillingDetails.Builder()
```
up to where we're also declaring the other builder, we can consolidate the two `if (billingDetails != null) {` blocks down to one. 

maybe something like:

```
  val billingDetailsBuilder =  PaymentMethod.BillingDetails.Builder()
  val addressBuilder = Address.Builder()

  if (billingDetails != null) {
    billingDetailsBuilder
      .setName(getValOr(billingDetails, "name"))
      .setPhone(getValOr(billingDetails, "phone"))
      .setEmail(getValOr(billingDetails, "email"))
    addressBuilder
      .setPostalCode(getValOr(billingDetails, "addressPostalCode"))
      .setCity(getValOr(billingDetails, "addressCity"))
      .setCountry(getValOr(billingDetails, "addressCountry"))
      .setLine1(getValOr(billingDetails, "addressLine1"))
      .setLine2(getValOr(billingDetails, "addressLine2"))
      .setState(getValOr(billingDetails, "addressState"))
  }
```
?

android/src/main/java/com/reactnativestripesdk/Mappers.kt
```diff
@@ -419,34 +419,39 @@ internal fun mapToAddress(addressMap: ReadableMap?, cardAddress: Address?): Addr
...
+  cardAddress?.let { ca ->
+    ca.postalCode?.let {
+      if (it.isNotEmpty()) {
+        address.setPostalCode(it)
       }
     }
+    ca.country?.let {
+      address.setCountry(it)
+    }
+  }
```
charliecruzan-stripe, line 444: Nit: I know this is what was here before, but maybe we should change it to be easier to read, like:

```
  if (cardAddress?.postalCode?.isNotEmpty() == true) {
    addressBuilder.setPostalCode(cardAddress.postalCode)
  }
  if (cardAddress?.country?.isNotEmpty() == true) {
    addressBuilder.setCountry(cardAddress.country)
  }
```

?

charliecruzan-stripe (commented):
thanks for this fix! and sorry for the long wait

android/src/main/java/com/reactnativestripesdk/Mappers.kt
seanzhang-stripe: Yup, I agreed! this is much more readable. 

Perhaps we should drop the `==true` as well?

seanzhang-stripe: Ah, `isNotEmpty()` returns Boolean or null, so `==true` is necessary.

android/src/main/java/com/reactnativestripesdk/Mappers.kt
seanzhang-stripe: Thanks for spotting the code smell. I'll fix it!

seanzhang-stripe:
@charliecruzan-stripe Thanks for review! I have fix the code smell, could you please take a look and see if it's good to go now?

PR #774, opened 2022-01-14
Add `currencyCode` support to initPaymentSheet for Google Pay & Setup Intents

brendanm-stripe:
Necessary to support Setup Intents including Google Pay

```
const { error } = await initPaymentSheet({
      customerId: customer,
      customerEphemeralKeySecret: ephemeralKey,
      setupIntentClientSecret: setupIntentClientSecret,
      customFlow: false,
      merchantDisplayName: 'Example Inc.',
      applePay: true,
      merchantCountryCode: 'US',
      currencyCode: 'USD', // this is required for setup intents
      style: 'automatic',
      googlePay: true,
      testEnv: true,
      primaryButtonColor: '#635BFF', // Blurple
      returnURL: 'stripe-example://stripe-redirect',
      defaultBillingDetails: billingDetails,
      allowsDelayedPaymentMethods: true,
    });
```

### DIFF
--- a/src/types/PaymentSheet.ts
+++ b/src/types/PaymentSheet.ts
@@ -37,11 +37,13 @@ export declare namespace PaymentSheet {
     | {
         googlePay?: true;
         merchantCountryCode: string;
+        currencyCode?: string;
         testEnv?: boolean;
       }
     | {
         googlePay?: false;
         merchantCountryCode?: string;
+        currencyCode?: string;
         testEnv?: boolean;
       };
   export interface Address {
--- a/android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt
+++ b/android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt
@@ -46,6 +46,7 @@ class PaymentSheetFragment : Fragment() {
     val customerId = arguments?.getString("customerId").orEmpty()
     val customerEphemeralKeySecret = arguments?.getString("customerEphemeralKeySecret").orEmpty()
     val countryCode = arguments?.getString("merchantCountryCode").orEmpty()
+    val currencyCode = arguments?.getString("currencyCode").orEmpty()
     val googlePayEnabled = arguments?.getBoolean("googlePay")
     val testEnv = arguments?.getBoolean("testEnv")
     val allowsDelayedPaymentMethods = arguments?.getBoolean("allowsDelayedPaymentMethods")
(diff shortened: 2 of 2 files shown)


brnunes-stripe:
This fix is correct.
Google Pay requires a currency code and Setup Intents don't have that field, so we need the user to tell us in the Google Pay configuration.
I'm not familiar with the React codebase, but it would be good to document that somewhere, similar to how we do [on the Android SDK](https://github.com/stripe/stripe-android/blob/master/paymentsheet/src/main/java/com/stripe/android/paymentsheet/PaymentSheet.kt#L291-L295).

porter-stripe (approved):

brendanm-stripe:
While there is not currently an example screen for Payment Sheet + Setup Intent, the non-Payment Sheet Google Pay docs do mention this `currencyCode` requirement: 
https://github.com/stripe/stripe-react-native/blob/master/docs/GooglePay.md#present-google-pay-sheet

and this is also noted in the Android SDK docs: https://stripe.com/docs/payments/save-and-reuse?platform=android&ui=payment-sheet#add-google-pay

In parallel to this PR, I've got a draft of changed to add this to the RN docs here: https://stripe.com/docs/payments/save-and-reuse?platform=react-native&ui=payment-sheet#add-google-pay

julestruong:
Hey guyz, thx a lot for this PR this could save my day. 
Any update on the ETA  ?

src/types/PaymentSheet.ts
```diff
@@ -37,11 +37,13 @@ export declare namespace PaymentSheet {
     | {
         googlePay?: true;
         merchantCountryCode: string;
+        currencyCode?: string;
```
charliecruzan-stripe, line 40:
```suggestion
        currencyCode: string;
```
if `googlePay === true`, then the currencycode _must_ be supplied, right?

charliecruzan-stripe (commented):
thanks for this fix 🙏 just one small question

src/types/PaymentSheet.ts
brendanm-stripe: Only when used with a setup intent for "setup mode". When used for _payment_ (with a payment intent), the currency is inherited from the payment intent (where it is a required parameter). 

The `SetupParams` type name I think is a bit of a misnomer here, or overlaps with Stripe nomenclature with Setup Intents, but I believe refers generally to _setting up the payment sheet_. The `ClientSecretParams ` for example provide cases to handle both payment intent and setup intent usage.

charliecruzan-stripe (approved):

brendanm-stripe:
Docs updated here: https://stripe.com/docs/payments/save-and-reuse?platform=react-native&ui=payment-sheet#react-native-google-pay

```
const { error, paymentOption } = await initPaymentSheet({
  // ...
  googlePay: true,
  merchantCountryCode: 'US',
  currencyCode: 'USD', // currency must be specified for setup mode
  testEnv: true, // use test environment
});
```

PR #783, opened 2022-01-23
call ON_FRAGMENT_CREATED before INIT_PAYMENTSHEET to prevent potential null pointer crashes

remonh87:
We noticed that we get from the Flutter Stripe community some nullpinter exceptions that happen a bit randomly. See issue https://github.com/flutter-stripe/flutter_stripe/issues/416 .  In my current project I also saw this issue happen and it looks like a race condition that causes the following cast being executed in the StripeModule while the fragment is being null.

```kotlin
if (intent.action == ON_FRAGMENT_CREATED) {
        paymentSheetFragment = (currentActivity as AppCompatActivity).supportFragmentManager.findFragmentByTag("payment_sheet_launch_fragment") as PaymentSheetFragment
      }
```

After some investigation it looks like the promise is being resolved before the fragment is created. This fix will prevent this from happening.

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt
+++ b/android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt
@@ -113,6 +113,10 @@ class PaymentSheetFragment : Fragment() {
       ) else null
     )
 
+    // Call ON_FRAGMENT_CREATED before ON_INIT_PAYMENT_SHEET
+    val intent = Intent(ON_FRAGMENT_CREATED)
+    localBroadcastManager.sendBroadcast(intent)
+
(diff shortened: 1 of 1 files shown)


charliecruzan-stripe:
Since intents are asynchronous, I think it's still possible to run into this issue even afterwards. What do you think of removing the need for the `ON_FRAGMENT_CREATED` intent altogether, like this: https://github.com/stripe/stripe-react-native/pull/788?

remonh87:
> Since intents are asynchronous, I think it's still possible to run into this issue even afterwards. What do you think of removing the need for the `ON_FRAGMENT_CREATED` intent altogether, like this: #788?

@charliecruzan-stripe yes that can work. Let's go for your solution I will close this one

PR #787, opened 2022-02-01
chore: Upgrade example app to RN 0.66

charliecruzan-stripe:
# Summary 

This upgrades the example app and confirms `stripe-react-native` works up to `react-native@0.66.4`. Also contains some light docs improvements

### Expo

Expo's most recent SDK still relies on RN .64, but I think it's safe to upgrade the example app's version of `react-native`. I added a section to CONTRIBUTING on [testing within expo](https://github.com/stripe/stripe-react-native/pull/787/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R56), and confirmed that the example app runs within Expo Go on both Android and iOS after following those steps


charliecruzan-stripe:
sorry everyone for all the review requests 😓  I think we might want to edit the [codeowners](https://github.com/stripe/stripe-react-native/blob/master/CODEOWNERS) so that you only get requested for reviews on changes to `android/` and `ios/`, and not `example/[android|ios]`? leave that up to you all, though

davidme-stripe:
I think we should move CODEOWNERS to use a team — @stripe-ios/ios-reviewers could be a good one? (Though I don't think it's usable from the @stripe org, so we might need to recreate it.) Then GitHub will automatically load balance the PRs across us.

example/ios/StripeSdkExample.xcodeproj/project.pbxproj
```diff
@@ -383,6 +383,7 @@
 				COPY_PHASE_STRIP = NO;
 				ENABLE_STRICT_OBJC_MSGSEND = YES;
 				ENABLE_TESTABILITY = YES;
+				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = "arm64 ";
```
davidme-stripe, line 386: Do we need this? I think it will break users on M1 MBPs.

davidme-stripe (approved):
Looks reasonable overall, one comment

example/ios/StripeSdkExample.xcodeproj/project.pbxproj
charliecruzan-stripe: It's included in the upgrade guide for v0.66, but looks like then gets removed again for 0.67- https://react-native-community.github.io/upgrade-helper/?from=0.66.4&to=0.67.2 (which is still very new so not upgrading just yet)

Builds fine on my non-M1 with & without these additions, so I'm fine removing this

Issue #725, opened 2021-11-17
GooglePay - PaymentConfiguration was not initialized.


AdamLee321:
**Bug Description**
When running initGooglePay on Android the app immediately crashes and throws this error: 

<img width="1026" alt="Screenshot 2021-11-17 at 16 14 51" src="https://user-images.githubusercontent.com/22890658/142239128-0373af69-bc02-495c-9411-7bd53339b51a.png">

**To Reproduce**
Removing this block of code stops the app from crashing, so the issue lies here:

```
componentDidMount() {
  initGooglePay({
    testEnv: true,
    merchantName: 'Test App',
    countryCode: 'US',
  })
    .then(response => {
      console.log('response', response);
    })
    .catch(err => {
      console.log('error', error);
    });
}
```

**Expected behavior**
This should return a response which will allow the app to render the google pay button.

**Test Device:**
 - Device: Samsung S20 Ultra
 - OS: Android
 - Version: 10

**Additional setup information:**
package.json: 
"@stripe/stripe-react-native": "^0.2.2",
"react": "17.0.2",
"react-native": "0.66.0",

app/build.gradle: 
implementation 'com.google.android.gms:play-services-wallet:18.1.3'

build.gradle: 
minSdkVersion = 21
compileSdkVersion = 30
targetSdkVersion = 30

If there is anything else I can provide to help identifying the issue let me know. If anyone knows a workaround could you post a fix please?

AdamLee321:
I found the issue and I am closing this as it has nothing to do with the library the issue was my app. I had code in build.gradle to use the exo player for playing video and had code which turned out was not needed. 

```
// jcenter repository is deprecated: https://developer.android.com/studio/build/jcenter-migration
        //noinspection JcenterRepositoryObsolete
        jcenter() {
            content {
                // however, react-native-video's native packages have not been migrated yet, so we keep this config for now
                includeModule("com.yqritc", "android-scalablevideoview")
                includeModule("com.google.android.exoplayer", "extension-okhttp")
                includeModule("com.google.android.exoplayer", "exoplayer")
                includeModule("com.google.android.exoplayer", "exoplayer-core")
                includeModule("com.google.android.exoplayer", "exoplayer-dash")
                includeModule("com.google.android.exoplayer", "exoplayer-hls")
                includeModule("com.google.android.exoplayer", "exoplayer-smoothstreaming")
                includeModule("com.google.android.exoplayer", "exoplayer-ui")
[...]

Issue #683, opened 2021-10-28
isGooglePaySupported without initGooglePay


SamKomesarook:
**Is your feature request related to a problem? Please describe.**
In order to simply check if Google Pay is supported on the device, it must be initialised with `initGooglePay`, then checked if an err occurred.

**Describe the solution you'd like**
Is it possible to create an `isGooglePaySupported` bool similar to `isApplePaySupported` returned from the `useGooglePay` hook?

**Describe alternatives you've considered**
Currently, just initialising Google Pay, then awkwardly passing the var from parent to children, as initialising multiple times in different components to check is leading to strange behaviour...

As for initialising once in a root component and saving the state to datastore, this could lead to issues where the store doesn't update after a user uninstalls Google Pay, which would throw an err. It would be best to check in real time when each component requires it.

**Additional context**
Different screens and components require the knowledge of if Google Pay is available, and initialising the pay each time is a little frustrating, so would be great if there was a cleaner way like the Apple Pay hook has.

charliecruzan-stripe:
Thanks for this request! It's in progress now

SamKomesarook:
Thanks for this yall!

Issue #617, opened 2021-09-30
Google Pay/Apple pay using CardField instead of PaymentSheet


Gabotron-ES:
Hi everybody, currently I can't seem to find how to use Google Pay or Apple Pay with CardField component, is it currently possible?

anija:
Hi, if you mean to avoid to use the Stripe Payment Sheet, then I've succesfully implemented Apple Pay and Google Pay with the CreatePaymentMethod flux in these days (with the latest version of the library). I don't use CardField component since both methods use native dialogues.

**Apple Pay**
use the hook:
`const { presentApplePay, isApplePaySupported, confirmApplePayPayment } = useApplePay();`

Then use isApplePaySupported to know if it's supported in order to enable/disable it.
Then i use 
```
const pm = await presentApplePay({
    cartItems: nativePaymentParams.items,
    country: 'IT',
    currency: cart.currency,
    shippingMethods: [],
    requiredShippingAddressFields: [],
    requiredBillingContactFields: [],
});
```
To obtain the Payment Method ID and proceed like with a normal Card on backend side.
After the payment intent has been requested to stripe by the backend, returning the client secret to me, i do:
`await confirmApplePayPayment(clientSecret);`
This will close the Apple Pay native dialogue.

**Google Pay**
Since i can't find any isGooglePaySupported, i do this:
```
const { error } = await initGooglePay({
  merchantName: config.merchantName,
  countryCode: config.merchantCountryCode,
  testEnv: version.environment !== 'prod',
});
```
Assuming that if there's an error, it means that Google Pay is not supported (but i still have to test this).
Then:
`const pm = await createGooglePayPaymentMethod(nativePaymentParams.googlePay);`
And proceed as always on backend side. No needs to close or confirm the payment as Apple Pay requires, as it can be confirmed directly on backend.

Hope this helps ;)

Gabotron-ES:
Hi my friend, I will take a look at your code, looks very nice I must say!

arekkubaczkowski:
@Gabotron-ES answering to your question
> Hi everybody, currently I can't seem to find how to use Google Pay or Apple Pay with CardField component, is it currently possible?

No it's not possible, CardField/CardForm components are dedicated to processing classic card payments and not related to any wallets.

However, considering such solution you must be aware that collecting the data on client side has to meet [PCI Compilance](https://stripe.com/docs/security/guide#validating-pci-compliance) requirements and it's not recommended without any experience within this area.

[...]

Issue #607, opened 2021-09-23
google pay is not available on this device (stripe react native).


Mukthaharshaik:
**Describe the bug**
Whenever i call initGooglePay, It is saying - google play is not available on this device. I have installed googlepay in my mobile. I enabled gpay in manifestfile.xml. 

**To Reproduce**

import {useStripe,  initGooglePay,
  presentGooglePay,
  createGooglePayPaymentMethod, } from '@stripe/stripe-react-native';

===========================
const error = await initGooglePay({
      testEnv: true,
      merchantName: 'com.stripe_app',
      countryCode: 'IN',
      billingAddressConfig: {
        format: 'FULL',
        isPhoneNumberRequired: true,
        isRequired: false,
      },
      existingPaymentMethodRequired: false,
      isEmailRequired: true,
    });

**Smartphone (please complete the following information):**
 - Device: samung m30s
 - OS: [android 11

anija:
Using latest version of the library, initGooglePay works fine for me on phisical device: Huaweii p30, Android 10. Are you passing `testEnv: true` as option?

charliecruzan-stripe:
@Mukthaharshaik are you still seeing this issue?

sailes-coding:
any idea on how can i resolve this problem on Expo sdk 44??. i already create the EAS dev-client too

charliecruzan-stripe:
If you're using the dev client, then you can pull in `@stripe/stripe-react-native` v0.3.0 and check the `isGooglePaySupported` method

sailes-shakya:
After i pull latest version . min sdk and complilesdk 31 doesnt match with expo sdk 44 .any idea to solve this problem. Thank you in advance

charliecruzan-stripe:
Please refer to this issue- https://github.com/stripe/stripe-react-native/issues/812

You'll either have to override those values in `build.gradle`, or wait until Expo updates

sailes-shakya:
Oo ok thank you for your reply

zoalna:
i have faced same issues but not getting any proper solution

nadeem36212:
I was also facing this issue, after spending couple of hours, I am able to find the solution
https://developers.google.com/pay/api/android/guides/setup

PR #811, opened 2022-02-15
Feat: add `isGooglePaySupported` method

charliecruzan-stripe:
## Summary
Currently people need to `initGooglePay` and `catch` an error to find out if the device supports Google pay, which isn't great. By adding this method, we have parity with the native wallets on android and ios (which offers `isApplePaySupported`).

## Motivation

Closes https://github.com/stripe/stripe-react-native/issues/683 (popular feature request)

This PR also closes https://github.com/stripe/stripe-react-native/issues/607, and closes https://github.com/stripe/stripe-react-native/issues/617, and closes https://github.com/stripe/stripe-react-native/issues/725

## Testing

- [x] I tested this manually (on devices with and without google play services)
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
+++ b/android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
@@ -1,14 +1,11 @@
 package com.reactnativestripesdk
 
 import android.app.Activity
-import android.app.Application
 import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
 import android.content.IntentFilter
 import android.os.AsyncTask
-import android.os.Build
-import android.os.Bundle
 import android.os.Parcelable
 import android.util.Log
 import androidx.activity.ComponentActivity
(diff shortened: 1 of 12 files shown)


charliecruzan-stripe:
@brnunes-stripe just need the OK on the Android code, feel free to ignore the TS code!

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
```diff
@@ -604,6 +604,26 @@ class StripeSdkModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
     }
   }
 
+  @ReactMethod
+  fun isGooglePaySupported(params: ReadableMap?, promise: Promise) {
```
brnunes-stripe, line 608: Is it possible to require an Activity or Fragment at this point? If so, can you use [GooglePayPaymentMethodLauncher](https://github.com/stripe/stripe-android/blob/master/payments-core/src/main/java/com/stripe/android/googlepaylauncher/GooglePayPaymentMethodLauncher.kt#L113-L154)? After you initialize it it'll make the same request and [call the `readyCallback`](https://github.com/stripe/stripe-android/blob/master/payments-core/src/main/java/com/stripe/android/googlepaylauncher/GooglePayPaymentMethodLauncher.kt#L203-L208) with the result. That way we don't need to add a dependency to `play-services-wallet` and have a single implementation of this check.

brnunes-stripe (changes requested):

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
charliecruzan-stripe: Good idea, most recent commits change to using a Fragment w/ `GooglePayPaymentMethodLauncher`

brnunes-stripe: Nice, thank you!

brnunes-stripe (approved):

PR #814, opened 2022-02-16
fix: incorrect payment sheet shown

### LINKED ISSUE #608: Incorrect payment sheet shown
**Describe the bug**

We are trying to update the package for a project so I'm bumping from `"@stripe/stripe-react-native": "^0.2.0"` to `"stripe/stripe-react-native": "^0.2.2"`. Especially, because of the improvements that were released in `0.2.1`.

But the new way of using `presentPaymentSheet`, introduced in #514, is causing issues when using multiple different `PaymentSheets` throughout the app. We use the `PaymentSheet` in 3 different ways:

 1.  ´setupIntentClientSecret´ and custom flow
 2.  ´setupIntentClientSecret´ and no custom flow
 3.  `paymentIntentClientSecret` and no custom flow

This works nicely with version `0.2.0`. The issue we get with version `0.2.1` or higher is, that after opening the sheet for case `1`(see list above), the same sheet opens for case `2` and `3`. After relaunch the correct sheet is shown, but only if you use case `2` or `3` first.

**To Reproduce**
1. use the PaymentSheet for setup in a custom flow(works as intended): 

```
await initPaymentSheet({
      setupIntentClientSecret,
      customerEphemeralKeySecret,
      customerId,
      customFlow: true,
      merchantCountryCode: MERCHANT_COUNTRY_CODE,
      style: 'automatic',
      testEnv: true,
});

// later present the sheet on click on the same screen
await presentPaymentSheet();
```

2. use the PaymentSheet for a purchase and without a custom flow, on a different screen (same session):

```
await initPaymentSheet({
      paymentIntentClientSecret,
      customerEphemeralKeySecret,
      customerId,
      customFlow: false,
      applePay: true,
      googlePay: true,
      merchantCountryCode: MERCHANT_COUNTRY_CODE,
      style: 'automatic',
      testEnv: true,
});

// on "Pay"-button click
await presentPaymentSheet();
```

3. See that the "Pay" button is missing the PaymentSheet and the title says: "Select your payment method"

**Expected behavior**
The intended `PaymentSheet` is presented even when using multiple different sheets during one session.

**Screenshots**

 - Device: iPhone 12
 - OS: iOS14.5 and iOS15

lbinscheck:
## Summary

On iOS, the "custom-flow"-PaymentSheet is presented also for "non-custom-flow"-PaymentSheets after it has been opened once during the session. In order to fix this bug, I added `self.paymentSheetFlowController = nil` in `initPaymentSheet` when `customFlow == false` in order to open the correct PaymentSheet in `presentPaymentSheet` later on.

## Motivation

Fixes #608 

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/ios/StripeSdk.swift
+++ b/ios/StripeSdk.swift
@@ -156,6 +156,7 @@ class StripeSdk: RCTEventEmitter, STPApplePayContextDelegate, STPBankSelectionVi
                 }
             } else {
                 self.paymentSheet = PaymentSheet(paymentIntentClientSecret: paymentIntentClientSecret, configuration: configuration)
+                self.paymentSheetFlowController = nil
                 resolve([])
             }
         } else if let setupIntentClientSecret = params["setupIntentClientSecret"] as? String {
@@ -171,6 +172,7 @@ class StripeSdk: RCTEventEmitter, STPApplePayContextDelegate, STPBankSelectionVi
                 }
             } else {
                 self.paymentSheet = PaymentSheet(setupIntentClientSecret: setupIntentClientSecret, configuration: configuration)
+                self.paymentSheetFlowController = nil
                 resolve([])
             }
         } else {
--- a/ios/StripeSdk.swift
+++ b/ios/StripeSdk.swift
@@ -65,6 +65,7 @@ class StripeSdk: RCTEventEmitter, STPApplePayContextDelegate, STPBankSelectionVi
     func initPaymentSheet(params: NSDictionary, resolver resolve: @escaping RCTPromiseResolveBlock,
                           rejecter reject: @escaping RCTPromiseRejectBlock) -> Void  {
         var configuration = PaymentSheet.Configuration()
+        self.paymentSheetFlowController = nil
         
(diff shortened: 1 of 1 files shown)


ios/StripeSdk.swift
```diff
@@ -156,6 +156,7 @@ class StripeSdk: RCTEventEmitter, STPApplePayContextDelegate, STPBankSelectionVi
                 }
             } else {
                 self.paymentSheet = PaymentSheet(paymentIntentClientSecret: paymentIntentClientSecret, configuration: configuration)
+                self.paymentSheetFlowController = nil
```
charliecruzan-stripe, line 159: thanks for this PR! 

I'm thinking it might be cleaner to call `self.paymentSheetFlowController = nil` just once at the top of the `initPaymentSheet` function, what do you think?

lbinscheck: Yeah, that's is much better. Thanks!

charliecruzan-stripe (approved):

PR #821, opened 2022-02-18
feat: add Klarna to `confirmPayment` payment methods

charliecruzan-stripe:
## Summary
Add the bindings for users to pay with Klarna programmatically

## Motivation
Klarna was supported on paymentsheet, but there were no bindings so folks couldn't use it programatically

## Testing

- [x] I tested this manually
- [x] I added automated tests

## Documentation

Looks like there used to be a Klarna docs page, but it was removed? I'm not sure why. Typescript will auto-doc that `confirmPayment` now accepts Klarna


charliecruzan-stripe:
requesting reviews from android/ios run cc @csabol-stripe @skyler-stripe whenever you get a chance to take a peak 😄

skyler-stripe (approved):
LGTM. I'm not good at react but the logic for the requirements makes sense.

example/server/index.ts
```diff
@@ -519,9 +519,33 @@ app.post('/payment-sheet', async (_, res) => {
...
+      name: 'Jane Doe',
+      address: {
+        state: 'Texas',
+        city: 'Houston',
+        line1: '1459  Circle Drive',
+        postal_code: '77063',
+        country: 'US',
+      },
+    },
+    // Edit the following to support different payment methods in your PaymentSheet
```
yuki-stripe, line 535: Interesting, looks like this is user-facing.  I wonder why we aren't using the example server(s) the iOS and Android platforms use (you can see them here: https://glitch.com/@stripe-mobile-payments)

yuki-stripe (approved):

example/server/index.ts
charliecruzan-stripe: good question! I'm honestly not sure why this approach was opted for instead...on CI, we rely on our glitch server rather than the local one here. May have just been added before glitch was considered and is still around. Suppose we could probably get rid of this server altogether from the repo

ios/PaymentMethodFactory.swift
```diff
@@ -344,6 +359,8 @@ extension PaymentMethodError: LocalizedError {
             return NSLocalizedString("You must provide CVC number", comment: "Create payment error")
         case .weChatPayPaymentMissingParams:
             return NSLocalizedString("You must provide appId", comment: "Create payment error")
+        case .klarnaPaymentMissingParams:
+            return NSLocalizedString("Klarna requires that you provide the following billing details: email, country", comment: "Create payment error")
```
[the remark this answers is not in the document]
charliecruzan-stripe: Yeah tbh, I was following the pattern here 😅 

It looks like the error message we get back from the server is essentially the same: 

Screen Shot 2022-03-03 at 3 37 02 PM


I'm okay with simply letting the server error bubble up from here. Can delete a lot of code if we do that for all the payment methods

charliecruzan-stripe:
Created tickets for both of those points, merging this now though

PR #824, opened 2022-02-22
Fix/google pay error code

### LINKED ISSUE #704: Cannot differentiate Google Pay cancel error
**Device:**

Xiaomi Redmi Note 10

**Stripe version:**

0.2.3

**Describe the bug:**

After calling presentGooglePay, if the user cancel the flow by pressing the close (X) icon, this method will throw an error like this:

`{"code": "Failed", "declineCode": null, "localizedMessage": "Google Pay has been canceled", "message": "Google Pay has been canceled", "stripeErrorCode": null, "type": null}`

We need a way to differentiate normal errors from the cancel one. Of course there are workarounds like checking the `localizedMessage` value, but is not the best approach at all

arekkubaczkowski:
closes #704


charliecruzan-stripe (approved):
👍 

Test failures are unrelated, probably just some flakiness

PR #832, opened 2022-03-04
fix: Native code audit, mostly Android

charliecruzan-stripe:
## Summary / Motivation

Get rid of warnings, fix some code styling, better `null` checks, remove usage of deprecated Android features, better file naming, use `PaymentLauncher` instead of `stripe.onPaymentResult`

For reviewing- it's probably most efficient to review this PR commit by commit

## Testing

- [x] I tested this manually
- Existing automated tests should pass

## Documentation

Select one: 


charliecruzan-stripe:
@michelleb-stripe requesting your review since you did the Android audit ([here](https://paper.dropbox.com/doc/React-native-Library-Review--BdCTZAdYDo~RJG97vYw1Zed4Ag-VeQmK2tx7MHta9xKYIf94)). Really the only commit that I think could use your attention is [[this one]](https://github.com/stripe/stripe-react-native/pull/832/commits/8040319bda5b286028c5b8b9f44d118fbf6b61e7)

android/src/main/java/com/reactnativestripesdk/PaymentLauncherFragment.kt
```diff
@@ -0,0 +1,115 @@
...
+      it.visibility = View.GONE
+    }
+  }
+
+  private fun createPaymentLauncher(): PaymentLauncher {
+    return PaymentLauncher.create(this, publishableKey, stripeAccountId) { paymentResult ->
+      when (paymentResult) {
+        is PaymentResult.Completed -> {
+          clientSecret?.let {
+            retrievePaymentIntent(it, stripeAccountId)
```
michelleb-stripe, line 40: How is the result of of retrievePaymentIntent being used?

CHANGELOG.md
```diff
@@ -3,7 +3,12 @@
 ## Unreleased
 
 - [#630](https://github.com/stripe/stripe-react-native/pull/630) Fix: card field focus jumps back to the card number field on Android
-- [#824](https://github.com/stripe/stripe-react-native/pull/824) Fix: Google Pay error code when canceled should be 'Canceled' 
+- [#824](https://github.com/stripe/stripe-react-native/pull/824) Fix: Google Pay error code when canceled should be 'Canceled'
+- [#832](https://github.com/stripe/stripe-react-native/pull/832)
+  - BREAKING CHANGE: Rename `handleCardAction` to `handleNextAction`.
```
michelleb-stripe, line 8: For this I wonder if you can change the handleCardAction to be deprecated and in the annotation suggest to use handleNextAction, and then in a later release remove handleCardAction.  This will give users a little opportunity to update to the new method.

android/src/main/java/com/reactnativestripesdk/CardFormView.kt
```diff
@@ -46,27 +45,28 @@ class CardFormView(private val context: ThemedReactContext) : FrameLayout(contex
...
-      multilineWidgetBinding.tlExpiry.hint = it
-    }
-    cvcPlaceholder?.let {
-      multilineWidgetBinding.tlCvc.hint = it
-    }
-    postalCodePlaceholder?.let {
-      cardFormView.postalCodeContainer.hint = it
-    }
-  }
+  // TODO: uncomment when ios-sdk allows for this
```
michelleb-stripe, line 48: Will this be a change of behavior that should be documented in the changelog?

android/src/main/java/com/reactnativestripesdk/Mappers.kt
```diff
@@ -366,7 +366,7 @@ internal fun mapFromPaymentIntentResult(paymentIntent: PaymentIntent): WritableM
   paymentIntent.amount?.let {
     map.putDouble("amount", it.toDouble())
   }
-  paymentIntent.canceledAt?.let {
+  paymentIntent.canceledAt.let {
     map.putString("canceledAt", convertToUnixTimestamp(it))
   }
```
michelleb-stripe, line 371: If canceledAt is not null, you no longer need the let.
```suggestion
  map.putString("canceledAt", convertToUnixTimestamp(paymentIntent.canceledAt))
```

android/src/main/java/com/reactnativestripesdk/PaymentMethodCreateParamsFactory.kt
```diff
@@ -145,21 +145,22 @@ class PaymentMethodCreateParamsFactory(
...
+      val iban = getValOr(params, "iban", null) ?: run {
+        throw PaymentMethodCreateParamsException("You must provide IBAN")
+      }
 
-    val sepaParams = PaymentMethodCreateParams.SepaDebit(iban)
-    val createParams =
-      PaymentMethodCreateParams.create(sepaDebit = sepaParams, billingDetails = billingDetails)
+      val sepaParams = PaymentMethodCreateParams.SepaDebit(iban)
+      val createParams =
+        PaymentMethodCreateParams.create(sepaDebit = sepaParams, billingDetails = it)
```
michelleb-stripe, line 155: nit: below we put each of the parameters on different lines, just for styling it might be nice to do the same here
```suggestion
        PaymentMethodCreateParams.create(
                     sepaDebit = sepaParams, 
                     billingDetails = it
        )
```

android/src/main/java/com/reactnativestripesdk/PaymentMethodCreateParamsFactory.kt
```diff
@@ -145,21 +145,22 @@ class PaymentMethodCreateParamsFactory(
...
   private fun createSepaPaymentSetupParams(): ConfirmSetupIntentParams {
-    val billingDetails = billingDetailsParams?.let { it } ?: run {
-      throw PaymentMethodCreateParamsException("You must provide billing details")
-    }
-    val iban = getValOr(params, "iban", null)?.let { it } ?: run {
-      throw PaymentMethodCreateParamsException("You must provide IBAN")
-    }
+    billingDetailsParams?.let {
+      val iban = getValOr(params, "iban", null) ?: run {
+        throw PaymentMethodCreateParamsException("You must provide IBAN")
```
michelleb-stripe, line 150: In the future we should consider how we localize these error strings for different languages.

android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt
```diff
@@ -94,9 +90,9 @@ class PaymentSheetFragment : Fragment() {
                                           addressBundle?.getString("postalCode"),
                                           addressBundle?.getString("state"))
       defaultBillingDetails = PaymentSheet.BillingDetails(address,
-                                                          billingDetailsBundle?.getString("email"),
-                                                          billingDetailsBundle?.getString("name"),
-                                                          billingDetailsBundle?.getString("phone"))
+                                                          billingDetailsBundle.getString("email"),
```
michelleb-stripe, line 93: nit: I have kinda like the styling of all parameters on a new line, then they also alight closer to the left.

android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt
```diff
@@ -142,20 +138,18 @@ class PaymentSheetFragment : Fragment() {
...
-
-          intent.putExtra("label", paymentOption.label)
-          intent.putExtra("image", imageString)
-        }
-        localBroadcastManager.sendBroadcast(intent)
+    val onFlowControllerConfigure = PaymentSheet.FlowController.ConfigCallback { _, _ ->
+      val paymentOption = flowController?.getPaymentOption()
+      val intent = Intent(ON_CONFIGURE_FLOW_CONTROLLER)
+
+      if (paymentOption != null) {
```
michelleb-stripe, line 145:
```suggestion
      paymentOption?.let {
```

android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt
```diff
@@ -142,20 +138,18 @@ class PaymentSheetFragment : Fragment() {
...
+        val bitmap = getBitmapFromVectorDrawable(context, paymentOption.drawableResourceId)
+        val imageString = getBase64FromBitmap(bitmap)
+
+        intent.putExtra("label", paymentOption.label)
+        intent.putExtra("image", imageString)
```
michelleb-stripe, line 150:
```suggestion
      paymentOption?.let {
        val bitmap = getBitmapFromVectorDrawable(context, it.drawableResourceId)
        val imageString = getBase64FromBitmap(bitmap)

        intent.putExtra("label", it.label)
        intent.putExtra("image", imageString)
```

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
```diff
@@ -271,6 +215,7 @@ class StripeSdkModule(private val reactContext: ReactApplicationContext) : React
   }
 
   @ReactMethod
+  @SuppressWarnings("unused")
   fun initialise(params: ReadableMap, promise: Promise) {
     val publishableKey = getValOr(params, "publishableKey", null) as String
     val appInfo = getMapOrNull(params, "appInfo") as ReadableMap
```
michelleb-stripe, line 221: On line 278: setUrlSchemeOnAndroid, it appears this is the returnUrl on Android.  I wonder if we can deprecate setUrlSchemeOnAndroid and move to setReturnUrlSchemeOnAndroid?

Let's check with @ccen-stripe, but I think we recommend that user's don't set this, but not sure where we indicate that to users.  ([Stripe.kt](https://github.com/stripe/stripe-android/blob/3f632b03d26853077d47ba80ecf6123664793d0c/payments-core/src/main/java/com/stripe/android/Stripe.kt#L251) has some commentary on the effect of the returnUrl, but doesn't say anything about recommending it is not set.)

CHANGELOG.md
charliecruzan-stripe: That's a good point, I was hoping to wrap a couple different breaking changes into our v1 release, but it's def a better user experience to make this less jarring.

android/src/main/java/com/reactnativestripesdk/CardFormView.kt
charliecruzan-stripe: Nope, the ReactMethod that calls into this was already commented out so it's basically dead code: https://github.com/stripe/stripe-react-native/pull/832/files#diff-30505d89bcbbe719ef37810e46ecf58b541642866e602a43f39f36804e0a3ee3R41-R44

android/src/main/java/com/reactnativestripesdk/PaymentMethodCreateParamsFactory.kt
charliecruzan-stripe: I would have believed these are for developers, but it looks like the iOS equivalent messages are localized -> https://github.com/stripe/stripe-react-native/blob/master/ios/PaymentMethodFactory.swift#L338-L363

Either way, the server returns a usefully formatted and clear error message in these cases, so we'll probably remove a lot of these custom error messages- https://github.com/stripe/stripe-react-native/pull/821#discussion_r819045389

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
```diff
@@ -486,53 +441,55 @@ class StripeSdkModule(private val reactContext: ReactApplicationContext) : React
...
-            clientSecret = paymentIntentClientSecret,
-            paymentMethodOptions = PaymentMethodOptionsParams.WeChatPay(
-              appId
-            )
-          )
-        stripe.confirmPayment(activity, confirmPaymentIntentParams)
-      }
-    }
-  }
+// TODO: Uncomment when WeChat is re-enabled in stripe-ios
```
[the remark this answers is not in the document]
charliecruzan-stripe: nope, we're okay here since the only reference to this method is commented out- https://github.com/stripe/stripe-react-native/blob/master/android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt#L545

android/src/main/java/com/reactnativestripesdk/PaymentLauncherFragment.kt
charliecruzan-stripe: `retrievePaymentIntent` allows us to get the relevant `paymentIntent` (or the error) and pass it back to the React Native app via `promise.resolve`. Went with `stripe.retrievePaymentIntent` based on [this comment](https://github.com/stripe/stripe-android/issues/2640#issuecomment-929682178)

android/src/main/java/com/reactnativestripesdk/CardFormView.kt
michelleb-stripe: ok, great

android/src/main/java/com/reactnativestripesdk/PaymentLauncherFragment.kt
michelleb-stripe: I am not sure I am understanding.  It looks like the retrievePaymentIntent is only be retrieved in the "Complete" case and isn't being returned.

android/src/main/java/com/reactnativestripesdk/PaymentMethodCreateParamsFactory.kt
michelleb-stripe: sounds good

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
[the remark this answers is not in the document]
michelleb-stripe: cool, thanks

android/src/main/java/com/reactnativestripesdk/PaymentLauncherFragment.kt
charliecruzan-stripe: That's correct- `retrievePaymentIntent` is only called in the complete case. In the case of cancellation or failure, we send an error object back to the user (L46 and L50)

In the case of a completed payment result, we retrieve the payment intent, and `resolve` the promise that was passed in via Javascript. We aren't explicitly calling `return` here with the payment intent object, but the `promise.resolve` call is passing the data back over the React Native bridge into "Javascript land". 

The previous code functioned the same way, [see here](https://github.com/stripe/stripe-react-native/blob/master/android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt#L82-L135). But, if there's a better way to fetch the payment intent associated with the result then let me know!

michelleb-stripe:
I just sent @ccen-stripe a message this morning to get some feedback on one remaining comment: https://github.com/stripe/stripe-react-native/pull/832/files#r822856777

android/src/main/java/com/reactnativestripesdk/PaymentLauncherFragment.kt
ccen-stripe: `PaymentLauncher` actually [processes](https://github.com/stripe/stripe-android/blob/262885d0c02a54430ae8f0c3966de694abc6bb0a/payments-core/src/main/java/com/stripe/android/StripeIntentResult.kt#L33) the status of the Intent internally and [returns](https://github.com/stripe/stripe-android/blob/d8460ee2f2c554df245cf7a0ca92d76be6195bca/payments-core/src/main/java/com/stripe/android/payments/paymentlauncher/PaymentLauncherViewModel.kt#L222) `Completed`/`Failed`/`Cancelled` according to the processed `Outcome`s.

Compared the logic with `retrievePaymentIntent` a bit, it seems we can just save the retrievePaymentIntent all and do the following mapping:
```
PaymentResult.Completed-> {
  val paymentIntent = createResult("paymentIntent", mapFromPaymentIntentResult(result))
  promise.resolve(paymentIntent)}
PaymentResult.Failed-> {
  val error = result.lastPaymentError
  promise.resolve(createError(ConfirmPaymentErrorType.Failed.toString(), error))
}
PaymentResult.Cancelled-> {
if (isPaymentIntentNextActionVoucherBased(result.nextActionType)) {
    val paymentIntent = createResult("paymentIntent", mapFromPaymentIntentResult(result))
    promise.resolve(paymentIntent)
  } else {
    (result.lastPaymentError)?.let {
      promise.resolve(createError(ConfirmPaymentErrorType.Canceled.toString(), it))
    } ?: run {
      promise.resolve(createError(ConfirmPaymentErrorType.Canceled.toString(), "The payment has been canceled"))
    }
  }
}
```

charliecruzan-stripe: Not sure I understand- `mapFromPaymentIntentResult` accepts a `PaymentIntent`, but the `result` we're getting here from `PaymentLauncher` is a `PaymentResult`. I thought in order to retrieve the full `PaymentIntent` object, we pretty much needed to use `Stripe.retrievePaymentIntent`

michelleb-stripe (approved):

PR #835, opened 2022-03-08
fix: only call requestLayout on focus change

charliecruzan-stripe:
## Summary & Motivation

Previously, we were calling `requestLayout` whenever the global layout state or the visibility of views within the view tree changed. Now, we only call it when we need to (when focus changes)

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 


charliecruzan-stripe:
Closing since it's probably better to keep the call where it is even though it's overkill. This way, any layout change will be accounted for rather than us having to manually add another call to address whatever need comes up in the future.

Issue #844, opened 2022-03-15
initGooglePay have the same init error that INIT_PAYMENTSHEET had


carl0shd:
**Describe the bug**
Hi Guys, I have multiple errors (+80 daily) in my production app using Stripe, when I call `initGooglePay` at the launch in my app, I have this error:

`java.lang.NullPointerException: null cannot be cast to non-null type androidx.appcompat.app.AppCompatActivity`

I think is the same error that the `INIT_PAYMENTSHEET` have in old versions: https://github.com/stripe/stripe-react-native/pull/783

This is the full log:

<img width="859" alt="image" src="https://user-images.githubusercontent.com/8359234/158400986-416a720f-cfa1-4624-b3d6-cbc6cb21dc91.png">

apparently, the error doesn't always occur, it seems to be random.

**To Reproduce**
N/A, I don't have instructions to reproduce this :(

**Expected behavior**
Not have this crash.

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

Stripe RN version: 0.2.4
RN Version: 0.66

carl0shd:
Hi @charliecruzan-stripe, actually the app uses version 0.2.4

charliecruzan-stripe:
thanks :) fix incoming, it's extremely helpful that you included the crash logs in your issue report 👏

goguda:
![Screenshot_20220408_143750](https://user-images.githubusercontent.com/26887502/162501761-4a17c156-68ae-4c98-9566-c686cf16e3f0.png)

Same issue seems to occur at line 244 in StripeSdkModule.kt as well... null check needed here too?

I find it strange because based on visit logs I believe this is occurring without the user even interacting with any Stripe APIs yet, other than StripeProvider wrapping the root of the app.

Looking at line 244 of StripeSdkModule.kt...

`(currentActivity as AppCompatActivity).supportFragmentManager.beginTransaction().add(paymentLauncherFragment, "payment_launcher_fragment").commit()`

Taking a quick look at the internals of the library, my concern is, is a null check enough, or are null checks really just masking another problem? Should currentActivity ever actually be null in this case if the RN app has to be open for any of these API calls to even occur (as far as I know)?

It also doesn't seem to be happening to every user, but I am having a few Android users per day trigger this error on my side in production. I cannot seem to reproduce it myself, sadly. It seems to happen across a wide range of Android devices. All my errors have interestingly happened on Android 10 or Android 11:

[...]

PR #846, opened 2022-03-15
fix: add check to avoid crashes when currentActivity is null

charliecruzan-stripe:
## Summary

We weren't always appropriately null-checking `currentActivity`, and the typecast could result in crashes. We now resolve with a consistent and useful error telling the user they can retry the method.

## Motivation
closes https://github.com/stripe/stripe-react-native/issues/844 and I'm sure other instances of crashes on Android
closes https://github.com/stripe/stripe-react-native/issues/881

## Testing

- [ ] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
+++ b/android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
@@ -50,9 +50,6 @@ class StripeSdkModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
   private var initGooglePayPromise: Promise? = null
   private var presentGooglePayPromise: Promise? = null
 
-  private val currentActivity: AppCompatActivity?
-    get() = (super.getCurrentActivity() as? AppCompatActivity)
-
   private val mActivityEventListener = object : BaseActivityEventListener() {
     override fun onActivityResult(activity: Activity, requestCode: Int, resultCode: Int, data: Intent?) {
       if (::stripe.isInitialized) {
@@ -239,7 +236,7 @@ class StripeSdkModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
   @ReactMethod
   @SuppressWarnings("unused")
   fun initPaymentSheet(params: ReadableMap, promise: Promise) {
-    currentActivity?.let { activity ->
+    getCurrentActivityOrResolveWithError(promise)?.let { activity ->
       this.initPaymentSheetPromise = promise
 
       paymentSheetFragment = PaymentSheetFragment().also {
@@ -249,8 +246,6 @@ class StripeSdkModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
       activity.supportFragmentManager.beginTransaction()
         .add(paymentSheetFragment!!, "payment_sheet_launch_fragment")
         .commit()
-    } ?: run {
-      promise.resolve(createMissingActivityError())
     }
   }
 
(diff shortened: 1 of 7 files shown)


android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
```diff
@@ -595,41 +607,52 @@ class StripeSdkModule(private val reactContext: ReactApplicationContext) : React
     val factory = PaymentMethodCreateParamsFactory(setupIntentClientSecret, params, cardFieldView, cardFormView)
 
     try {
-      val activity = currentActivity as ComponentActivity
       val confirmParams = factory.createSetupParams(paymentMethodType)
-      stripe.confirmSetupIntent(activity, confirmParams)
+      (currentActivity as? ComponentActivity)?.let {
+        stripe.confirmSetupIntent(it, confirmParams)
+      } ?: run {
+        promise.resolve(createMissingActivityError())
```
fbartho, line 614: When is it possible for currentActivity to be `null`?

fbartho: Please forgive my lack of experience with Kotlin code, is `currentActivity` a dynamic getter here where it might become non-null on future invocations? or is `currentActivity` a member variable that is captured by the React-Native-Module? (If it's the latter, then retrying wouldn't fix the problem)

Additionally, it feels like _very early_ in startup is not an expected time to call the Stripe SDK

charliecruzan-stripe: it's dynamic, react modules should NOT have any long lived reference to activities

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
```diff
@@ -172,7 +172,9 @@ class StripeSdkModule(private val reactContext: ReactApplicationContext) : React
   private val googlePayReceiver: BroadcastReceiver = object : BroadcastReceiver() {
     override fun onReceive(context: Context?, intent: Intent) {
       if (intent.action == ON_GOOGLE_PAY_FRAGMENT_CREATED) {
-        googlePayFragment = (currentActivity as AppCompatActivity).supportFragmentManager.findFragmentByTag("google_pay_launch_fragment") as GooglePayFragment
+        (currentActivity as? AppCompatActivity)?.let {
```
michelleb-stripe, line 175: I was really surprised that you have to check the current activity for null everywhere, even from @ReactMethod, but I didn't really see another way.

In order to prevent this type of error in the future I would recommend:
```
class StripeSdkModule ...{
...
  // Set currentActivity so it uses this instead of the base class `getCurrentActivity()`
  // this will give you type safety checking and convert it to an appCompat activity
  val currentActivity: AppCompatActivity?
    get() = (super.getCurrentActivity() as? AppCompatActivity)
...
    private val googlePayReceiver: BroadcastReceiver = object : BroadcastReceiver() {
        override fun onReceive(context: Context?, intent: Intent) {
          if (intent.action == ON_GOOGLE_PAY_FRAGMENT_CREATED) {
            currentActivity?.let {
...
}
```


It would also be nice to see some enforcement in the code that if you can't get the appCompat activity that a promise is fired to return a result.

michelleb-stripe (changes requested):

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
charliecruzan-stripe: thanks for the feedback! created a helper method for getting the activity or resolving the promise. It's not the perfect solution since someone can still call `currentActivity`, but it does avoid a lot of repeated code and hopefully anyone working here would see the current pattern for getting the activity 🤞

charliecruzan-stripe:
closes https://github.com/stripe/stripe-react-native/issues/881

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
charliecruzan-stripe: @michelleb-stripe do you approve the most recent changes? I'll need to rebase but since https://github.com/stripe/stripe-react-native/pull/879 contains substantial changes to the same file, i'd rather wait till after that lands to rebase (to avoid duplicating the work)

michelleb-stripe (approved):

PR #861, opened 2022-03-24
feat: Add ACHv2 support for iOS

charliecruzan-stripe:
## Summary

This exposes 4 methods that React Native devs can use:

- `collectBankAccountForPayment`: opens a modal for customers to provide their US Bank account info and attach it to the specified payment intent
- `collectBankAccountForSetup`: opens a modal for customers to provide their US Bank account info and attach it to the specified setup intent
- `verifyMicrodepositsForPayment`: verifies microdeposits for a payment intent
- `verifyMicrodepositsForSetup`: verifies microdeposits for a setup intent

These all match methods exposed in stripe.js (in naming and parameters):

- https://stripe.com/docs/js/payment_intents/collect_bank_account_for_payment
- https://stripe.com/docs/js/setup_intents/collect_bank_account_for_setup
- https://stripe.com/docs/js/payment_intents/verify_microdeposits_for_payment
- https://stripe.com/docs/js/setup_intents/verify_microdeposits_for_setup

## Motivation

add support for ACHv2 via bindings on iOS since GA was yesterday,

## Testing

- [x] I tested this manually
- [x] I added automated tests

Tests are failing because Webdriver isn't playing nicely with macos-latest on GHA, but we need to upgrade to support the newer version of stripe-ios :/ will look into that, but the tests are passing locally on my machine.

## Documentation

Select one: 
- [x] I have added relevant documentation for my changes


android/build.gradle
```diff
@@ -132,7 +132,7 @@ dependencies {
   api 'com.facebook.react:react-native:+'
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
   implementation "androidx.lifecycle:lifecycle-runtime-ktx:2.3.1"
-  implementation 'com.stripe:stripe-android:19.3.+'
+  implementation 'com.stripe:stripe-android:19.3.1'
```
charliecruzan-stripe, line 135: reverting this

android/src/main/java/com/reactnativestripesdk/Mappers.kt
```diff
@@ -311,6 +348,22 @@ internal fun mapFromPaymentMethod(paymentMethod: PaymentMethod): WritableMap {
 
   upi.putString("vpa", paymentMethod.upi?.vpa)
 
+  // USING REFLECTION, DELETE BEFORE MERGING (ONCE IT IS PUBLIC)
+  val usBankAccountIndex = PaymentMethod::class.java.getDeclaredField("usBankAccount")
+  usBankAccountIndex.isAccessible = true
+  val bankAccountPM = usBankAccountIndex.get(paymentMethod) as PaymentMethod.USBankAccount
```
charliecruzan-stripe, line 354: temporary while API is private

android/src/main/java/com/reactnativestripesdk/PaymentLauncherFragment.kt
```diff
@@ -70,16 +70,8 @@ class PaymentLauncherFragment(
...
-            if (isPaymentIntentNextActionVoucherBased(result.nextActionType)) {
-              val paymentIntent = createResult("paymentIntent", mapFromPaymentIntentResult(result))
-              promise.resolve(paymentIntent)
-            } else {
-              (result.lastPaymentError)?.let {
-                promise.resolve(createError(ConfirmPaymentErrorType.Canceled.toString(), it))
-              } ?: run {
-                promise.resolve(createError(ConfirmPaymentErrorType.Canceled.toString(), "The payment has been canceled"))
-              }
-            }
```
charliecruzan-stripe, line 82: I'm not sure why this was returning a `Canceled` error before, but that seems wrong to me. We're now returning a map containing `nextAction` data, so I hope this is safe but unfortunately don't have context on it (no commit history on this)

example/android/app/build.gradle
```diff
@@ -211,7 +211,7 @@ dependencies {
     implementation "com.facebook.react:react-native:+"  // From node_modules
     implementation "androidx.swiperefreshlayout:swiperefreshlayout:1.0.0"
     implementation 'com.google.android.material:material:1.4.0'
-    implementation 'com.stripe:stripe-wechatpay:17.1.0'
+    // implementation 'com.stripe:stripe-wechatpay:17.1.0' // UNCOMMENT
```
charliecruzan-stripe, line 214: temporary

example/android/build.gradle
```diff
@@ -28,6 +28,7 @@ allprojects {
             // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm
             url("$rootDir/../node_modules/react-native/android")
         }
+        maven { url "https://jitpack.io" } // REMOVE
```
charliecruzan-stripe, line 31: temporary

example/ios/Podfile
```diff
@@ -11,6 +11,11 @@ target 'StripeSdkExample' do
     hermes_enabled: false
   )
   pod 'stripe-react-native', path: '../..'
+  # BEGIN REMOVE
+  pod 'Stripe', :git => 'https://github.com/stripe-ios/stripe-ios.git', :branch => 'private'
+  pod 'StripeUICore', :git => 'https://github.com/stripe-ios/stripe-ios.git', :branch => 'private'
+  pod 'StripeCore', :git => 'https://github.com/stripe-ios/stripe-ios.git', :branch => 'private'
+  # END REMOVE
```
charliecruzan-stripe, line 18: temporary

example/android/app/build.gradle
davidme-stripe: We probably shouldn't be shipping WeChat Pay anyway, it's disabled for the Mobile SDKs afaik

ios/StripeSdk.swift
```diff
@@ -778,6 +778,80 @@ class StripeSdk: RCTEventEmitter, STPApplePayContextDelegate, STPBankSelectionVi
...
+        let descriptorCode = params["descriptorCode"] as? String
+        
+        if (amounts != nil && descriptorCode != nil || amounts == nil && descriptorCode == nil) {
+            resolve(Errors.createError("Failed", "You must provide either amounts OR descriptorCode, not both."))
+            return
+        }
+        
+        if let amounts = amounts {
+            if (amounts.count != 2) {
+                resolve(Errors.createError("Failed", "Expected 2 integers in the amounts array, but received " + String(amounts.count)))
```
charliecruzan-stripe, line 799: Will update these to not be just `"Failed"` strings and instead a Constant from Errors.swift

android/src/main/java/com/reactnativestripesdk/PaymentLauncherFragment.kt
davidme-stripe: I don't know what this is supposed to do either, and your change seems to make sense. In the iOS SDK, we don't appear to have any logic to handle RequiresAction + a voucher-based payment method differently.

davidme-stripe (approved):

android/src/main/java/com/reactnativestripesdk/PaymentLauncherFragment.kt
jameswoo-stripe: I don't have a lot of context on this either, but on Android and iOS we have the same thing: https://github.com/stripe/stripe-android/blob/master/payments-core/src/main/java/com/stripe/android/StripeIntentResult.kt#L34

We return success based on this check here: https://github.com/stripe/stripe-android/blob/master/payments-core/src/main/java/com/stripe/android/StripeIntentResult.kt#L71

I assume that we return canceled when we are in an invalid state for the payment intent type. Should we leave it as it was? I am not sure how we can verify this.

android/src/main/java/com/reactnativestripesdk/PaymentMethodCreateParamsFactory.kt
```diff
@@ -439,6 +454,45 @@ class PaymentMethodCreateParamsFactory(
...
+
+    return ConfirmPaymentIntentParams.createWithPaymentMethodCreateParams(
+      paymentMethodCreateParams = createUSBankAccountParams(params),
+      clientSecret,
+      setupFutureUsage = mapToPaymentIntentFutureUsage(getValOr(params, "setupFutureUsage"))
+    )
+  }
+
+  @Throws(PaymentMethodCreateParamsException::class)
+  private fun createUSBankAccountParams(params: ReadableMap): PaymentMethodCreateParams {
```
jameswoo-stripe, line 472: Technically when creating a `us_bank_account` payment method type, we only require the `type` and `name`.

android/src/main/java/com/reactnativestripesdk/PaymentLauncherFragment.kt
charliecruzan-stripe: @jameswoo-stripe that link is super helpful, looks like we're making the same check on iOS: https://github.com/stripe-ios/stripe-ios/blob/private/Stripe/STPPaymentHandler.swift#L201, and is set up with the same true/false cases. The solution here is to move from `isPaymentIntentNextActionVoucherBased` to something like `isPaymentIntentNextActionVoucherBased` to `isNextActionSuccessState`

charliecruzan-stripe:
Tests are failing because Webdriver isn't playing nicely with `macos-latest` on GHA, but we need to upgrade to support the newer version of `stripe-ios` :/ will look into that, but the tests _are_ passing locally on my machine.

ios/PaymentMethodFactory.swift
```diff
@@ -47,6 +47,8 @@ class PaymentMethodFactory {
                 return try createAfterpayClearpayPaymentMethodParams()
             case STPPaymentMethodType.klarna:
                 return try createKlarnaPaymentMethodParams()
+            case STPPaymentMethodType.USBankAccount:
+                return nil
```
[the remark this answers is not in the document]
charliecruzan-stripe: for the USBankAccount payment method type, the payment method is created/attached to the intent during `collectBankAccountForPayment`, so the `STPPaymentIntentParams.paymentMethodParams` aren't required in this case. Unless we want to allow folks to pass in account number, routing number, etc. directly (after collecting in their own UI, for instance)?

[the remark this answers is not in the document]
charliecruzan-stripe: Ah okay, thought that was a security issue and that was why we had a hosted modal for collecting bank account info. I can add that in easily then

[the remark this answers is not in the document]
charliecruzan-stripe: @csabol-stripe added in https://github.com/stripe/stripe-react-native/pull/861/commits/91150578a2636cbfab1a8a99dd0ff17d9345287e

Following this change, it's made me realize we should probably change the `confirmPayment` (and other similar methods) type to include two separate fields: payment method type and payment method data (right now it's just payment method data, which includes the type). That will make future additions like this one easier. That is out of scope for this PR though, but I will most likely land that change before our V1 release.

csabol-stripe (approved):

PR #875, opened 2022-03-31
fix: handle require payment method error on confirmSetupIntent on android

wijskinner:
## Summary
Handle RequiresPaymentMethod error from confirmSetupIntent so that error information matches iOS implementation

## Motivation
Whilst testing our switch from tipsi-stripe to this package, I noticed a difference in the error's returned from confirmSetupIntent on iOS vs Android. If a card is declined the error is now available on android and iOS for display to the user. Previous to this PR - instead of the actual error android returned `unhandled error: requires payment method`

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
+++ b/android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
@@ -66,6 +66,9 @@ class StripeSdkModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
               StripeIntent.Status.RequiresAction -> {
                 confirmSetupIntentPromise?.resolve(createError(ConfirmSetupIntentErrorType.Canceled.toString(), setupIntent.lastSetupError))
               }
+              StripeIntent.Status.RequiresPaymentMethod -> {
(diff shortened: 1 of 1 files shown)


charliecruzan-stripe:
Good catch and change! letting e2e tests run before merging

wijskinner:
@charliecruzan-stripe Anything else need to be done here? Would really appreciate getting this out in a release as soon as possible?

charliecruzan-stripe:
sorry @wijskinner , this passed e2e and completely slipped my mind! I overhauled our setup-intent handling logic as part of https://github.com/stripe/stripe-react-native/pull/879/files#diff-f694d33641390cc6897469c6942c0cefda04833cff587e7d337ae0403d7f2b5eR81-R120, so this will actually be addressed in there. We plan on releasing that linked PR in a week or so, does that work?

wijskinner:
No probs at all @charliecruzan-stripe. Yep your new setup-intent handling looks fine so will happily wait for that. Thanks for the quick response. I'll close this PR.

PR #879, opened 2022-04-07
feat: add support for ACHv2 on Android

charliecruzan-stripe:
## Summary

Added native implementations for 

- collectBankAccountForPayment
- collectBankAccountForSetup
- verifyMicrodepositsForPayment
- verifyMicrodepositsForSetup

Adds support for `USBankAccount` type to `confirmPayment` and `confirmSetupIntent`

Adds `com.stripe:connections:20.0.+` as a package dependency. We could also remove it and make this a user step, i don't have strong feelings either way

This PR also cleans up a lot of our [PaymentLauncherFragment.kt](https://github.com/stripe/stripe-react-native/pull/879/files#diff-f694d33641390cc6897469c6942c0cefda04833cff587e7d337ae0403d7f2b5e) code:
- added setup intent handling and removed that from StripeSdkModule.kt
- wrapped the `.confirm` and `.handleNextAction` methods so that it's harder for us to forget to assign the appropriate promise and client secret properties

Until Android officially releases ACHv2, I will re-comment the `if (isAndroid) {` Javascript checks before landing this PR.

## Motivation

Support in iOS was added in https://github.com/stripe/stripe-react-native/pull/861, this sets us up to be ready for Android's official release

## Testing

- [x] I tested this manually
- [x] I added automated tests

## Documentation

Select one: 


android/build.gradle
```diff
@@ -132,7 +132,8 @@ dependencies {
   api 'com.facebook.react:react-native:+'
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
   implementation "androidx.lifecycle:lifecycle-runtime-ktx:2.3.1"
-  implementation 'com.stripe:stripe-android:19.3.+'
+  implementation 'com.stripe:stripe-android:20.0.+'
+  implementation 'com.stripe:connections:20.0.+'
```
charliecruzan-stripe, line 136: Undecided on if this is included by default or we require users to add it

android/src/main/java/com/reactnativestripesdk/Mappers.kt
```diff
@@ -315,6 +350,28 @@ internal fun mapFromPaymentMethod(paymentMethod: PaymentMethod): WritableMap {
 
   upi.putString("vpa", paymentMethod.upi?.vpa)
 
+  // TODO: Remove reflection once USBankAccount is public payment method in stripe-android
+  try {
+    PaymentMethod::class.java.getDeclaredField("usBankAccount").let {
```
charliecruzan-stripe, line 355: reflection for now for testing, won't need this once feature is public

jameswoo-stripe (commented):
Just some concerns around translations

android/src/main/java/com/reactnativestripesdk/PaymentMethodCreateParamsFactory.kt
```diff
@@ -422,11 +424,31 @@ class PaymentMethodCreateParamsFactory(
...
+    params.getString("accountNumber")?.let {
+      if (billingDetailsParams?.name.isNullOrBlank()) {
+        throw PaymentMethodCreateParamsException("When creating a US bank account payment method, you must provide the following billing details: name")
+      }
+      return ConfirmSetupIntentParams.create(
+        paymentMethodCreateParams = createUSBankAccountParams(params),
+        clientSecret = clientSecret,
+      )
+    } ?: run {
+      // Payment method is assumed to be already attached through via collectBankAccountForSetup
```
carlosmuvi-stripe, line 438: @jameswoo-stripe I think it's fine, but can we double check we can make this assumption?

android/build.gradle
charliecruzan-stripe: yep- my question here is if the added size of the connections library is large enough to make users add this to their build.gradle manually. Even small additional steps like that can really cause issues for RN users. And then they're in charge of keeping that version up-to-date with the version of `com.stripe:stripe-android` we use is stripe-react-native internally. So I lean towards including it in our own library so users don't have to think about it.

android/src/main/java/com/reactnativestripesdk/CollectBankAccountLauncherFragment.kt
```diff
@@ -0,0 +1,83 @@
...
+    }
+  }
+
+  private fun createBankAccountLauncher(): CollectBankAccountLauncher {
+    return CollectBankAccountLauncher.create(this) { result ->
+      when (result) {
+        is CollectBankAccountResult.Completed -> {
+          val intent = result.response.intent
+          if (intent.status === StripeIntent.Status.RequiresPaymentMethod) {
+            promise.resolve(createError(ErrorType.Canceled.toString(), "Bank account collection was canceled."))
```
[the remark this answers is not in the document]
charliecruzan-stripe: This is developer facing

android/src/main/java/com/reactnativestripesdk/PaymentMethodCreateParamsFactory.kt
```diff
@@ -439,6 +461,60 @@ class PaymentMethodCreateParamsFactory(
         clientSecret = clientSecret,
       )
   }
+
+  @Throws(PaymentMethodCreateParamsException::class)
+  private fun createUSBankAccountPaymentConfirmParams(): ConfirmPaymentIntentParams {
+    params.getString("accountNumber")?.let {
+      if (billingDetailsParams?.name.isNullOrBlank()) {
+        throw PaymentMethodCreateParamsException("When creating a US bank account payment method, you must provide the following billing details: name")
```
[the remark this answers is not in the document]
charliecruzan-stripe: These are developer facing- the exception is caught and then we build a StripeError object which is then `resolve`d via the promise to Javascript. And then yeah, up to the developer to decide what action to take (i.e. show an error message or show some retry option)

jameswoo-stripe (approved):
lgtm

android/src/main/java/com/reactnativestripesdk/PaymentMethodCreateParamsFactory.kt
jameswoo-stripe: Went through the flow with @charliecruzan-stripe, it looks good to me

carlosmuvi-stripe (approved):
:shipit:

Issue #2, opened 2021-01-06
Add card provisioning bindings


bg-stripe:
Stripe Issuing lets users provision cards from apps to their Apple Pay / Google Pay wallets.

We'd like to add React Native bindings for the functionality in this integration guide:
https://stripe.com/docs/issuing/cards/digital-wallets

The iOS & Android integrations are slightly different, so it's possible the end result will need to start with shared content on setting up your EphemeralKeyProvider, and then split into two sections documenting how to set this up for Apple Pay vs. Google Pay. 

We'll want to verify that this works end-to-end in a sample app before we release it. We can give your Stripe account access to push provisioning (email bg [at] stripe.com and we can coordinate).

leekuo:
Hi Stripe team!  Are any plans to revive this effort and add the card provisioning bindings?  We're a React Native shop and this is also a blocker for us adding to Apple / Google Wallets.

PR #889, opened 2022-04-15
Feat: add support for push provisioning (adding cards to native wallets)

charliecruzan-stripe:
## Summary

- Adds an `<AddToNativeWallet />` component. When pressed, this button starts the push provisioning flow. success/failure is forwarded along to the `onComplete` callback
- Exposes a `Constants` object that will include that native libraries' api versions that are used, so that React Native users can have consistent ephemeral key results
- Adds an `isCardInWallet` method that devs can use to find if a card has already been added to the device's native wallet (if it has, both platforms recommend not showing the "add to X wallet" button

## Motivation

closes https://github.com/stripe/stripe-react-native/issues/2

https://jira.corp.stripe.com/browse/DX-7219

[API Proposal](https://paper.dropbox.com/doc/react-2022-03-11-React-Native-Push-Provisioning-Mobile-API-Review--Bfs3E9~5tkbyDPvEn4miXFy~Ag-fZfVfNJ09mluqwJBP5sUl)

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 
- [x] I have added relevant documentation for my changes. (these are in progress)


android/src/main/java/com/reactnativestripesdk/pushprovisioning/PushProvisioningProxy.kt
```diff
@@ -0,0 +1,109 @@
...
+import android.util.Log
+import com.facebook.react.bridge.*
+import com.reactnativestripesdk.pushprovisioning.AddToWalletButtonView
+import com.reactnativestripesdk.pushprovisioning.EphemeralKeyProvider
+import com.reactnativestripesdk.pushprovisioning.TapAndPayProxy
+import com.stripe.android.pushProvisioning.PushProvisioningActivity
+import com.stripe.android.pushProvisioning.PushProvisioningActivityStarter
+
+
+object PushProvisioningProxy {
```
charliecruzan-stripe, line 15: `com.stripe.android.pushProvisioning` is an optional dependency, so we proxy to it from this object. If it's not found, we send an error event to the onComplete` hook, but we don't crash. Basically- all methods will be a no-op without that dependency.

android/src/main/java/com/reactnativestripesdk/pushprovisioning/TapAndPayProxy.kt
```diff
@@ -0,0 +1,133 @@
...
+import android.app.Activity
+import android.util.Log
+import com.facebook.react.bridge.Promise
+import com.facebook.react.bridge.ReadableMap
+import com.facebook.react.bridge.WritableMap
+import com.facebook.react.bridge.WritableNativeMap
+import com.reactnativestripesdk.createError
+import com.google.android.gms.tasks.Task
+
+object TapAndPayProxy {
```
charliecruzan-stripe, line 12: Since google's tap and pay library is distributed as an `.aar`, we cannot use `compileOnly` 😢  so this object uses a a decent amount of reflection to get the job done

ios/pushprovisioning/AddToWalletButtonView.swift
```diff
@@ -0,0 +1,148 @@
...
+                    error as NSError?
+                ) as? [AnyHashable : Any]
+            )
+        } else {
+            onCompleteAction!([
+                "error": NSNull(),
+            ] as [AnyHashable : Any])
+        }
+        let vc = findViewControllerPresenter(from: UIApplication.shared.delegate?.window??.rootViewController ?? UIViewController())
+        vc.dismiss(animated: true, completion: nil)
```
ramont-stripe, line 129: Dismissing `controller` can be safer here.

```suggestion
        controller.dismiss(animated: true, completion: nil)
```

ramont-stripe (commented):
iOS changes lgtm!

android/src/main/java/com/reactnativestripesdk/pushprovisioning/AddToWalletButtonView.kt
```diff
@@ -0,0 +1,152 @@
...
+
+
+class AddToWalletButtonView(private val context: ThemedReactContext, private val requestManager: RequestManager) : AppCompatImageView(context) {
+  private var cardDescription: String? = null
+  private var cardLastFour: String? = null
+  private var ephemeralKey: String? = null
+  private var sourceMap: ReadableMap? = null
+  private var token: ReadableMap? = null
+
+  private var mEventDispatcher: EventDispatcher? = context.getNativeModule(UIManagerModule::class.java)?.eventDispatcher
```
jameswoo-stripe, line 31: nit: remove m

android/src/main/java/com/reactnativestripesdk/pushprovisioning/AddToWalletButtonView.kt
```diff
@@ -0,0 +1,152 @@
...
+  private var loadedSource: GlideUrl? = null
+  private var heightOverride: Int = 0
+  private var widthOverride: Int = 0
+
+  override fun performClick(): Boolean {
```
jameswoo-stripe, line 36: While testing performClick without push provisioning dependency, I see a few unhandled promise rejections. Should these be handled?

```
(node:88963) UnhandledPromiseRejectionWarning: Error: Passing apiVersion in a separate options hash is required to create an ephemeral key. See https://stripe.com/docs/api/versioning?lang=node
    at Object.validator (/Users/jameswoo/stripe/stripe-react-native/example/node_modules/stripe/lib/resources/EphemeralKeys.js:16:15)
    at getRequestOpts (/Users/jameswoo/stripe/stripe-react-native/example/node_modules/stripe/lib/makeRequest.js:59:10)
    at /Users/jameswoo/stripe/stripe-react-native/example/node_modules/stripe/lib/makeRequest.js:83:14
    at new Promise (<anonymous>)
    at makeRequest (/Users/jameswoo/stripe/stripe-react-native/example/node_modules/stripe/lib/makeRequest.js:80:10)
    at Constructor.create (/Users/jameswoo/stripe/stripe-react-native/example/node_modules/stripe/lib/StripeMethod.js:38:7)
    at _callee7$ (/Users/jameswoo/stripe/stripe-react-native/example/dist/index.js:1:14451)
    at tryCatch (/Users/jameswoo/stripe/stripe-react-native/example/node_modules/regenerator-runtime/runtime.js:63:40)
    at Generator.invoke [as _invoke] (/Users/jameswoo/stripe/stripe-react-native/example/node_modules/regenerator-runtime/runtime.js:294:22)
    at Generator.next (/Users/jameswoo/stripe/stripe-react-native/example/node_modules/regenerator-runtime/runtime.js:119:21)
(node:88963) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 11)
(node:88963) UnhandledPromiseRejectionWarning: Error: No such issuing card: 'ic_1KnTM2F05jLespP6wNLZQ1mu'
    at Function.generate (/Users/jameswoo/stripe/stripe-react-native/example/node_modules/stripe/lib/Error.js:40:16)
    at res.toJSON.then.StripeAPIError.message (/Users/jameswoo/stripe/stripe-react-native/example/node_modules/stripe/lib/StripeResource.js:220:35)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
(node:88963) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 12)
```

android/src/main/java/com/reactnativestripesdk/pushprovisioning/PushProvisioningProxy.kt
```diff
@@ -0,0 +1,109 @@
...
+      description = cardDescription
+      tokenRequiringTokenization = token
+      createActivityEventListener(context, view)
+      DefaultPushProvisioningProxy().beginPushProvisioning(
+        context.currentActivity!!,
+        description,
+        EphemeralKeyProvider(ephemeralKey)
+      )
+    } catch (e: Exception) {
+      Log.e(TAG, "PushProvisioning dependency not found")
```
jameswoo-stripe, line 48: I wasn't able to see this log in the console when I removed the dependency and pressed add to gpay, am I missing something?

android/src/main/java/com/reactnativestripesdk/pushprovisioning/PushProvisioningProxy.kt
```diff
@@ -0,0 +1,109 @@
...
+          ephemeralKey: String,
+          token: ReadableMap?
+  ) {
+    try {
+      Class.forName("com.stripe.android.pushProvisioning.PushProvisioningActivityStarter")
+      description = cardDescription
+      tokenRequiringTokenization = token
+      createActivityEventListener(context, view)
+      DefaultPushProvisioningProxy().beginPushProvisioning(
+        context.currentActivity!!,
```
jameswoo-stripe, line 43: is this safe?

android/src/main/java/com/reactnativestripesdk/pushprovisioning/AddToWalletButtonView.kt
```diff
@@ -0,0 +1,152 @@
...
+          override fun onLoadFailed(e: GlideException?, model: Any?, target: Target<Drawable>?, isFirstResource: Boolean): Boolean {
+            dispatchEvent(
+              createError("Failed", "Failed to load the source from $sourceToLoad")
+            )
+            return true
+          }
+          override fun onResourceReady(resource: Drawable?, model: Any?, target: Target<Drawable>?, dataSource: DataSource?, isFirstResource: Boolean): Boolean {
+            setImageDrawable(
+              RippleDrawable(
+                ColorStateList.valueOf(Color.parseColor("#e0e0e0")),
```
jameswoo-stripe, line 92: Is this color Google pay brand guideline compliant?

android/src/main/java/com/reactnativestripesdk/pushprovisioning/PushProvisioningProxy.kt
```diff
@@ -0,0 +1,109 @@
...
+    TapAndPayProxy.invoke(activity, cardLastFour, promise)
+  }
+
+  private fun createActivityEventListener(context: ReactApplicationContext, view: AddToWalletButtonView) {
+    val listener = object : BaseActivityEventListener() {
+      override fun onActivityResult(activity: Activity, requestCode: Int, resultCode: Int, data: Intent?) {
+        super.onActivityResult(activity, requestCode, resultCode, data)
+        if (requestCode == TapAndPayProxy.REQUEST_CODE_TOKENIZE) {
+          view.dispatchEvent(
+            if (resultCode == RESULT_OK) null else mapError("Failed", "Failed to resolve yellow path.", null, null, null, null)
```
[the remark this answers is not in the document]
charliecruzan-stripe: This is google's terminology for that particular state- https://developers.google.com/pay/issuers/apis/push-provisioning/android/wallet-operations#resolving_yellow_path. Maybe better to change it to something like "Failed to verify identity" since this path is only encountered when the token state is `TOKEN_STATE_NEEDS_IDENTITY_VERIFICATION`

android/src/main/java/com/reactnativestripesdk/pushprovisioning/AddToWalletButtonView.kt
charliecruzan-stripe: This is a rejection coming from the client app code, not our SDK code. Your getting an empty string value for [this constants value](https://github.com/stripe/stripe-react-native/blob/charliecruzan-push-provisioning/example/src/screens/GooglePayScreen.tsx#L157) so the server is complaining when you try to fetch an ephemeral key.

charliecruzan-stripe: so basically- it's expected

android/src/main/java/com/reactnativestripesdk/pushprovisioning/PushProvisioningProxy.kt
charliecruzan-stripe: since [you're not passing an `ephemeralKey`](https://github.com/stripe/stripe-react-native/pull/889#discussion_r856429833), I bet you're actually receiving [this error via the `onComplete` hook?](https://github.com/stripe/stripe-react-native/blob/charliecruzan-push-provisioning/android/src/main/java/com/reactnativestripesdk/pushprovisioning/AddToWalletButtonView.kt#L48)

charliecruzan-stripe: so the `.invoke` function isn't actually called.

android/src/main/java/com/reactnativestripesdk/pushprovisioning/AddToWalletButtonView.kt
charliecruzan-stripe: I used the same color code that is used by default in the google pay buttons, but couldn't find any branding guidelines about ripple color [here](https://developers.google.com/pay/issuers/apis/push-provisioning/android/branding-guidelines#flutter)

android/src/main/java/com/reactnativestripesdk/pushprovisioning/PushProvisioningProxy.kt
charliecruzan-stripe: Since `currentActivity` is pretty much only ever `null` in React Native modules super early on during app startup, and this requires a touch event to trigger, I believe this is pretty safe. can add a null check though if you think it's necessary?

jameswoo-stripe: I think it would be better to have a null check so we don't have unexpected crashes.

charliecruzan-stripe: agreed 👍  added here- https://github.com/stripe/stripe-react-native/pull/889/commits/187757831fd759dc13c09cf5327c86f67a4b76c4

android/src/main/java/com/reactnativestripesdk/pushprovisioning/TapAndPayProxy.kt
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  private fun mapFromTokenState(status: Int): String {
+    return when (status) {
+      getTokenState("TOKEN_STATE_NEEDS_IDENTITY_VERIFICATION") -> "TOKEN_STATE_NEEDS_IDENTITY_VERIFICATION"
+      getTokenState("TOKEN_STATE_PENDING") -> "TOKEN_STATE_PENDING"
+      getTokenState("TOKEN_STATE_SUSPENDED") -> "TOKEN_STATE_SUSPENDED"
+      getTokenState("TOKEN_STATE_ACTIVE") -> "TOKEN_STATE_ACTIVE"
+      getTokenState("TOKEN_STATE_FELICA_PENDING_PROVISIONING") -> "TOKEN_STATE_FELICA_PENDING_PROVISIONING"
+      getTokenState("TOKEN_STATE_UNTOKENIZED") -> "TOKEN_STATE_UNTOKENIZED"
```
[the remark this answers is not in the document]
charliecruzan-stripe: Great catch, since these are ints we actually need to use reflection again 🪞 

fixed here-
https://github.com/stripe/stripe-react-native/pull/889/commits/3259f1d15a2a2c1594951405d0caab8c8c1e2c59

charliecruzan-stripe:
@jameswoo-stripe lmk if recent changes are good and android has the ✅

jameswoo-stripe (approved):
Android lgtm

PR #890, opened 2022-04-15
BREAKING: Change parameters for `confirmPayment`, `confirmSetupIntent`, and `createPaymentMethod`

charliecruzan-stripe:
## Summary

Please see the [migration guide]() added in this PR for a full description of changes

## Motivation

This more closely matches types in the stripe-js library which will streamline the experience, and it's more clear what the parameters are. Plus, this will make adding future payment methods that have the option to attach outside of the confirm step (like `USBankAccount`) easier.

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 


android/src/main/java/com/reactnativestripesdk/PaymentMethodCreateParamsFactory.kt
```diff
@@ -5,11 +5,12 @@ import com.stripe.android.model.*
 
 class PaymentMethodCreateParamsFactory(
   private val clientSecret: String,
-  private val params: ReadableMap,
+  private val paymentMethodData: ReadableMap?,
+  private val options: ReadableMap,
```
acomley-stripe, line 9: Just a quick clarification to make sure I understand the change. Previously params could contain the paymentMethodData or options. Now we are being explicit to include paymentMethodData and what was params now becomes options?

acomley-stripe, line 9: Just a quick clarification to make sure I understand the change. Previously params could contain the paymentMethodData or options. Now we are being explicit to include paymentMethodData and what was params now becomes options?

charliecruzan-stripe: previously we accepted `params` and `options`. `options` was empty, we expected nothing from there, and params contained:
1. the payment method type
2. the associated data with that payment method (id, iban number, billing info, etc.)
3. `setupFutureUsage` 

we have changed the name of 1 from `type` to `paymentMethodType`, but the value type is still just a `String`. we've moved all of 2 to go inside of a sub-object (inside of `params`) named `paymentMethodData`. And we moved 3 out of `params` and into `options`, where i believe it makes more sense to have it.

acomley-stripe (approved):

Issue #841, opened 2022-03-11
error: Error: Unable to resolve module ../../package.json


marlon07021:
**Describe the bug**
Getting this error after installing the module. Already tried everything from #770  but no success.

```
error: Error: Unable to resolve module ../../package.json from /Users/user/Turix/turix-rn-mobile/node_modules/@stripe/stripe-react-native/lib/commonjs/components/StripeProvider.js: 

None of these files exist:
  * node_modules/@stripe/stripe-react-native/lib/package.json(.native|.ios.js|.native.js|.js|.ios.json|.native.json|.json|.ios.ts|.native.ts|.ts|.ios.tsx|.native.tsx|.tsx|.ios.cjs|.native.cjs|.cjs)
  * node_modules/@stripe/stripe-react-native/lib/package.json/index(.native|.ios.js|.native.js|.js|.ios.json|.native.json|.json|.ios.ts|.native.ts|.ts|.ios.tsx|.native.tsx|.tsx|.ios.cjs|.native.cjs|.cjs)
```

**To Reproduce**
1 - create a new typescript react native project.
2 - install stripe module for react native

**Expected behavior**
Normally running

**Dependencies:**
 ```
 "dependencies": {
    "@apollo/client": "^3.5.5",
    "@react-native-async-storage/async-storage": "^1.13.4",
    "@react-native-community/masked-view": "^0.1.10",
    "@react-native-picker/picker": "^2.3.1",
    "@react-navigation/bottom-tabs": "^5.11.7",
    "@react-navigation/native": "^5.9.2",
    "@react-navigation/stack": "^5.14.2",
    "@reduxjs/toolkit": "^1.5.0",
    "@stripe/stripe-react-native": "^0.4.0",
    "@titelmedia/react-apollo-multiple-clients": "^4.0.5",
    "graphql": "^16.0.1",
    "i18n-js": "^3.8.0",
    "intl": "^1.2.5",
    "lodash.memoize": "^4.1.2",
    "moment": "^2.29.1",
    "react": "17.0.2",
    "react-apollo": "^3.1.5",
    "react-native": "0.67.3",
    "react-native-animatable": "^1.3.3",
    "react-native-appearance": "^0.3.4",
    "react-native-calendar-picker": "^7.1.2",
    "react-native-device-info": "^8.0.1",
    "react-native-dotenv": "^3.3.0",
    "react-native-elements": "^3.2.0",
    "react-native-fbsdk": "^3.0.0",
[...]

PR #902, opened 2022-04-22
fix: create custom babel plugin for package.json imports in `src/`

charliecruzan-stripe:
## Summary

babel plugin identifies import declarations that need to be modified for transpilation. right now, that's only the `../../package.json` import

## Motivation
fixes https://github.com/stripe/stripe-react-native/issues/841

## Testing

run `yarn prepare` with these changes, and `lib/commonjs/components/StripeProvider.js` and `lib/module/components/StripeProvider.js` have the correct file path

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -4,10 +4,10 @@ module.exports = {
     function ModifyPackageJsonImportForTranspilation() {
       return {
         visitor: {
-          ImportDeclaration(path, _state) {
-            const { node } = path;
-            if (node.source.value === '../../package.json') {
-              node.source.value = '../../../package.json';
+          ImportDeclaration(path) {
+            let importResource = path.node.source.value ?? '';
+            if (importResource.includes('../package.json')) {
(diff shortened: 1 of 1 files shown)


byCedric:
Thx bae ❤️

peterpme:
Thx baes ❤️

babel.config.js
```diff
@@ -1,3 +1,17 @@
 module.exports = {
   presets: ['module:metro-react-native-babel-preset'],
+  plugins: [
+    function ModifyPackageJsonImportForTranspilation() {
+      return {
+        visitor: {
+          ImportDeclaration(path) {
+            let importResource = path.node.source.value ?? '';
+            if (importResource.includes('../package.json')) {
+              importResource = '../' + importResource;
```
charliecruzan-stripe, line 10: hm actually i think this isn't assign the new path to the actual `path.node.source.value`, will commit another fix

PR #912, opened 2022-04-28
fix: Providing zip code straight from CardField component on Android

charliecruzan-stripe:
## Summary

Previously, you had to pass billingDetails to capture zipcode from CardFieldView (which doesn't make much sense)

## Motivation

It's a bug, people should be able to rely on CardField for zip code and pass null for `billingDetails`

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -292,7 +292,7 @@ PODS:
     - StripeApplePay (= 22.0.0)
     - StripeCore (= 22.0.0)
     - StripeUICore (= 22.0.0)
-  - stripe-react-native (0.7.0):
+  - stripe-react-native (0.8.0):
     - React-Core
     - Stripe (~> 22.0.0)
     - StripeConnections (~> 22.0.0)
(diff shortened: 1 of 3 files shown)


example/ios/Podfile.lock
```diff
@@ -292,7 +292,7 @@ PODS:
     - StripeApplePay (= 22.0.0)
     - StripeCore (= 22.0.0)
     - StripeUICore (= 22.0.0)
-  - stripe-react-native (0.7.0):
+  - stripe-react-native (0.8.0):
```
acomley-stripe, line 295: Should this be added as a release step in the future?

acomley-stripe (approved):

example/ios/Podfile.lock
charliecruzan-stripe: eh, kinda doesn't matter, this is just the example app 🤷 I think to include this all we need is to `yarn bootstrap` again after `yarn -s --ignore-scripts publish --non-interactive --access=public`. So it'll add time but probably make things cleaner

PR #913, opened 2022-04-28
breaking: use a cardDetails prop for AddToWalletButton component

charliecruzan-stripe:
## Summary

Modified the props for the `AddToWalletButton` component. Instead of passing `cardHolderName`, `cardLastFour`, `cardDescription`, and `cardBrand` directly as props, you will instead pass a `cardDetails` prop, which is an object containing the following fields:
  - `primaryAccountIdentifier`: The `wallet.primary_account_identifier` value from the issued card.
  - `name`: The card holder name (previously `cardHolderName`).
  - `description`: A user-facing description of the card (previously `cardDescription`).
  - `lastFour`: Last 4 digits of the card, optional (previously `cardLastFour`).
  - `brand`: The card brand, optional (previously `cardBrand`).

## Motivation

Stripe iOS [recently updated to expose a method that sets the `primaryAccountIdentifier`](https://stripe.com/docs/issuing/cards/digital-wallets?platform=iOS), and without this push provisioning can fail on iOS. This provided a good opportunity to instead pass a `cardDetails` rather than individual props for each field. 

## Testing

- [x] I tested this manually
- [ ] I added automated tests

Note: This was tested via the `STPFakeAddPaymentPassViewController`

## Documentation

I will update https://stripe.com/docs/issuing/cards/digital-wallets?platform=react-native accordingly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,13 @@
 
 ## Unreleased
 
-- [#913](https://github.com/stripe/stripe-react-native/pull/913) chore: Updated `stripe-ios` from 22.0.0 to 22.2.0.
-- [#913](https://github.com/stripe/stripe-react-native/pull/913) BREAKING CHANGE: Changed props for the `AddToWalletButton` component. Instead of passing `cardHolderName`, `cardLastFour`, `cardDescription`, and `cardBrand` directly as props, you will instead pass a `cardDetails` prop, which is an object containing the following fields:
+- [#913](https://github.com/stripe/stripe-react-native/pull/913) BREAKING CHANGE: Changed props for the `<AddToWalletButton />` component. Instead of passing `cardHolderName`, `cardLastFour`, `cardDescription`, and `cardBrand` directly as props, you will instead pass a `cardDetails` prop, which is an object containing the following fields:
   - `primaryAccountIdentifier`: The `wallet.primary_account_identifier` value from the issued card.
   - `name`: The card holder name (previously `cardHolderName`).
   - `description`: A user-facing description of the card (previously `cardDescription`).
   - `lastFour`: Last 4 digits of the card, optional (previously `cardLastFour`).
   - `brand`: The card brand, optional (previously `cardBrand`).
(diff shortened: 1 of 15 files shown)


davidme-stripe (approved):

davidme-stripe (changes requested):

davidme-stripe (approved):

src/components/AddToWalletButton.tsx
```diff
@@ -29,14 +29,19 @@ export interface Props extends AccessibilityProps {
...
-  /** Last 4 digits of the card. Required on Android. */
-  cardLastFour: string;
-  /** Sets the card holder name (used only on iOS) */
-  cardDescription?: string;
-  /** Optional, only used on iOS */
-  cardBrand?: Token.CardBrand;
+  /** Details of the Issued Card you'd like added to the device's wallet */
+  cardDetails: {
+    /** The `primary_account_identifier` value from the issued card. */
+    primaryAccountIdentifier: string;
```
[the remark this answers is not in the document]
davidme-stripe: Should it be nullable here, too? Not sure how React works.

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # CHANGELOG
 
+## Unreleased
+
+- [#913](https://github.com/stripe/stripe-react-native/pull/913) BREAKING CHANGE: Changed props for the `<AddToWalletButton />` component. Instead of passing `cardHolderName`, `cardLastFour`, `cardDescription`, and `cardBrand` directly as props, you will instead pass a `cardDetails` prop, which is an object containing the following fields:
+  - `primaryAccountIdentifier`: The `wallet.primary_account_identifier` value from the issued card.
```
lukas-stripe, line 6: Should we mention that it's optional only in the sense that it's nilable, but users should aways pass it through as-is (i.e. nil or non-nil)?

lukas-stripe (changes requested):
One suggestion, otherwise looks good (as far as I can follow React Native 😅)

CHANGELOG.md
charliecruzan-stripe: Sure, that makes sense and I can add that to the docs :)

PR #935, opened 2022-05-10
Fix Expo Config Plugin Deps for SDK45

ScreamZ:
## Summary

After SDK 45 Release, we need to update expo-config plugin library. Otherwise it won't build.

## Testing

- [ ] I tested this manually
- [ ] I added automated tests

## Documentation

This is a symptom caused by having a dependency that uses an older version of @expo/config-plugins. If you run yarn why @expo/config-plugins or npm why @expo/config-plugins you can see which package uses an older version. Make sure that you update that package to the latest version, as older versions of @expo/config-plugins are not compatible with SDK 45.

Select one: 


charliecruzan-stripe:
thanks @ScreamZ ! is there a changelog for what breaking changes were made from v3->v4?

ScreamZ:
@charliecruzan-stripe 
Guess related to this commit : https://github.com/expo/expo-cli/commit/d614ee9c98268686dbe5eeea5bfa512b89325d3f

https://github.com/expo/expo-cli/blob/main/CHANGELOG.md#tue-21-sep-2021-143900--0700

charliecruzan-stripe:
So many changes to the `yarn.lock` seems odd to me 🤔 going to merge https://github.com/stripe/stripe-react-native/pull/936 instead, is that alright @ScreamZ ?

ScreamZ:
> So many changes to the `yarn.lock` seems odd to me 🤔 going to merge #936 instead, is that alright @ScreamZ ?

Of course, I've no idea what happened with my lock. Thanks for this quick fix 🙏🏻 I'm gonna check the patch tomorrow :-) so many issues with various lib on SDK45 (just released)

PR #936, opened 2022-05-10
fix: upgrade expo config plugins

charliecruzan-stripe:
replaces https://github.com/stripe/stripe-react-native/pull/935

### DIFF
--- a/package.json
+++ b/package.json
@@ -154,6 +154,6 @@
     ]
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


brentvatne:
hey @charliecruzan-stripe! any chance you can back port this to 0.6.x? we included that version in the sdk 45 release and it looks like there are some breaking changes between 0.6.0 and 0.9.0 that would impact expo go

PR #942, opened 2022-05-16
Update iOS SDK to 22.3.1

davidme-stripe:
## Summary
To pull in the fix for https://github.com/stripe/stripe-ios/issues/1960, we should update to the latest iOS SDK.

(Feel free to correct this PR if needed — I'm not sure if we need to update this value elsewhere in the RN SDK!)

## Testing
Existing CI tests

## Documentation

Select one: 

### DIFF
--- a/stripe-react-native.podspec
+++ b/stripe-react-native.podspec
@@ -1,7 +1,7 @@
 require 'json'
 
 package = JSON.parse(File.read(File.join(__dir__, 'package.json')))
-stripe_version = '~> 22.3.0'
+stripe_version = '~> 22.3.1'
 
 Pod::Spec.new do |s|
   s.name         = 'stripe-react-native'


stripe-react-native.podspec
```diff
@@ -1,7 +1,7 @@
 require 'json'
 
 package = JSON.parse(File.read(File.join(__dir__, 'package.json')))
-stripe_version = '~> 22.3.0'
+stripe_version = '~> 22.3.1'
```
charliecruzan-stripe, line 4: `~> 22.3.0` should match up to but not including `22.4.0`, so I don't think we need to deploy this change for users to get the most recent `stripe-ios` version 👍

charliecruzan-stripe (approved):

charliecruzan-stripe:
@davidme-stripe could you re-run `yarn bootstrap` from the root of the repo? You may need to clean up pods in `example/ios` before doing so, otherwise it'll fail to `pod install` Stripe and StripeConnections

charliecruzan-stripe (changes requested):
@davidme-stripe could you re-run yarn bootstrap from the root of the repo? You may need to clean up pods in example/ios before doing so, otherwise it'll fail to pod install Stripe and StripeConnections

stripe-react-native.podspec
davidme-stripe: Okay, sounds good! I'll just close this then.

PR #946, opened 2022-05-19
chore: Remove usage of LocalBroadCastManager

charliecruzan-stripe:
## Summary

We used to rely on LocalBroadcastManager to send/receive intents in order to resolve promises based on behavior in fragments.

## Motivation

We don't need to rely on intents for this: it's confusing, more bug prone, and although it's safe to use LocalBroadcastManager, some folks assume we're using BroadcastManager which is **not** secure (to be clear, we are NOT using BroadcastManager).

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 


jameswoo-stripe (approved):
lgtm 🎉

Issue #930, opened 2022-05-09
Could not find constructor when recreating Activity


RafaRuiz:
**Describe the bug**
We are having some crash reports of some users:
```
    java.lang.RuntimeException: Unable to start activity ComponentInfo{com.example.flavouruat/com.example.MainActivity}: androidx.fragment.app.Fragment$InstantiationException: Unable to instantiate fragment com.reactnativestripesdk.PaymentLauncherFragment: could not find Fragment constructor
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3774)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3936)
        at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:103)
        at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
        at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2341)
        at android.os.Handler.dispatchMessage(Handler.java:106)
        at android.os.Looper.loopOnce(Looper.java:233)
        at android.os.Looper.loop(Looper.java:344)
        at android.app.ActivityThread.main(ActivityThread.java:8184)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:584)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1034)
     Caused by: androidx.fragment.app.Fragment$InstantiationException: Unable to instantiate fragment com.reactnativestripesdk.PaymentLauncherFragment: could not find Fragment constructor
        at androidx.fragment.app.Fragment.instantiate(Fragment.java:630)
        at androidx.fragment.app.FragmentContainer.instantiate(FragmentContainer.java:57)
        at androidx.fragment.app.FragmentManager$2.instantiate(FragmentManager.java:448)
        at androidx.fragment.app.FragmentState.instantiate(FragmentState.java:81)
        at androidx.fragment.app.FragmentStateManager.<init>(FragmentStateManager.java:85)
        at androidx.fragment.app.FragmentManager.restoreSaveStateInternal(FragmentManager.java:2410)
        at androidx.fragment.app.FragmentManager.attachController(FragmentManager.java:2584)
[...]

PR #965, opened 2022-05-25
fix: instantiate PaymentLauncher as needed instead of on init

charliecruzan-stripe:
## Summary

Previously we were instantiating payment launcher on SDK initialization. Now, we init whenever we need, and then tear down the Fragment afterwards

## Motivation

closes https://github.com/stripe/stripe-react-native/issues/930
closes https://github.com/stripe/stripe-react-native/issues/968

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/PaymentLauncherFragment.kt
+++ b/android/src/main/java/com/reactnativestripesdk/PaymentLauncherFragment.kt
@@ -5,8 +5,10 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.widget.FrameLayout
+import androidx.appcompat.app.AppCompatActivity
 import androidx.fragment.app.Fragment
 import com.facebook.react.bridge.Promise
+import com.facebook.react.bridge.ReactApplicationContext
 import com.stripe.android.ApiResultCallback
 import com.stripe.android.Stripe
 import com.stripe.android.model.*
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/reactnativestripesdk/PaymentLauncherFragment.kt
```diff
@@ -5,83 +5,171 @@ import android.view.LayoutInflater
...
+      } ?: run {
+        promise.resolve(createMissingActivityError())
+      }
+    }
+  }
 
   override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?,
                             savedInstanceState: Bundle?): View {
     paymentLauncher = createPaymentLauncher()
+    if (paymentIntentClientSecret != null && confirmPaymentParams != null) {
```
skyler-stripe, line 124: Is it possible to test this? 

I was thinking you could just verify the correct methods are being called with given params. If that's a lot of work feel free to just merge.

skyler-stripe (approved):

android/src/main/java/com/reactnativestripesdk/PaymentLauncherFragment.kt
charliecruzan-stripe: This is technically covered by tests, just in E2E tests, not unit tests.

android/src/main/java/com/reactnativestripesdk/PaymentLauncherFragment.kt
```diff
@@ -5,83 +5,171 @@ import android.view.LayoutInflater
...
+        stripeAccountId,
+        promise,
+        handleNextActionClientSecret = handleNextActionClientSecret,
+      )
+      addFragment(paymentLauncherFragment, context, promise)
+      return paymentLauncherFragment
+    }
+
+    private fun addFragment(fragment: PaymentLauncherFragment, context: ReactApplicationContext, promise: Promise) {
+      (context.currentActivity as? AppCompatActivity)?.let {
```
jakub-cosmose, line 107: The cast should be to `FragmentActivity` not `AppCompatActivity`. The current solution breaks the Flutter integration. See: https://github.com/flutter-stripe/flutter_stripe/issues/771

PR #986, opened 2022-06-09
feat: add `canAddCardToWallet` method for Push Provisioning

charliecruzan-stripe:
## Summary

Adds a method to check if a particular card can be provisioned with the current app on this particular device.

## Motivation

It's useful to ensure that a card can be added _before_ showing the AddToWallet component. A card can be added if:
- the device supports paying with native wallet cards in the first place
- the app can provision cards
- the card in question hasn't already been added

## Testing

- [x] I tested this manually
- [x] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
+++ b/android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
@@ -7,15 +7,17 @@ import android.util.Log
 import androidx.appcompat.app.AppCompatActivity
 import com.facebook.react.bridge.*
 import com.facebook.react.module.annotations.ReactModule
+import com.reactnativestripesdk.pushprovisioning.PushProvisioningProxy
 import com.stripe.android.*
-import com.stripe.android.core.AppInfo
 import com.stripe.android.core.ApiVersion
+import com.stripe.android.core.AppInfo
 import com.stripe.android.model.*
 import com.stripe.android.payments.bankaccount.CollectBankAccountConfiguration
 import com.stripe.android.view.AddPaymentMethodActivityStarter
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
+
 
 @ReactModule(name = StripeSdkModule.NAME)
 class StripeSdkModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaModule(reactContext) {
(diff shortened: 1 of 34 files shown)


ios/pushprovisioning/PushProvisioningUtils.swift
```diff
@@ -0,0 +1,58 @@
...
+
+import Foundation
+import Stripe
+
+class PushProvisioningUtils {
+    class func canAddCardToWallet(
+        last4: String,
+        primaryAccountIdentifier: String,
+        testEnv: Bool
+    ) -> (canAddCard: Bool, details: NSDictionary) {
```
yuki-stripe, line 16: Seems like we could just pass a status enum back instead of a details dictionary, wdyt?  

I'm not sure but it looks like the caller of this method is redundantly putting the details returned here into another dictionary too.

yuki-stripe (commented):
Why no unit tests for this method, if only for the test env scenarios?

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
```diff
@@ -525,14 +527,47 @@ class StripeSdkModule(reactContext: ReactApplicationContext) : ReactContextBaseJ
...
+  fun canAddCardToWallet(params: ReadableMap, promise: Promise) {
+    val last4 = getValOr(params, "cardLastFour", null) ?: run {
+      promise.resolve(createError("Failed", "You must provide cardLastFour"))
+      return
+    }
+
+    if (!PushProvisioningProxy.isNFCEnabled(reactApplicationContext)) {
+      promise.resolve(createCanAddCardResult(false, "UNSUPPORTED_DEVICE"))
+      return
+    }
```
brnunes-stripe, line 540: Is it a valid use case to tokenize a card into a device that doesn't have NFC, so that it can be used to pay online?

android/src/main/java/com/reactnativestripesdk/pushprovisioning/TapAndPayProxy.kt
```diff
@@ -2,54 +2,68 @@ package com.reactnativestripesdk.pushprovisioning
...
-      val getClientMethod = tapAndPayClass.getMethod("getClient", Activity::class.java)
+      val getClientMethod = tapAndPayClass.getMethod(
+        "getClient",
+        Activity::class.java)
+      val client = getClientMethod.invoke(null, activity)
 
-      tapAndPayClient = getClientMethod.invoke(null, activity).also {
-        val tapAndPayClientClass = Class.forName("com.google.android.gms.tapandpay.TapAndPayClient")
-        val listTokensMethod = tapAndPayClientClass.getMethod("listTokens")
+      val tapAndPayClientClass = Class.forName("com.google.android.gms.tapandpay.TapAndPayClient")
```
brnunes-stripe, line 26: Can you depend on the library as `compileOnly` so that you don't need to use reflection?

charliecruzan-stripe: No, unfortunately since google's library is distributed as an .aar, we cannot use compileOnly 😞

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
charliecruzan-stripe: Hm, I hadn't considered that use case. So you're saying to still return the `UNSUPPORTED_DEVICE` result, but attempt to get the google pay card token anyways?

charliecruzan-stripe:
@yuki-stripe added iOS unit tests. Adding android ones next, but would like to get this all-else approved besides the Android unit tests if possible :)

charliecruzan-stripe:
Yeah I understand the confusion- was hesitant to have a `failureStatus` or `errorCode` field _outside_ of the `error` response, though, since that seemed odd to me. 

> I'd expect something called AddCardToWalletStatus to include the case where the card was successfully added to the wallet.

This method is just checking for the device's & app's _ability_ to add a card to the wallet, not actually adding it.

yuki-stripe:
> Yeah I understand the confusion- was hesitant to have a `failureStatus` or `errorCode` field _outside_ of the `error` response, though, since that seemed odd to me.
> 
> > I'd expect something called AddCardToWalletStatus to include the case where the card was successfully added to the wallet.
> 
> This method is just checking for the device's & app's _ability_ to add a card to the wallet, not actually adding it.

It looks like this is the same pattern as the CanAddCardToWalletResult, so maybe I really have a problem with how that's structured - but since we've already gaveled the API I'm okay to drop.

yuki-stripe (approved):

charliecruzan-stripe:
> It looks like this is the same pattern as the CanAddCardToWalletResult

@yuki-stripe what if we added another status that corresponds to `canAddCard == true`? So the diff would be something like:


```diff
        status?:
+         | 'CAN_ADD_CARD',
          | 'MISSING_CONFIGURATION'
          | 'UNSUPPORTED_DEVICE'
          | 'CARD_ALREADY_EXISTS';
```

PR #989, opened 2022-06-13
fix: Make CardField postal code behavior consistent on iOS and Android

charliecruzan-stripe:
## Summary & Motivation

previously, the `complete` field in the `onCardChange` callback would incorrectly be set to `true` if the postal code wasn't filled out on Android (not ios). This fixes that, and also implements the same behavior as ios- if the user's current device locale is a country/region without postal codes, the postal code entry isn't shown or required. 

Users can also override the country code by passing the `countryCode` prop

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/CardFieldView.kt
+++ b/android/src/main/java/com/reactnativestripesdk/CardFieldView.kt
@@ -33,6 +33,7 @@ class CardFieldView(context: ThemedReactContext) : FrameLayout(context) {
   private var mEventDispatcher: EventDispatcher? = context.getNativeModule(UIManagerModule::class.java)?.eventDispatcher
   private var dangerouslyGetFullCardDetails: Boolean = false
   private var currentFocusedField: String? = null
+  private var isCardValid = false
 
   init {
     cardInputWidgetBinding.container.isFocusable = true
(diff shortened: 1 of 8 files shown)


android/src/main/java/com/reactnativestripesdk/CardFieldView.kt
```diff
@@ -197,6 +200,19 @@ class CardFieldView(context: ThemedReactContext) : FrameLayout(context) {
     mCardWidget.postalCodeEnabled = isEnabled
   }
 
+  fun setCountryCode(countryCode: String?) {
+    val defaultLocale = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+      if (context.resources.configuration.locales.isEmpty) "US"
+      else context.resources.configuration.locales[0].country
+    } else {
+      context.resources.configuration.locale.country
+    }
```
brnunes-stripe, line 209: [`LocaleListCompat#getAdjustedDefault`](https://developer.android.com/reference/kotlin/androidx/core/os/LocaleListCompat#getAdjustedDefault()) gets you the default locale for any version.

brnunes-stripe (approved):

Issue #999, opened 2022-06-17
[0.12.0] Opening > Closing > Re-opening the payment sheet makes the app crash on android


xseignard:
**Describe the bug**
Opening > Closing > Re-opening the payment sheet makes the app crash on android only (happening in debug and prod builds)

**To Reproduce**
1. Open a payment sheet
2. Dismiss it with the close cross icon
3. Try to re-open it
4. The app crashes

**Expected behavior**
The payment sheet should re-open

**Screenshots**
N/A

**Smartphone (please complete the following information):**
 - Device: Pixel 5
 - OS: Android 12
 - Browser N/A (react-native)
 - Version `"@stripe/stripe-react-native": "0.12.0"`

**Additional context**
Please find the stack trace below:
```java
[...]

PR #1014, opened 2022-06-23
fix: don't aggressively remove payment sheet fragment

charliecruzan-stripe:
## Summary

Previously we would remove the payment sheet fragment after the result callback completed. But if customers cancel and then launch the same payment sheet again, this would cause an issue. Now, if the payment succeeds, we know we can remove the fragment. Otherwise, we don't remove it until a new payment sheet is being initialized (there are never two payment sheets initialized at once)

## Motivation
fixes https://github.com/stripe/stripe-react-native/issues/999

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 


charliecruzan-stripe:
ios test failure looks due to flakiness (this PR makes no changes whatsoever to the iOS codebase)

PR #1051, opened 2022-07-20
feat: add dark google pay button option

charliecruzan-stripe:
## Summary

Added support for `pay_dark` and `standard_dark` to the `<GooglePayButton />` component's `type` prop. This allows you to display the [dark Google Pay button](https://developers.google.com/pay/api/android/guides/brand-guidelines). 

**This includes a breaking change.**
This changes the `<GooglePayButton />` component behavior in that it no longer automatically uses the dark version when the device is in Dark Mode. Instead, merchants should detect Dark Mode in their app (part of [react native core](https://reactnative.dev/docs/appearance)) and use the `<GooglePayButton />` `type` prop appropriately (according to [Google's guidelines](https://developers.google.com/pay/api/android/guides/brand-guidelines)).

This PR doesn't add any additional assets, it just moves assets from the `*-night` directories into the main `drawable` directories. Previously, the dark google pay buttons were only shown if the device was in dark mode. (see breaking change)
Actually- [this PR _removes_ some unused assets](https://github.com/stripe/stripe-react-native/pull/1051/commits/b98ae935fb28910757b16c79bffb92b63e0a8bc1) 😄 

## Motivation

requested by users, Google will sometimes reject an app if the app isn't using the dark button when on a light background

## Testing

- [x] I tested this manually
- [ ] I added automated tests

https://user-images.githubusercontent.com/97612659/180094747-57cbe11c-a3a9-4f57-a51f-c27d4e436315.mov

## Documentation

Select one: 
- [x] I have added relevant documentation for my changes. (API Ref)


CHANGELOG.md
```diff
@@ -4,8 +4,11 @@
 
 ### Breaking changes
 
+- The `<GooglePayButton />` component no longer automatically uses the dark version when the device is in Dark Mode. Instead, you should detect Dark Mode in your app and use the `<GooglePayButton />` `type` prop appropriately (according to [Google's guidelines](https://developers.google.com/pay/api/android/guides/brand-guidelines)). [#1051](https://github.com/stripe/stripe-react-native/pull/1051)
```
brnunes-stripe, line 7: Could the library still support the native light/dark mode? Something like, you can force it to be light, dark, or not set any value and it will match the current theme?

brnunes-stripe: It looks like that's what it does in `GooglePayButtonView.kt` if you don't set a type, or set it to an invalid value, but this descriptions says otherwise.

src/components/GooglePayButton.tsx
```diff
@@ -11,11 +11,17 @@ import {
...
 
 /**
- *  Apple Pay Button Component Props
+ *  Google Pay Button Component Props
  */
 export interface Props extends AccessibilityProps {
   style?: StyleProp<ViewStyle>;
-  type?: 'pay' | 'pay_shadow' | 'standard' | 'standard_shadow';
+  type?:
+    | 'pay'
```
brnunes-stripe, line 19: I think a `default` here would do what I'm suggesting.

CHANGELOG.md
charliecruzan-stripe: Ah yeah- it _does_ do exactly what you said right now: you can force it to be light, dark, or not set any value and it will match the current theme

The change is that previously, even if you set it to a light button type explicitly, if the device was in dark mode, it would override that and set the dark button. Honestly this isn't much of a breaking change...more of a bug fix 😅  wdyt?

charliecruzan-stripe: I say move this to bug fix and say 

```suggestion
- The `<GooglePayButton />` component no longer overrides the `type` when the device is in Dark Mode and forces a dark button. Instead, you should detect Dark Mode in your app and use the `<GooglePayButton />` `type` prop appropriately (according to [Google's guidelines](https://developers.google.com/pay/api/android/guides/brand-guidelines)). [#1051](https://github.com/stripe/stripe-react-native/pull/1051)
```

brnunes-stripe: I agree with the way it's implemented then, but the comment here is still confusing, especially the last part where you're still telling people to detect Dark Mode themselves. What about this:

- The `<GooglePayButton />` component no longer overrides the `type` to use the dark mode version when the device is in Dark Mode. If you set the `type` value, it will always be respected. If you don't set the `type` value, it will match system's theme. [#1051](https://github.com/stripe/stripe-react-native/pull/1051)

charliecruzan-stripe: im good with that, just a small addition at the end:
```suggestion
- The `<GooglePayButton />` component no longer overrides the `type` to use the dark mode version when the device is in Dark Mode. If you set the `type` value, it will always be respected. If you don't set the `type` value, it will match system's theme (`standard_shadow` when in Light Mode, and `standard_dark` when in Dark Mode). [#1051](https://github.com/stripe/stripe-react-native/pull/1051)
```

Issue #681, opened 2021-10-26
Custom font family


dqii:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Support custom font families (e.g., those imported with [useFonts](https://docs.expo.dev/guides/using-custom-fonts/))

dqii:
Adding context here for anyone else ---
https://github.com/stripe/stripe-react-native/pull/247
It seems like custom fonts are only supported as long as they're system fonts.

edi:
I'm interested in this as well.

binitgurzu:
Still looking for  solutions

charliecruzan-stripe:
Reopening- is there desire for this in the components like CardField and CardForm? Or does the Payment Sheet integration suit everyone's needs here?

dqii:
Would absolutely love for it to be in CardField and CardForm.

charliecruzan-stripe:
@dqii can I ask why Payment Sheet doesn't work for your use case? It would help us in making improvements to Payment Sheet :)

dqii:
I think purely cosmetic reasons. We wanted more flexibility with the design... we wanted payment method to be provided "inline" for example
@charliecruzan-stripe

charliecruzan-stripe:
👍 fair enough, that makes sense

charliecruzan-stripe:
closed in https://github.com/stripe/stripe-react-native/pull/1150

dqii:
@charliecruzan-stripe sorry could this be reopened? I think @perqin's diff only added support for Android

perqin:
It seems that custom font is already supported on iOS.

dqii:
Is it supported? Last I had heard, this was a blocker to supporting custom fonts on iOS: https://github.com/stripe/stripe-ios/issues/1932

dqii:
@charliecruzan-stripe Is it supported? Last I had heard, this was a blocker to supporting custom fonts on iOS: https://github.com/stripe/stripe-ios/issues/1932

charliecruzan-stripe:
If you're looking for custom font control on iOS, I'd recommend either `CardField`, or the Payment Sheet :) I'm not sure about the timeline for adding this to CardForm so it could take a while

dqii:
Oh I use CardField. I didn't realize it was supported there! Thank you. :)

PR #1150, opened 2022-09-26
feat: add support of custom fonts to card elements for Android

perqin:
## Summary
Add support of custom `fontFamily` to the CardField and CardForm elements for Android.

This PR uses `ReactTypefaceUtils.applyStyles` to load typeface from assets, which is the same as the approach in ReactTextView.

Note that the font of the dropdown options TextView in CardForm's country text input can't be changed by this PR, because those TextViews are created by [the adapter in the stripe-android library](https://github.com/stripe/stripe-android/blob/master/payments-core/src/main/java/com/stripe/android/view/CountryAdapter.kt).

## Motivation
#681 

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 


charliecruzan-stripe:
Hi @perqin ! Thanks for the contribution, sorry it's taking so long to get a review. I'll take a look this week!

dqii:
@charliecruzan-stripe any chance you'd be able to have a look this week?

android/src/main/java/com/reactnativestripesdk/CardFormView.kt
```diff
@@ -134,6 +135,12 @@ class CardFormView(context: ThemedReactContext) : FrameLayout(context) {
       cardFormViewBinding.cardMultilineWidget.expiryDateEditText,
       cardFormViewBinding.postalCode
     )
+    val inputLayoutBindings = setOf(
+      cardFormViewBinding.cardMultilineWidget.cardNumberTextInputLayout,
+      cardFormViewBinding.cardMultilineWidget.cvcInputLayout,
+      cardFormViewBinding.cardMultilineWidget.expiryTextInputLayout,
+      cardFormViewBinding.postalCodeContainer,
+    )
```
charliecruzan-stripe, line 143: I think we can simplify the code in here just a little bit more:

let's use the same bindings for fontFamily that we do for placeholderColor. So here, let's change:
```suggestion
val placeholderTextBindings = setOf(
      multilineWidgetBinding.tlExpiry,
      multilineWidgetBinding.tlCardNumber,
      multilineWidgetBinding.tlCvc,
      cardFormViewBinding.postalCodeContainer,
    )
```

Then we can change the `placeholderColor?.let` block to:
```
placeholderColor?.let {
      for (binding in placeholderTextBindings) {
        binding.defaultHintTextColor = ColorStateList.valueOf(Color.parseColor(it))
      }
    }
```

and then I'll suggest the code changes below as well

android/src/main/java/com/reactnativestripesdk/CardFormView.kt
```diff
@@ -159,9 +166,17 @@ class CardFormView(context: ThemedReactContext) : FrameLayout(context) {
       }
     }
     fontFamily?.let {
+      // Load custom font from assets, and fallback to default system font
+      val typeface = ReactTypefaceUtils.applyStyles(null, -1, -1, it.takeIf { it.isNotEmpty() }, context.assets)
       for (binding in editTextBindings) {
-        binding.typeface = Typeface.create(it, Typeface.NORMAL)
+        binding.typeface = typeface
+      }
+      for (binding in inputLayoutBindings) {
```
charliecruzan-stripe, line 174:
```suggestion
      for (binding in placeholderTextBindings) {
```

see suggestion above for context

charliecruzan-stripe (changes requested):
small changes, but then looks good to me! re request review once these changes are made please :)

charliecruzan-stripe:
Also- you can remove the font file, that font file already exists at `example/src/assets/fonts/`

android/src/main/java/com/reactnativestripesdk/CardFormView.kt
```diff
@@ -159,9 +166,17 @@ class CardFormView(context: ThemedReactContext) : FrameLayout(context) {
...
+        binding.typeface = typeface
       }
+      cardFormViewBinding.countryLayout.typeface = typeface
```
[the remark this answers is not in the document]
perqin: This is needed to also set the font for the label and error text of the TextInputLayout. You could refer to the [documentation](https://developer.android.com/reference/com/google/android/material/textfield/TextInputLayout#setTypeface(android.graphics.Typeface)).

android/src/main/java/com/reactnativestripesdk/CardFormView.kt
perqin: Changed.

android/src/main/java/com/reactnativestripesdk/CardFormView.kt
perqin: Changed.

charliecruzan-stripe (approved):

Issue #1146, opened 2022-09-22
In Android: the brand and last4  field in CardField  onCardChange callback wouldn't be set.


umer-viabletree:
**Describe the bug**

"@stripe/stripe-react-native": "^0.19.0",
 "react-native": "0.64.2",
  on Android 
  compileSdkVersion = 33
  kotlinVersion = "1.7.10" 
  on Ios 
  xcode 14. and working fine.

**To Reproduce**
on CardField onCardChange callback getting this Object:
{
brand: null
complete: true
expiryMonth: 2
expiryYear: 42
last4: null
validCVC: "Valid"
validExpiryDate: "Valid"
validNumber: "Valid"
}

and on createToken method return error image below.
NOTE:  If I enter the same Test Card on Ios it works fine.

**Expected behavior**
If card valid it should be return brand and last4 as in Ios.

![1663837525603](https://user-images.githubusercontent.com/81979880/191707509-cc360552-e93d-41d0-a4ec-26f5aa3cef72.JPEG)

![1663837525603](https://user-images.github
<img width="562" alt="Screen Shot 2022-09-22 at 2 09 54 PM" src="https://user-images.githubusercontent.com/81979880/191707017-ab46c96e-d072-4ab9-b085-8d8a629c51ed.png">
usercontent.com/81979880/191706148-00040a26-4f7f-4d26-92d9-21156b3e8a55.JPEG)

 

@charliecruzan-stripe

charliecruzan-stripe:
Hi, please provide the code to reproduce. I tested this myself and am not seeing the same behavior

SimonErlandsson:
Having the same issue. Entering 4242 4242 4242 4242 04/24 242 while in test mode yields "Card details not complete". The exact same configuration worked previously.

charliecruzan-stripe:
Please provide a reproducible demo or code sample, otherwise it's very difficult to help. Tested this in [our example app](https://github.com/stripe/stripe-react-native/blob/master/example/src/screens/WebhookPaymentScreen.tsx) and it's working correctly.

SimonErlandsson:
`
<CardField
    onCardChange={e => console.log(e)}
    postalCodeEnabled={false}
    cardStyle={{
        backgroundColor: '#FFFFFF',
        textColor: '#000000',
    }}
    style={{
        width: '100%',
        height: 50,
        marginTop: 20,
        borderRadius: 5,
        borderWidth: 1
    }}
/>
`

[...]

PR #1213, opened 2022-11-27
[Issue-1146] Fix CardField validation problem on Android

pavelglebov:
Issue: https://github.com/stripe/stripe-react-native/issues/1146

Order of react props apply is not as expected:
setCountryCode applied first, then setPostalCodeEnabled. As postalCodeEnabled is true by default, it leads to issues with postalCodeEnabled={false} on react side.

## Summary

Safely change ```mCardWidget.postalCodeRequired = false``` when ```postalCodeEnabled={false}```

## Motivation

For some reason, react props are applied to Android native element in unexpected order (countryCode first, then postalCodeEnabled), which in conjunction with default value of ```polstalCodeEnabled === true``` leads to non-working ```CardField``` on Android with ```postalCodeEnabled={false}```

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 


charliecruzan-stripe (approved):
nice catch! thanks for the PR 👍

PR #1162, opened 2022-10-05
fix: check for existence of pass on paired watch

charliecruzan-stripe:
## Summary

`canAddCardToWallet` now check specifically for the existence of passes that exist on paired device like apple watches. If a pass exists on another paired device, then the merchant **should** show the 'add to apple wallet' button. added another status to the enum for debugging/clarity.

## Motivation

Apple's guidelines around this have changed.

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 
- [ ] I have added relevant documentation for my changes.
- [x] This PR does not result in any developer-facing changes.

PR #1219, opened 2022-12-01
fix: check for existence of pass on paired watch 

charliecruzan-stripe:
## Summary

`canAddCardToWallet` now check specifically for the existence of passes that exist on paired device like apple watches if the `hasPairedAppleWatch` bool is true. If a pass exists on only one of the device and its paired device, then the merchant **should** show the 'add to apple wallet' button. added new statuses to the enum for debugging/clarity.

## Motivation

Apple's guidelines around this have changed.

## Testing

- [x] I tested this manually
- [x] I added automated tests

## Documentation

Select one: 
- [] This PR does not result in any developer-facing changes.

overrides https://github.com/stripe/stripe-react-native/pull/1162

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # CHANGELOG
 
+### Breaking changes
+
+### New features
+
+## Fixes
+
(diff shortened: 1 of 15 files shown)


grantmac:
Hi @charliecruzan-stripe 
Some questions if that's ok:
- What happens if this is set to true, and the device doesn't have a paired watch?
- Does this handle the use case where the pass exists on the iPhone but not the paired watch?

Thanks

charliecruzan-stripe:
> What happens if this is set to true, and the device doesn't have a paired watch?

The SDK would check the paired devices for the payment pass, and wouldn't find any

> Does this handle the use case where the pass exists on the iPhone but not the paired watch?

Yep!

PR #1223, opened 2022-12-02
fix(types): add id to token.card result type

walkerdb:
A PR from back in March 2022 (https://github.com/stripe/stripe-react-native/pull/863) started passing the new card ID back from the `createToken` response in native code, but it looks like the JS-land types weren't updated accordingly.

This adds it in!

## Summary
Adds missing field from JS token card response type.

## Motivation
We use this field in our code. It's present but Typescript fails because it doesn't know it's supposed to be there.

## Testing

- [ ] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/src/types/Token.ts
+++ b/src/types/Token.ts
@@ -56,6 +56,7 @@ export type BankAccountStatus =
   | 'Verified';
 
 export interface Card {
+  id: string;
   country: string;
   brand: CardBrand;
   currency?: string;


charliecruzan-stripe (approved):

charliecruzan-stripe:
thanks!

PR #1237, opened 2022-12-08
fix(types): support `--isolatedModules` typescript option

walkerdb:
The [isolatedModules](https://www.typescriptlang.org/tsconfig#isolatedModules) typescript option requires that re-exports of types must use the `type` keyword.

Without this change running `tsc` in my app fails with the following:

```
node_modules/@stripe/stripe-react-native/src/types/index.ts:52:10 - error TS1205: Re-exporting a type when the '--isolatedModules' flag is provided requires using 'export type'.

52 export { Address, BillingDetails, AddressDetails } from './Common';
            ~~~~~~~
```

## Testing

- [x] I tested this manually
- [ ] I added automated tests

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -49,7 +49,7 @@ export {
 
 export * from './PushProvisioning';
 export * from './Errors';
-export { Address, BillingDetails, AddressDetails } from './Common';
+export type { Address, BillingDetails, AddressDetails } from './Common';
 
 /**
  * @ignore


charliecruzan-stripe:
hey @walkerdb , hope you don't mind but i opened a separate PR for this to add the option to our TS config to prevent future issues like this, and also address a couple other instances of this: https://github.com/stripe/stripe-react-native/pull/1239

Issue #1306, opened 2023-02-19
android.nfc.NfcAdapter.isEnabled() - null object reference


calebmackdavenport:
**Describe the bug**

We use bugsnag for crashlytics and error reporting.
Have seen a handful of instances of this error:
![image](https://user-images.githubusercontent.com/30021449/219907034-f5c31c14-1335-48ca-9b4d-22f6dc2a1f65.png)

We are currently on:
`"@stripe/stripe-react-native": "0.22.0",`

Here is a stack trace of the error:
```
java.lang.NullPointerException: Attempt to invoke virtual method 'boolean android.nfc.NfcAdapter.isEnabled()' on a null object reference
        at com.reactnativestripesdk.pushprovisioning.PushProvisioningProxy.isNFCEnabled(PushProvisioningProxy.kt:36)
        at com.reactnativestripesdk.StripeSdkModule.canAddCardToWallet(StripeSdkModule.kt:680)
        at java.lang.reflect.Method.invoke(Method.java:-2)
        at com.facebook.react.bridge.JavaMethodWrapper.invoke(JavaMethodWrapper.java:372)
        at com.facebook.react.bridge.JavaModuleWrapper.invoke(JavaModuleWrapper.java:188)
        at com.facebook.jni.NativeRunnable.run(NativeRunnable.java:-2)
        at android.os.Handler.handleCallback(Handler.java:942)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:27)
        at android.os.Looper.loopOnce(Looper.java:240)
        at android.os.Looper.loop(Looper.java:351)
        at com.facebook.react.bridge.queue.MessageQueueThreadImpl$4.run(MessageQueueThreadImpl.java:228)
        at java.lang.Thread.run(Thread.java:1012)
```

**To Reproduce**
I have not successfully recreated the error and am not fully certain how it is presenting, other than following a call to canAddCardToWallet on a handful of android devices.

**Expected behavior**
I am uneducated on android fundamentals. I don't pretend to know anything, but I am guessing that this is crashing instead of presenting a `null` for some reason, as is suggested in the [documentation](https://developer.android.com/reference/android/nfc/NfcAdapter#getDefaultAdapter(android.content.Context)).

**Smartphone (please complete the following information):**
 - Device List:
[...]

PR #1308, opened 2023-02-21
feat: make nfc check optional on android

charliecruzan-stripe:
## Summary

Fixes a null pointer access error, and also introduces the `supportsTapToPay` field for `canAddCardToWallet()`. This field defaults to true, but when set to false, it will disable the check for NFC compatibility and allow merchants to provision cards on devices without tap to pay functionality

## Motivation
closes https://github.com/stripe/stripe-react-native/issues/1306

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 


mrgithub:
Does this mean we can create apps to collect payments with tap to pay without needing a separate card reader terminal?

PR #1347, opened 2023-04-05
feat: add isThreeDSecureSupported on CardResult type

huextrat:
## Summary
Added `isThreeDSecureSupported` to the `CardResult` type.

This field uses `threeDSecureUsage` to indicate whether the card supports 3DS or not.

## Motivation
Native SDK has this information, but the bridge does not pass this data to the RN package.

This information can be useful to improve the user experience by indicating to the user that a validation from their bank may be required `isThreeDSecureSupported` is `true`.

## Testing

- [X] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/utils/Mappers.kt
+++ b/android/src/main/java/com/reactnativestripesdk/utils/Mappers.kt
@@ -346,8 +346,8 @@ internal fun mapFromPaymentMethod(paymentMethod: PaymentMethod): WritableMap {
     it.putString("fingerprint", paymentMethod.card?.fingerprint)
     it.putString("preferredNetwork", paymentMethod.card?.networks?.preferred)
     it.putArray("availableNetworks", paymentMethod.card?.networks?.available?.toList() as? ReadableArray)
-    it.putMap("threeDSecureUsage", WritableNativeMap().also {
-      it.putBoolean("isSupported", paymentMethod.card?.threeDSecureUsage?.isSupported ?: false)
+    it.putMap("threeDSecureUsage", WritableNativeMap().also { threeDSecureUsageMap ->
+      threeDSecureUsageMap.putBoolean("isSupported", paymentMethod.card?.threeDSecureUsage?.isSupported ?: false)
     })
   })
   pm.putMap("SepaDebit", WritableNativeMap().also {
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/reactnativestripesdk/utils/Mappers.kt
```diff
@@ -346,6 +346,7 @@ internal fun mapFromPaymentMethod(paymentMethod: PaymentMethod): WritableMap {
     it.putString("fingerprint", paymentMethod.card?.fingerprint)
     it.putString("preferredNetwork", paymentMethod.card?.networks?.preferred)
     it.putArray("availableNetworks", paymentMethod.card?.networks?.available?.toList() as? ReadableArray)
+    it.putBoolean("isThreeDSecureSupported", paymentMethod.card?.threeDSecureUsage?.isSupported ?: false)
```
charliecruzan-stripe, line 349: let's match the native libraries and make this a dictionary, like this:
```
it.putMap("ThreeDSecureUsage", WritableNativeMap().also { threeDSecureUsageMap ->
  it.putBoolean("isSupported", paymentMethod.card?.threeDSecureUsage?.isSupported ?: false)
})
```

charliecruzan-stripe (changes requested):

android/src/main/java/com/reactnativestripesdk/utils/Mappers.kt
huextrat: Agree with that even it looks like the native type will never be update, it's coherent to be in sync

android/src/main/java/com/reactnativestripesdk/utils/Mappers.kt
```diff
@@ -346,6 +346,9 @@ internal fun mapFromPaymentMethod(paymentMethod: PaymentMethod): WritableMap {
     it.putString("fingerprint", paymentMethod.card?.fingerprint)
     it.putString("preferredNetwork", paymentMethod.card?.networks?.preferred)
     it.putArray("availableNetworks", paymentMethod.card?.networks?.available?.toList() as? ReadableArray)
+    it.putMap("threeDSecureUsage", WritableNativeMap().also {
+      it.putBoolean("isSupported", paymentMethod.card?.threeDSecureUsage?.isSupported ?: false)
```
charliecruzan-stripe, line 350:
```suggestion
    it.putMap("threeDSecureUsage", WritableNativeMap().also { threeDSecureUsageMap ->
      threeDSecureUsageMap.putBoolean("isSupported", paymentMethod.card?.threeDSecureUsage?.isSupported ?: false)
```

i think the `it` here will clash with the outer `it`, let's provide an explicit variable name

charliecruzan-stripe (changes requested):

charliecruzan-stripe (approved):

PR #1367, opened 2023-04-24
fix: send shippingContact with createPaymentMethod

p-sebastian:
## Summary
`createPlatformPayPaymentMethod` wasn't returning the `shippingContact` information which contains the address. 

this modifies both `createPlatformPayPaymentMethod` and `presentApplePay` to pass along the `shippingContact`

## Motivation
both `createPlatformPayPaymentMethod & presentApplePay` allow **confirming** the `paymentIntent` **after** the user has passed the biometric auth during apple pay, which allows getting the full address (that contains the `street`), this is useful for setting options on the server before the `paymentIntent` is confirmed with `confirmApplePayPayment || stripe.confirmPayment`
___
In our case we run hooks that trigger on `paymentIntentConfirm` which block setting the address after the fact (for security of an order) which is a problem on `confirmPlatformPayPayment` since it confirms without setting an address (a bug on `StripeApplePay` Pod. but even if that were fixed, it sets the address during confirmation which causes the same problem with the hook on our use case

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/ios/ApplePayViewController.swift
+++ b/ios/ApplePayViewController.swift
@@ -33,10 +33,18 @@ extension StripeSdk : PKPaymentAuthorizationViewControllerDelegate, STPApplePayC
                 if let error = error {
                     self.createPlatformPayPaymentMethodResolver?(Errors.createError(ErrorType.Failed, error))
                 } else {
-                    let promiseResult = [
-                        "paymentMethod": Mappers.mapFromPaymentMethod(paymentMethod?.splitApplePayAddressByNewline()) ?? [:]
-                    ]
-                    self.createPlatformPayPaymentMethodResolver?(promiseResult)
+                    let mapped = Mappers.mapFromPaymentMethod(paymentMethod?.splitApplePayAddressByNewline()) ?? [:]
+                    if let shippingContact = payment.shippingContact {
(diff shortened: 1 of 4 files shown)


charliecruzan-stripe:
thanks @p-sebastian ! Can I ask- why not use the `PlatformPayButton`'s `onShippingContactSelected` method instead?

charliecruzan-stripe:
> The problem is that unless you do the biometric (faceID/TouchID), the street won't be included in the postalAddress, same for the name | phone. only after it will you get all the data. And that listener doesn't get called again after the biometric pass.

Makes sense. I'm going to hold off on merging this since I want to add this same support to the Android side with Google Pay

iamjoncannon:
@charliecruzan-stripe would it be possible to merge this without Android coverage? This unblocks a business requirement for my team, and I'm sure there are others (as discussed, if you have a checkout flow where a payment method is created before a payment intent, theres no way to receive the user's full shipping address without this commit).  Thanks!

charliecruzan-stripe:
fixed in https://github.com/stripe/stripe-react-native/pull/1500

PR #1375, opened 2023-05-01
bump Android to 20.24.+ and iOS to 20.7.0

ccen-stripe:
## Summary

[Android](https://github.com/stripe/stripe-android/releases/tag/v20.24.1): Pick up the latest minor update fixing the [dagger duplicate name](https://github.com/stripe/stripe-android/pull/6602) issue

[iOS](https://github.com/stripe/stripe-ios/releases/tag/23.7.0): Bug fixes and test mode for Identity

Related change: https://github.com/stripe/stripe-identity-react-native/pull/117

## Testing

- [ ] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 
- [ ] I have added relevant documentation for my changes.
- [ ] This PR does not result in any developer-facing changes.

PR #1376, opened 2023-05-02
chore: upgrade native libraries

charliecruzan-stripe:
## Summary
supersedes https://github.com/stripe/stripe-react-native/pull/1375/

upgrade android to 20.24+
upgrade ios to 23.7.+


stripe-react-native.podspec
```diff
@@ -1,7 +1,7 @@
 require 'json'
 
 package = JSON.parse(File.read(File.join(__dir__, 'package.json')))
-stripe_version = '23.6.0'
+stripe_version = '~> 23.7.0'
```
ccen-stripe, line 4: any reason why we don't use `23.7.0` ?

ccen-stripe (changes requested):

stripe-react-native.podspec
```diff
@@ -1,7 +1,7 @@
 require 'json'
 
 package = JSON.parse(File.read(File.join(__dir__, 'package.json')))
-stripe_version = '23.6.0'
```
ccen-stripe, line 4: Also, could we add a line here
```
When stripe_version is updated, also need to update stripe_version in https://github.com/stripe/stripe-identity-react-native/blob/main/stripe-identity-react-native.podspec
```

android/gradle.properties
```diff
@@ -1,2 +1,2 @@
 StripeSdk_kotlinVersion=1.8.0
-StripeSdk_stripeVersion=20.23.+
+StripeSdk_stripeVersion=20.24.+
```
ccen-stripe, line 2: Could we add a line here:
```
When StripeSdk_kotlinVersion and StripeSdk_stripeVersion is updated, also need to update StripeIdentityReactNative_kotlinVersion and StripeIdentityReactNative_stripeVersion in https://github.com/stripe/stripe-identity-react-native/blob/main/android/gradle.properties
```

stripe-react-native.podspec
charliecruzan-stripe: this matches up to but not including 23.8.0, so we'll get in any patch fixes that `stripe-ios` releases. it's the equivalent of the android's `20.24.+`

ccen-stripe: ok, will apply the similar change in identity

ccen-stripe (changes requested):

Issue #1453, opened 2023-07-13
canAddCardToWallet not returning virtualised last 4 digits 


grantmac:
**Describe the bug**
When calling `canAddCardToWallet` the returned details object contains `token.cardLastFour`. `cardLastFour` is the last 4 digits of the card as provided by Stripe rather than the last 4 digits of the virtualised card inside the google/apple wallet. 

Looking at **TapAndPayProxy.kt** suggests that the virtualised last 4 should be returned. 

**To Reproduce**
1. Call `canAddCardToWallet` with a card that has been added to the device wallet
2. Inspect the returned details object. 

**Expected behavior**
That `token.cardLastFour should be the last 4 digits of the tokenised card. 

**Screenshots**
N/A

**Smartphone (please complete the following information):**
 - Device: Android Pixel 6a
 - OS: Anroid13

PR #1466, opened 2023-08-04
fix: return both last four types from google tokenized cards

charliecruzan-stripe:
fixes https://github.com/stripe/stripe-react-native/issues/1453

`canAddCardToWallet` now returns the `fpanLastFour` and `dpanLastFour` in the `GooglePayCardToken` type. The `cardLastFour` field (which is now superseded by `fpanLastFour`) is deprecated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+**Features**
+
+- `canAddCardToWallet` now returns the `fpanLastFour` and `dpanLastFour` in the `GooglePayCardToken` type. The `cardLastFour` field (which is now superseded by `fpanLastFour`) is deprecated.
+
 **Fixes**
 
 - Fixed an error on Android where `AddressSheet` would throw an error if submitted with the keyboard open. [#1462](https://github.com/stripe/stripe-react-native/pull/1462)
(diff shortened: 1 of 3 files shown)


grantmac:
Awesome @charliecruzan-stripe ! any plans to do the same for iOS?

charliecruzan-stripe:
@grantmac are you grabbing this value for display purposes? We return it on Android since you need the `token` object for an edge case, but on iOS that edge case doesn't exist. I haven't checked the iOS APIs to make sure that both of these are available, so I'm not sure that it's doable, but I'd just like to know more about your use case

grantmac:
@charliecruzan-stripe we need the dpanLastFour when trying to get receipts from some retailers on behalf of the user. I've implemented this myself by patching the module.

grantmac:
Hey @charliecruzan-stripe - just picking this up again if that's ok. My solution isn't working great and wondered if you might be able to think of a better implementation. 

Again, our use case is that we (as the platform) need to know the device pan as to communicate with merchants when requesting invoices.

charliecruzan-stripe:
what does your patch look like @grantmac? I'm not sure what the relevant iOS API for this is, at first glance im thinking its `deviceAccountNumberSuffix` - https://developer.apple.com/documentation/passkit/pksecureelementpass/3543361-deviceaccountnumbersuffix

Issue #1484, opened 2023-08-29
`onShippingContactSelected` is not available in `usePlatformPay`


huextrat:
`useApplePay` has been removed, so the callback `onShippingContactSelected` is no longer available except if you are using `PlatformPayButton`.

Would be super useful to keep both callback `onShippingMethodSelected` and `onShippingContactSelected`  when we are using a custom button to start native payment with `usePlatformPay`.

If there is a way to do this, I can help with the PR.

huextrat:
Our use case is pretty simple, we need to retrieve `shippingContact` in particular the postal address before payment is made.

I would be glad to use the native button, but for now there is missing styling props. This is how it looks when user presses the `PlatformPayButton`, that opacity during the `onPress` is not acceptable for my app.

![Capture d’écran 2023-08-30 à 21 41 27](https://github.com/stripe/stripe-react-native/assets/22995315/14d8bbea-a961-4ea6-91ff-76d68c906f79)

I'll take a look on `ApplePayButtonView.swift` and I'll try to contribute to make it more customizable.

huextrat:
@charliecruzan-stripe Just made a PR https://github.com/stripe/stripe-react-native/pull/1487 to make the opacity during `onPress`

Looks kinda hacky, but it seems that Apple does not provide any API to custom style except the provided one :/

charliecruzan-stripe:
So you'd apply your own custom opacity to the button? That would run the risk of app store rejection since the Apple Pay button needs to match Apple's branding guidelines

PR #1487, opened 2023-08-30
feat: allow ApplePayButtonNative usage with RN `onPress` opacity only

huextrat:
## Summary
Allow dev to use `PlatformPayButton` with opacity from RN TouchableOpacity and not from the native Apple Pay button.

## Motivation
The native opacity from Apple Pay button is not pretty.

## Testing

- [X] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

fixes: https://github.com/stripe/stripe-react-native/issues/1484

### DIFF
--- a/src/components/PlatformPayButton.tsx
+++ b/src/components/PlatformPayButton.tsx
@@ -32,6 +32,8 @@ export interface Props extends AccessibilityProps {
   onPress(): void;
   /** Set to `true` to disable the button from being pressed & apply a slight opacity to indicate that it is unpressable. Defaults to false. */
   disabled?: boolean;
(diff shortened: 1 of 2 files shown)


src/components/PlatformPayButton.tsx
```diff
@@ -144,7 +147,6 @@ export function PlatformPayButton({
   return (
     <TouchableOpacity
       disabled={disabled}
-      activeOpacity={disabled ? 0.3 : 1}
```
huextrat, line 147: Not sure why there is an `activeOpacity` from my POV there is no need, but you may have an explanation ;)

PR #1497, opened 2023-09-12
Update example app to use two activities

### LINKED ISSUE #355: [Android] 3DSecure confirmation screen closed if app is put in background
Hi all. 
We currently have a bug happening on Android only. 
After using `confirmPaymentMethod` on a payment, the 3DS screen is brought up. If instead of confirming / denying the payment on this screen we go into background and then bring the app back to active again, 3DS screen is just dismissed, so the app is stuck. 
Obviously, at that moment Stripe payment is pending, but there's no way to bring up the payment screen again. 
Even if there was, then the process would be repeated - if the user has to go to the bank's app in order to confirm the payment or whatever, the problem would come back and always be present.

Note: this does not happen on iOs. 

**To Reproduce**
Steps to reproduce the behaviour:
1. Boot Android React Native app. 
2. Confirm payment with a 3DS card. 
3. When the `hooks.stripe.com` screen with the confirmation gets brought up, go into background. 
4. Wait 2 seconds, open the app again. 
5. There you have a problem. 

**Expected behavior**
The expected behaviour would be the same as in iOs, for this screen not to be dismissed until the user does some action upon it. 

**Desktop (please complete the following information):**
OS: MacOs Big Sur 11.2.3
Browser: Chrome
Version: 91.0.4472.77 (Official Build) (x86_64)

Device: Honor 9 Lite
Android version: 9

**Additional context**
Stripe version: 0.1.4
React native version: 0.63.4

Thanks in advance!

stianjensen:
Using singleTask in an Android application for the activity that has an intent-filter on the LAUNCHER, means that any activities running on top of this activity will disappear when the app's icon is pressed again from the launcher.

This is a problem, because when confirming a SetupIntent or PaymentIntent, Stripe may need to present a web browser (its own activity) for verification with a customer's bank or similar, and the user may even need to leave this web browser activity temporarily to navigate to their bank's own application. If the user is not careful when navigating back to the original app, and taps the app's icon from the launcher, the app will open to its original activity, and the web browser is gone. Instead, the user will need to navigate back only using Android's app switcher, which will correctly bring you back into the open web browser activity.

The change in this commit side-steps this issue by splitting the example application into two acitivites. One that is only responsible for launching the application, and a second activity that actually represents the react native application, and will be started by the first activity, ensuring to only start it once.

This fixes #355.

## Testing

- [ ] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/example/android/app/src/main/AndroidManifest.xml
+++ b/example/android/app/src/main/AndroidManifest.xml
@@ -15,6 +15,13 @@
       <meta-data
         android:name="com.google.android.gms.wallet.api.enabled"
         android:value="true" />
+      <activity android:name=".LaunchActivity" android:exported="true">
+        <intent-filter>
+          <action android:name="android.intent.action.MAIN" />
+          <category android:name="android.intent.category.LAUNCHER" />
+        </intent-filter>
+      </activity>
+
       <activity
         android:name=".MainActivity"
         android:label="@string/app_name"
@@ -22,11 +29,6 @@
         android:launchMode="singleTask"
         android:windowSoftInputMode="adjustResize"
         android:exported="true">
-        <intent-filter>
-            <action android:name="android.intent.action.MAIN" />
-            <category android:name="android.intent.category.LAUNCHER" />
-        </intent-filter>
-
         <intent-filter>
           <action android:name="android.intent.action.VIEW" />
           <category android:name="android.intent.category.DEFAULT" />
--- a/example/android/app/src/main/java/com/example/reactnativestripesdk/LaunchActivity.java
+++ b/example/android/app/src/main/java/com/example/reactnativestripesdk/LaunchActivity.java
@@ -1,0 +1,20 @@
+package com.example.reactnativestripesdk;
+
+import android.os.Bundle;
+import android.app.Activity;
+import android.content.Intent;
+
+public class LaunchActivity extends Activity {
+
+  @Override
+  protected void onCreate(Bundle savedInstanceState) {
+    super.onCreate(savedInstanceState);
+    MainApplication application = (MainApplication) getApplication();
+    // check that MainActivity is not started yet
+    if (!application.isActivityInBackStack(MainActivity.class)) {
+      Intent intent = new Intent(this, MainActivity.class);
+      startActivity(intent);
+    }
+    finish();
+  }
+}
--- a/example/android/app/src/main/java/com/example/reactnativestripesdk/MainActivity.java
+++ b/example/android/app/src/main/java/com/example/reactnativestripesdk/MainActivity.java
@@ -15,6 +15,18 @@ public class MainActivity extends ReactActivity {
     return "main";
   }
 
+  @Override
+  protected void onCreate(Bundle savedInstanceState) {
+    super.onCreate(savedInstanceState);
+    ((MainApplication) getApplication()).addActivityToStack(this.getClass());
+  }
+
+  @Override
+  protected void onDestroy() {
+    super.onDestroy();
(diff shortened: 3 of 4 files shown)


stianjensen:
I based this on the fix originally described here:
https://github.com/proyecto26/react-native-inappbrowser/issues/213#issuecomment-878915862

It may be that the code can be further simplified, by not even maintaining a list, since in my experiments, I've never seen that it's possible for the activity to be started more than once (which makes sense, since it has launchMode 'singleTask). But for now I kept it like this, since it's equivalent to the code we ended up deploying to production.

zdnk:
what about expo apps using managed workflow?

example/android/app/src/main/AndroidManifest.xml
```diff
@@ -15,18 +15,20 @@
...
+          <action android:name="android.intent.action.MAIN" />
+          <category android:name="android.intent.category.LAUNCHER" />
+        </intent-filter>
+      </activity>
+
       <activity
         android:name=".MainActivity"
         android:label="@string/app_name"
         android:configChanges="keyboard|keyboardHidden|orientation|screenSize|uiMode"
         android:launchMode="singleTask"
```
huextrat, line 29: Replacing `singleTask` by `singleTop` may improve the app compatibility with deep linking.

Opening a deep link with that `launchMode` will cause the app to be restarted and may cause the lost of the current state.

stianjensen: singleTask is what is generally recommended (and the only thing that seems to work well in practice) by Branch.io for instance. As long as the activity that handles deeplinks is singleTask, but the one that handles the launcher intent is not, both the Stripe use case, and the Branch use case seem to work well in the same app at the same time.

stianjensen:
> what about expo apps using managed workflow?

I'm not sure, and not really that familiar with android activites except what I've been forced to try to understand while debugging and fixing the issue in our react native app.

I would assume that the issue is something expo would need to support in their native code – but unsure if they could do it without risking breaking some other use case.

I still think there's not much Stripe could do about it without abandoning the use of chrome custom tabs altogether though (and that's not really an option for 3d secure flows either).

charliecruzan-stripe (approved):

example/android/app/src/main/AndroidManifest.xml
```diff
@@ -15,18 +15,20 @@
       <meta-data
         android:name="com.google.android.gms.wallet.api.enabled"
         android:value="true" />
+      <activity android:name=".LaunchActivity" android:exported="true">
```
huextrat, line 18: To handle deeplink `.LaunchActivity` should be `singleTop` otherwise it may cause two instances of the app

```suggestion
      <activity android:launchMode="singleTop" android:name=".LaunchActivity" android:exported="true">
```

stianjensen: The trick with this split into two activities is that deeplinks should be handled by MainActivity directly, not LaunchActivity. MainActivity can then be singleTask or singleTop depending on what you prefer.

huextrat: Yes, and kept deeplinks handled by `MainActivity` with `singleTask` launchMode (required to prevent issue with 3DS) will cause multiple instances. Using `android:launchMode="singleTop"` for `LaunchActivity` will prevent the multiple instances, as the deeplink will route the intent to the existing instance instead of creating a new one.

But I agree that it may depend on the use case you have to handle deeplink. This change is not required, but may be necessary depending on the project configuration.

Issue #1518, opened 2023-09-28
`phoneNumber` from GooglePay ShippingContact is null


huextrat:
**Describe the bug**
`phoneNumber` on GooglePay ShippingContact is always null 

**To Reproduce**
Make a payment with GooglePay, and check `shippingContact` from `createPlatformPayPaymentMethod` result.

**Expected behavior**
If phone number is required it should not be null.

PR #1519, opened 2023-09-28
fix: phone number is in shippingInformation

huextrat:
## Summary
Get `phoneNumber` from `shippingInformation` to avoid null value if required

## Testing

- [X] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

fixes: https://github.com/stripe/stripe-react-native/issues/1518

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/utils/Mappers.kt
+++ b/android/src/main/java/com/reactnativestripesdk/utils/Mappers.kt
@@ -898,7 +898,7 @@ internal fun mapFromShippingContact(googlePayResult: GooglePayResult): WritableM
   googlePayResult.name
   name.putString("givenName", googlePayResult.shippingInformation?.name)
   map.putMap("name", name)
-  map.putString("phoneNumber", googlePayResult.phoneNumber)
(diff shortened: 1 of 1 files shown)


charliecruzan-stripe:
Hm, the value from `googlePayResult.shippingInformation?.phone` was null in my testing which is why I relied on this value instead. Let's load the current value, and if null, fall back to this one

huextrat:
> Hm, the value from `googlePayResult.shippingInformation?.phone` was null in my testing which is why I relied on this value instead. Let's load the current value, and if null, fall back to this one

@charliecruzan-stripe I made an update. According to Stripe Android SDK you are right `phoneNumber` is also existing in `GooglePayResult`. 

I just did the reverse logic and tried to retrieve the phone from `shippingInformation` first, as in my opinion it makes more sense to retrieve the phone number entered in the shipping contact form first and then fallback to the second phone number which comes from who knows where 😅

Let me know if it suits you or not ;)

charliecruzan-stripe (approved):

charliecruzan-stripe:
thanks @huextrat

Issue #1520, opened 2023-09-28
`street` in ShippingContact may contain null value with GooglePay


huextrat:
**Describe the bug**
`street` in ShippingContact for Google Pay is a concatenation of `address.line1` & `address.line2`. `line2` is null in many cases so a simple concatenation will show `null` in the street name. It should be handle in mapper.

**To Reproduce**
- Make a payment with GooglePay and a required shipping address.
- Shipping street address should contain only one line

**Expected behavior**
Get only `line1` if `line2` is null.

PR #1521, opened 2023-09-28
fix: prevent null in street name

huextrat:
## Summary
Handle case where `line2` is null to not display `null` during the concatenation.

There are many cases to do this in Kotlin, let me know if you prefer something else ;)

## Motivation
I want to get a street address with the right formatting.

## Testing

- [X] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

fixes: https://github.com/stripe/stripe-react-native/issues/1520

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/utils/Mappers.kt
+++ b/android/src/main/java/com/reactnativestripesdk/utils/Mappers.kt
@@ -904,7 +904,12 @@ internal fun mapFromShippingContact(googlePayResult: GooglePayResult): WritableM
   postalAddress.putString("country", googlePayResult.shippingInformation?.address?.country)
   postalAddress.putString("postalCode", googlePayResult.shippingInformation?.address?.postalCode)
   postalAddress.putString("state", googlePayResult.shippingInformation?.address?.state)
-  postalAddress.putString("street", googlePayResult.shippingInformation?.address?.line1 + "\n" + googlePayResult.shippingInformation?.address?.line2)
+  val line1: String? = googlePayResult.shippingInformation?.address?.line1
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/reactnativestripesdk/utils/Mappers.kt
```diff
@@ -904,7 +904,12 @@ internal fun mapFromShippingContact(googlePayResult: GooglePayResult): WritableM
   postalAddress.putString("country", googlePayResult.shippingInformation?.address?.country)
   postalAddress.putString("postalCode", googlePayResult.shippingInformation?.address?.postalCode)
   postalAddress.putString("state", googlePayResult.shippingInformation?.address?.state)
-  postalAddress.putString("street", googlePayResult.shippingInformation?.address?.line1 + "\n" + googlePayResult.shippingInformation?.address?.line2)
+  var line1: String? = googlePayResult.shippingInformation?.address?.line1
+  var line2: String? = googlePayResult.shippingInformation?.address?.line2
+  var street =
```
charliecruzan-stripe, line 909:
```suggestion
  val line1: String? = googlePayResult.shippingInformation?.address?.line1
  val line2: String? = googlePayResult.shippingInformation?.address?.line2
  val street =
```

charliecruzan-stripe (changes requested):

android/src/main/java/com/reactnativestripesdk/utils/Mappers.kt
huextrat: 👍

charliecruzan-stripe (approved):
thanks again @huextrat

PR #1552, opened 2023-11-02
Update build.gradle with namespace

Joatin:
This commit adds support for RN 0.73 android builds

## Summary
Add namespace to android section of build.gradle, this is a requirement for Gradle 8 which is the default for RN 0.73+. It's wrapped in a version check for backwards compatability.

## Motivation
Android won't build without it

## Testing
- [ x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 
- [ x] This PR does not result in any developer-facing changes.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -27,6 +27,11 @@ def getExtOrIntegerDefault(name) {
 }
 
 android {
+  def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION
+  if (agpVersion.tokenize('.')[0].toInteger() >= 7) {
+    namespace "com.reactnativestripesdk"
+  }
+  
   compileSdkVersion getExtOrIntegerDefault('compileSdkVersion')
   defaultConfig {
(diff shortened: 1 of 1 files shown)


huextrat (approved):

android/build.gradle
```diff
@@ -27,6 +27,11 @@ def getExtOrIntegerDefault(name) {
 }
 
 android {
+  def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION
+  if (agpVersion.tokenize('.')[0].toInteger() >= 7) {
```
charliecruzan-stripe, line 31: I think this could be an issue on < AGP 7.3, since that's when Google stated you can start using the `namespace` DSL property. Maybe it's better to just change 7 -> 8?

charliecruzan-stripe (changes requested):
thanks for taking care of this! one question

janicduplessis:
This had been added already

PR #1563, opened 2023-11-16
fix: RN 73 compatibility

felipecsl:
## Summary
Add new mandatory `android.namespace` config.

## Motivation
https://github.com/react-native-community/discussions-and-proposals/issues/671

## Testing
App builds with RN 73 and AGP 8

## Documentation

Select one: 

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -27,6 +27,10 @@ def getExtOrIntegerDefault(name) {
 }
 
 android {
+  def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION
+  if (agpVersion.tokenize('.')[0].toInteger() >= 7) {
+    namespace "com.stripe.reactnative"
+  }
   compileSdkVersion getExtOrIntegerDefault('compileSdkVersion')
   defaultConfig {
     minSdkVersion 21
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -28,7 +28,7 @@ def getExtOrIntegerDefault(name) {
 
 android {
   def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION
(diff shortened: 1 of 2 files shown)


android/build.gradle
```diff
@@ -27,6 +27,10 @@ def getExtOrIntegerDefault(name) {
 }
 
 android {
+  def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION
+  if (agpVersion.tokenize('.')[0].toInteger() >= 7) {
```
charliecruzan-stripe, line 31: I think this could be an issue on < AGP 7.3, since that's when Google stated you can start using the namespace DSL property (correct me if im wrong though). Maybe it's better to just change 7 -> 8?

charliecruzan-stripe (changes requested):

android/build.gradle
felipecsl: could also check for major version >= 7 and minor >= 3? would be more exact that way

charliecruzan-stripe (approved):

android/src/main/AndroidManifest.xml
```diff
@@ -1,3 +1,2 @@
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-          package="com.reactnativestripesdk">
```
huextrat, line 2: For better backward compatibility, I recommend keeping package name here but lowering the gradle version to 7 `if (agpVersion.tokenize('.')[0].toInteger() >= 7) {`

This is what has been done for the most part on RN lib

huextrat:
@charliecruzan-stripe This is an old PR, but it would be useful to have it on a future version

sfriedman-stripe:
We support a more recent RN version nowadays. Please reopen if you have any issues

PR #1624, opened 2024-03-14
chore: update native deps

charliecruzan-stripe:
## Summary

Just updating underlying ios and android libraries to the most recent versions to get bug fixes and new PMs


davidme-stripe (commented):
Approved

PR #1653, opened 2024-05-04
feat: :zap: Update payment details after calling initPaymentSheet

minawalphonce:
## Summary
- created a  new function "updatePaymentSheet" which updates the intent config info (e.g. amount) after payment sheet is initialized 

## Motivation
updating payment sheet intent config, is needed for payment methods that requires second level of authentication. 
so the amount can be displayed correctly on the 3rd party authentication. 
the functionality already exists in IOS and android native code. the PR is to expose the same in react native sdk. 

## Testing
- [X] I tested this manually
- [ ] I added automated tests

## Documentation
new screen updated in the example app to test the scenario. 

### step 1
in your component, use the hook useStripe to retrieve the new "updatePaymentSheet" function. 
```
const {
    updatePaymentSheet,
  } = useStripe();
```

### step2
call the updatePaymentSheet with the `PaymentSheet.IntentConfiguration` object containing all the params  
```
const { error, paymentOption } = await updatePaymentSheet({
        confirmHandler:  ...,
        mode: {
          amount: amount,
          currencyCode: 'SEK',
        },
      });
```
NOTE: the function replace the entier block, so be aware if you remove an attribute it will be removed. 

Select one: 

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt
+++ b/android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt
@@ -292,6 +292,29 @@ class PaymentSheetFragment(
     flowController?.confirm()
   }
 
+  fun updateIntentConfiguration(bundle: Bundle, promise: Promise){
+    val onFlowControllerConfigure = PaymentSheet.FlowController.ConfigCallback { _, _ ->
+      val result = flowController?.getPaymentOption()?.let {
+        val bitmap = getBitmapFromVectorDrawable(context, it.drawableResourceId)
+        val imageString = getBase64FromBitmap(bitmap)
+        val option: WritableMap = WritableNativeMap()
+        option.putString("label", it.label)
+        option.putString("image", imageString)
+        createResult("paymentOption", option)
+      } ?: run {
+        WritableNativeMap()
+      }
+      promise.resolve(result)
+    }
+
+    this.intentConfiguration =  buildIntentConfiguration(bundle);
+    this.flowController?.configureWithIntentConfiguration(
+      intentConfiguration = this.intentConfiguration!!,
+      configuration = paymentSheetConfiguration,
+      callback = onFlowControllerConfigure
+    );
+  }
+
   private fun configureFlowController() {
     val onFlowControllerConfigure = PaymentSheet.FlowController.ConfigCallback { _, _ ->
       val result = flowController?.getPaymentOption()?.let {
--- a/example/server/index.ts
+++ b/example/server/index.ts
@@ -710,14 +710,15 @@ app.post('/payment-intent-for-payment-sheet', async (req, res) => {
 
   try {
     const paymentIntent = await stripe.paymentIntents.create({
-      amount: 5099,
-      currency: 'usd',
+      amount: req.body.amount || 5099,
+      currency: req.body.currency || 'usd',
       payment_method: req.body.paymentMethodId,
       customer: req.body.customerId,
     });
 
     return res.send({ clientSecret: paymentIntent.client_secret });
   } catch (e) {
+    console.log(e);
     return res.send({ error: e });
   }
 });
--- a/example/src/screens/PaymentSheetDeferredIntentMultiStepScreenWithUpdates.tsx
+++ b/example/src/screens/PaymentSheetDeferredIntentMultiStepScreenWithUpdates.tsx
@@ -1,0 +1,311 @@
+import React, { useState } from 'react';
+import { Alert, StyleSheet, View, TextInput } from 'react-native';
+import {
+  useStripe,
+  // Address,
+  // BillingDetails,
+  PaymentMethod,
+  PaymentSheet,
+} from '@stripe/stripe-react-native';
+import { colors } from '../colors';
+import Button from '../components/Button';
+import PaymentScreen from '../components/PaymentScreen';
+import { API_URL } from '../Config';
+
+export default function PaymentSheetDeferredIntentMultiStepScreenWithUpdates() {
+  const {
+    initPaymentSheet,
+    presentPaymentSheet,
+    confirmPaymentSheetPayment,
+    updatePaymentSheet,
+  } = useStripe();
+  const [paymentSheetEnabled, setPaymentSheetEnabled] = useState(false);
+  const [loading, setLoading] = useState(false);
+  const [paymentMethodOption, setPaymentMethodOption] = useState<{
+    image: string;
+    label: string;
+  } | null>(null);
+
+  const [amount, setAmount] = useState(6099);
+
+  const fetchPaymentSheetParams = async () => {
+    const response = await fetch(`${API_URL}/payment-sheet`, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    });
+    const { customer, ephemeralKey } = await response.json();
+
+    return {
+      customer,
+      ephemeralKey,
+    };
+  };
+
+  const initialisePaymentSheet = async () => {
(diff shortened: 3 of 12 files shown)


android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt
```diff
@@ -292,6 +292,29 @@ class PaymentSheetFragment(
...
+        val imageString = getBase64FromBitmap(bitmap)
+        val option: WritableMap = WritableNativeMap()
+        option.putString("label", it.label)
+        option.putString("image", imageString)
+        createResult("paymentOption", option)
+      } ?: run {
+        WritableNativeMap()
+      }
+      promise.resolve(result)
+    }
```
charliecruzan-stripe, line 308: looks like this is copied from above, maybe we can extract this out to a helper function?

android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt
```diff
@@ -292,6 +292,29 @@ class PaymentSheetFragment(
...
+
+    this.intentConfiguration =  buildIntentConfiguration(bundle);
```
charliecruzan-stripe, line 310:
```suggestion
    this.intentConfiguration = buildIntentConfiguration(bundle);
```

android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt
```diff
@@ -292,6 +292,29 @@ class PaymentSheetFragment(
...
+    this.flowController?.configureWithIntentConfiguration(
+      intentConfiguration = this.intentConfiguration!!,
```
charliecruzan-stripe, line 312: `buildIntentConfiguration` returns a nullable value so this `!!` seems dangerous, maybe we can add some null-handling before this

example/src/screens/PaymentSheetDeferredIntentMultiStepScreenWithUpdates.tsx
```diff
@@ -0,0 +1,311 @@
+import React, { useState } from 'react';
+import { Alert, StyleSheet, View, TextInput } from 'react-native';
+import {
+  useStripe,
+  // Address,
+  // BillingDetails,
```
charliecruzan-stripe, line 6:
```suggestion
```

example/src/screens/PaymentSheetDeferredIntentMultiStepScreenWithUpdates.tsx
```diff
@@ -0,0 +1,311 @@
...
+    return {
+      customer,
+      ephemeralKey,
+    };
+  };
+
+  const initialisePaymentSheet = async () => {
+    setLoading(true);
+
+    try {
```
charliecruzan-stripe, line 49: i think we can factor out this try catch

example/src/screens/PaymentSheetDeferredIntentMultiStepScreenWithUpdates.tsx
```diff
@@ -0,0 +1,311 @@
...
+        customFlow: true,
+        customerId: customer,
+        customerEphemeralKeySecret: ephemeralKey,
+        merchantDisplayName: 'Example Inc.',
+        applePay: {
+          merchantCountryCode: 'US',
+        },
+        googlePay: {
+          merchantCountryCode: 'US',
+          testEnv: __DEV__,
```
charliecruzan-stripe, line 62:
```suggestion
          testEnv: true,
```

example/server/index.ts
```diff
@@ -710,14 +710,15 @@ app.post('/payment-intent-for-payment-sheet', async (req, res) => {
...
-      currency: 'usd',
+      amount: req.body.amount || 5099,
+      currency: req.body.currency || 'usd',
       payment_method: req.body.paymentMethodId,
       customer: req.body.customerId,
     });
 
     return res.send({ clientSecret: paymentIntent.client_secret });
   } catch (e) {
+    console.log(e);
```
charliecruzan-stripe, line 721: remove `console.log` here and in the other introduced locations

example/src/screens/PaymentSheetDeferredIntentMultiStepScreenWithUpdates.tsx
```diff
@@ -0,0 +1,311 @@
...
+      console.log('error', error);
+    } finally {
+      setLoading(false);
+    }
+  };
+
+  const updatePayment = async () => {
+    setLoading(true);
+
+    try {
```
charliecruzan-stripe, line 134: same here, don't need try catch

example/src/screens/PaymentSheetDeferredIntentMultiStepScreenWithUpdates.tsx
```diff
@@ -0,0 +1,311 @@
...
+      const { customer } = await fetchPaymentSheetParams();
```
charliecruzan-stripe, line 135: this might give us a new customer- should we save the old customer from the init step locally instead?

ios/StripeSdk+PaymentSheet.swift
```diff
@@ -194,6 +194,42 @@ extension StripeSdk {
...
+        if (params.object(forKey: "confirmHandler") == nil) {
+            resolve(Errors.createError(ErrorType.Failed, "You must provide `intentConfiguration.confirmHandler` if you are not passing an intent client secret"))
+            return
+        }
+        let captureMethodString = params["captureMethod"] as? String
+        let intentConfig = buildIntentConfiguration(
+            modeParams: modeParams,
+            paymentMethodTypes: params["paymentMethodTypes"] as? [String],
+            captureMethod: mapCaptureMethod(captureMethodString)
+        )
```
charliecruzan-stripe, line 214: this logic is also present in `preparePaymentSheetInstance`, maybe we could extract it out to avoid repeating it in two places?

ios/StripeSdk+PaymentSheet.swift
```diff
@@ -194,6 +194,42 @@ extension StripeSdk {
...
+        self.paymentSheetFlowController?.update(intentConfiguration: intentConfig){ [weak self] error in
+            if error != nil {
+
```
charliecruzan-stripe, line 217: add error handling

charliecruzan-stripe: most likely: `resolve(Errors.createError(ErrorType.Failed, error as NSError?))`

ios/StripeSdk.swift
```diff
@@ -127,6 +127,14 @@ class StripeSdk: RCTEventEmitter, STPBankSelectionViewControllerDelegate, UIAdap
         
         preparePaymentSheetInstance(params: params, configuration: configuration, resolve: resolve)
     }
+
+    @objc(updatePaymentSheet:resolver:rejecter:)
+    func updatePaymentSheet(params: NSDictionary, resolver resolve: @escaping RCTPromiseResolveBlock,
+                          rejecter reject: @escaping RCTPromiseRejectBlock) -> Void  {
+        DispatchQueue.main.async {
```
charliecruzan-stripe, line 134: why does the entire function need to be wrapped in `DispatchQueue.main.async`?

charliecruzan-stripe (changes requested):
**nice feature add!!!** thanks so much for your contribution, this looks excellent so far! Special thanks for doing such an incredible job following the conventions we've used so far in Payment Sheet functions

I have a couple comments, but overall this looks really great

alexdieudonne:
Any update on this pr pls ?

PR #1678, opened 2024-06-26
chore: update android dependency to 20.47.3

charliecruzan-stripe:
## Summary

updated stripe-android from 20.44.+ to 20.47.3


davidme-stripe (commented):
Approved

PR #1703, opened 2024-08-02
fix: make allowsRemovalOfLastSavedPaymentMethod default to true on android

charliecruzan-stripe:
## Summary

`getBoolean` defaults to false if the key isn't present, but we need it to default to true


davidme-stripe (commented):
Approved

PR #1767, opened 2024-10-30
Implement vertical mode

tjclawson-stripe:
## Summary

Add logic to parse `PaymentMethodLayout` configuration param for Android and iOS

## Motivation

Support for vertical mode in react native

## Testing

- [X] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/example/android/app/build.gradle
+++ b/example/android/app/build.gradle
@@ -283,6 +283,24 @@ dependencies {
         }
     }
     // END TODO
+
+    // TODO: Remove after upgrading react-native-screens
+    implementation ("androidx.lifecycle:lifecycle-livedata-core:2.8.7")
+      {
+      version {
+        strictly '2.7.0'
+      }
+    }
+    // END TODO
+
+    // TODO: Remove after upgrading react-native-screens
+    implementation ("androidx.fragment:fragment-ktx:1.8.5") {
+      version {
+        strictly '1.6.2'
+      }
+    }
+    // END TODO
+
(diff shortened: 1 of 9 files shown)


example/android/app/build.gradle
```diff
@@ -283,6 +283,24 @@ dependencies {
         }
     }
     // END TODO
+
```
tjclawson-stripe, line 286: These dependencies were conflicting with versions in `react-native-screens`. Should be able to remove after updating `react-native-screens` version

jaynewstrom-stripe: This really feels like something we shouldn't do. Is there another way around this?

jaynewstrom-stripe: Can we do the `react-native-screens` update before this PR?

tjclawson-stripe: Yeah I'm looking into this now. Based on [this](https://github.com/software-mansion/react-native-screens) we might have to upgrade the react native version to upgrade the react-native-screens dep to a version that doesn't cause conflicts. Alternatively, I could just update stripe-android to 20.52.0 and then we can handle the dependency conflicts for the major version update

tjclawson-stripe: Downgraded to 20.52.0

jaynewstrom-stripe (dismissed):

charliecruzan-stripe (approved):

Issue #1748, opened 2024-10-09
App crash when /create-intent request fails during apple pay


bartoszboruta:
**Describe the bug**
When you pay with applepay and any kind of error is returned from `/create-intent` the app is being crashed. 
Docs with the flow we follow: https://docs.stripe.com/payments/accept-a-payment-deferred?platform=react-native 

**To Reproduce**
return error from the backend `/create-intent` request during applepay from @stripe/stripe-react-native presentPaymentSheet method

**Expected behavior**
Nn error is displayed instead of the app crashing

**Screenshots**

✅  /create-intent returns client_secret

https://github.com/user-attachments/assets/62b423dd-53d6-4b8d-a11e-0a1027e92954

🔴  /create-intent throws an error - card is working well, apple pay - app crashes

https://github.com/user-attachments/assets/1ab84c80-8950-43d3-bcf2-5ea41d08264b

**Demo repository with app & server code**
just provide your sandbox keys in server, run the server and build the app locally
https://github.com/bartoszboruta/stripern

**Smartphone (please complete the following information):**
 - Device: iPhone 15 - any kind of iphones
 - OS: 17

**Additional context**
```
Thread 9 Crashed:
0   libsystem_kernel.dylib        	       0x1113353b0 0x11132c000 + 37808
1   libsystem_c.dylib             	       0x10fac75c0 0x10fa54000 + 472512
2   libswiftCore.dylib            	       0x10f12f0cc 0x10ee5c000 + 2961612
3   libswiftCore.dylib            	       0x10f12f0e8 0x10ee5c000 + 2961640
4   libswiftCore.dylib            	       0x10f179c50 0x10ee5c000 + 3267664
5   CoreFoundation                	       0x10d5b9288 0x10d488000 + 1249928
6   CoreFoundation                	       0x10d5bb5ac 0x10d488000 + 1258924
7   PassKitCore                   	       0x1328586dc 0x1323cc000 + 4769500
8   PassKitCore                   	       0x132858344 0x1323cc000 + 4768580
9   stripern                      	       0x10549c56c @nonobjc PKPaymentAuthorizationResult.init(status:errors:) + 152
10  stripern                      	       0x105494bf0 PKPaymentAuthorizationResult.__allocating_init(status:errors:) + 52
11  stripern                      	       0x105494894 closure #1 in STPApplePayContext.paymentAuthorizationController(_:didAuthorizePayment:handler:) + 536 (STPApplePayContext.swift:380)
[...]

PR #1818, opened 2025-01-24
use sendable struct instead of class

charliecruzan-stripe:
## Summary

The problem here is that a call within the stripe-ios library to initialize a PKPaymentAuthorizationResult is passing the error that we pass in from RN, which conforms to this [class](https://stripe.sourcegraphcloud.com/stripe/stripe-react-native/-/blob/ios/StripeSdk.swift?L141-149), which is not "sendable". but PKPaymentAuthorizationResult expects that it is sendable. classes _can_ be made sendable, but we should just use a struct here

## Motivation
fixes https://github.com/stripe/stripe-react-native/issues/1748
## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/ios/StripeSdk.swift
+++ b/ios/StripeSdk.swift
@@ -129,7 +129,7 @@ class StripeSdk: RCTEventEmitter, UIAdaptivePresentationControllerDelegate {
     }
 
     @objc(intentCreationCallback:resolver:rejecter:)
-    func intentCreationCallback(result: NSDictionary, resolver resolve: @escaping RCTPromiseResolveBlock,
+    @MainActor func intentCreationCallback(result: NSDictionary, resolver resolve: @escaping RCTPromiseResolveBlock,
                           rejecter reject: @escaping RCTPromiseRejectBlock) -> Void  {
         guard let paymentSheetIntentCreationCallback = self.paymentSheetIntentCreationCallback else {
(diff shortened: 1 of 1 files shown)


wooj-stripe (approved):

ios/StripeSdk.swift
```diff
@@ -138,18 +138,18 @@ class StripeSdk: RCTEventEmitter, UIAdaptivePresentationControllerDelegate {
...
-                    self.errorMessage = errorMessage
-                }
-                public var errorDescription: String? {
-                    return errorMessage
-                }
+          struct ConfirmationError: Error, LocalizedError {
+            private var errorMessage: String
+            init(errorMessage: String) {
+              self.errorMessage = errorMessage
+            }
```
wooj-stripe, line 145: nit: Technically, i think this is not needed since it is autosynthesized

wooj-stripe (approved):

ios/StripeSdk.swift
charliecruzan-stripe, line 145:
```suggestion
```

wooj-stripe: autosynthesized version turns to private. so we are reverting this change.

wooj-stripe (approved):

PR #1828, opened 2025-02-07
Support events in Financial Connections

tillh-stripe:
## Summary

This pull request adds support for passing an `onEvent` handler to Financial Connections methods, similar to the native Android and iOS SDKs.

## Motivation

[MOBILE_APIREVIEW-117](https://jira.corp.stripe.com/browse/MOBILE_APIREVIEW-117)

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/CollectBankAccountLauncherFragment.kt
+++ b/android/src/main/java/com/reactnativestripesdk/CollectBankAccountLauncherFragment.kt
@@ -7,6 +7,7 @@ import android.view.ViewGroup
 import android.widget.FrameLayout
 import androidx.appcompat.app.AppCompatActivity
 import androidx.fragment.app.Fragment
+import com.facebook.react.bridge.Arguments
 import com.facebook.react.bridge.Promise
 import com.facebook.react.bridge.ReactApplicationContext
 import com.reactnativestripesdk.utils.*
@@ -14,6 +15,7 @@ import com.reactnativestripesdk.utils.createError
 import com.reactnativestripesdk.utils.createResult
 import com.reactnativestripesdk.utils.mapFromPaymentIntentResult
 import com.reactnativestripesdk.utils.mapFromSetupIntentResult
+import com.stripe.android.financialconnections.FinancialConnections
 import com.stripe.android.model.PaymentIntent
 import com.stripe.android.model.SetupIntent
 import com.stripe.android.model.StripeIntent
@@ -31,6 +33,18 @@ class CollectBankAccountLauncherFragment(
   private val promise: Promise
 ) : Fragment() {
   private lateinit var collectBankAccountLauncher: CollectBankAccountLauncher
+
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+
+    val stripeSdkModule: StripeSdkModule? = context.getNativeModule(StripeSdkModule::class.java)
+    if (stripeSdkModule != null && stripeSdkModule.eventListenerCount > 0) {
+      FinancialConnections.setEventListener { event ->
+        val params = mapFromFinancialConnectionsEvent(event)
+        stripeSdkModule.sendEvent(context, "onFinancialConnectionsEvent", params)
+      }
+    }
+  }
 
   override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?,
                             savedInstanceState: Bundle?): View {
@@ -61,6 +75,13 @@ class CollectBankAccountLauncherFragment(
         collectParams
       )
     }
+  }
+
+  override fun onDestroy() {
+    super.onDestroy()
+
+    // Remove any event listener that might be set
+    FinancialConnections.clearEventListener()
   }
 
   private fun createBankAccountLauncher(): CollectBankAccountLauncher {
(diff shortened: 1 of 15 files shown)


android/src/main/java/com/reactnativestripesdk/CollectBankAccountLauncherFragment.kt
```diff
@@ -32,6 +34,18 @@ class CollectBankAccountLauncherFragment(
...
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+
+    val stripeSdkModule: StripeSdkModule? = context.getNativeModule(StripeSdkModule::class.java)
+    if (stripeSdkModule != null && stripeSdkModule.eventListenerCount > 0) {
+      FinancialConnections.setEventListener { event ->
+        val params = mapFromFinancialConnectionsEvent(event)
+        stripeSdkModule.sendEvent(context, "onFinancialConnectionsEvent", params)
+      }
+    }
```
tillh-stripe, line 46: This is a “best guess” approach, since I don’t see a way for us to reliably track if a listener is registered for `onFinancialConnectionsEvent`.

src/NativeStripeSdk.tsx
```diff
@@ -84,7 +84,7 @@ type NativeStripeSdkType = {
   collectBankAccount(
     isPaymentIntent: boolean,
     clientSecret: string,
-    params: PaymentMethod.CollectBankAccountParams
+    params: Omit<PaymentMethod.CollectBankAccountParams, 'onEvent'>
```
tillh-stripe, line 87: No need to pass `onEvent` around.

android/src/main/java/com/reactnativestripesdk/CollectBankAccountLauncherFragment.kt
```diff
@@ -32,6 +34,18 @@ class CollectBankAccountLauncherFragment(
 ) : Fragment() {
   private lateinit var collectBankAccountLauncher: CollectBankAccountLauncher
 
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+
+    val stripeSdkModule: StripeSdkModule? = context.getNativeModule(StripeSdkModule::class.java)
+    if (stripeSdkModule != null && stripeSdkModule.eventListenerCount > 0) {
```
charliecruzan-stripe, line 41: let's do something like this instead: https://github.com/stripe/stripe-react-native/blob/447804f5529fe91c2f6ffbf983d3151ece283ba4/android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt#L143-L148

android/src/main/java/com/reactnativestripesdk/FinancialConnectionsSheetFragment.kt
```diff
@@ -27,6 +27,18 @@ class FinancialConnectionsSheetFragment : Fragment() {
   private lateinit var configuration: FinancialConnectionsSheet.Configuration
   private lateinit var mode: Mode
 
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+
+    val stripeSdkModule: StripeSdkModule? = context.getNativeModule(StripeSdkModule::class.java)
+    if (stripeSdkModule != null && stripeSdkModule.eventListenerCount > 0) {
```
charliecruzan-stripe, line 34: same here, do something like this: https://github.com/stripe/stripe-react-native/blob/447804f5529fe91c2f6ffbf983d3151ece283ba4/android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt#L143-L148

src/functions.ts
```diff
@@ -468,13 +468,19 @@ export const collectBankAccountForPayment = async (
   clientSecret: string,
   params: PaymentMethod.CollectBankAccountParams
 ): Promise<CollectBankAccountForPaymentResult> => {
+  const subscription =
+    params.onEvent &&
+    eventEmitter.addListener('onFinancialConnectionsEvent', params.onEvent);
```
charliecruzan-stripe, line 473: we have an existing paradigm for this in this file: https://github.com/stripe/stripe-react-native/blob/f1400cc6ab0a9f1d266cfdca000d02e1d4d4659c/src/functions.ts#L357

I think it _should_ be fine with this setup overall, but i'd rather keep it consistent with the other subscriptions we have

charliecruzan-stripe (changes requested):

android/src/main/java/com/reactnativestripesdk/CollectBankAccountLauncherFragment.kt
tillh-stripe: The event listener is optional, so it wouldn’t be right to throw an error in this case.

My thinking here was: If we have an event listener in React, then this _might_ be a Financial Connections event listener, so let’s try to send the even across the bridge.

src/functions.ts
tillh-stripe: Gotcha, will take a stab at it!

charliecruzan-stripe (dismissed):
Looks good! just one comment on the changelog but otherwise 🚢

charliecruzan-stripe (approved):

PR #1849, opened 2025-03-07
Upgrade Stripe Android SDK to 21.6.0

samer-stripe:
## Summary
Upgrade Stripe Android SDK to 21.6.0

## Motivation
Allows access to new Android APIs developed since `20.52.+`

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/utils/Mappers.kt
+++ b/android/src/main/java/com/reactnativestripesdk/utils/Mappers.kt
@@ -427,7 +427,7 @@ internal fun mapFromPaymentMethod(paymentMethod: PaymentMethod): WritableMap {
       )
       it.putString("last4", paymentMethod.usBankAccount?.last4)
       it.putString("bankName", paymentMethod.usBankAccount?.bankName)
-      it.putString("linkedAccount", paymentMethod.usBankAccount?.linkedAccount)
+      it.putString("linkedAccount", paymentMethod.usBankAccount?.financialConnectionsAccount)
       it.putString("fingerprint", paymentMethod.usBankAccount?.fingerprint)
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/reactnativestripesdk/utils/Mappers.kt
```diff
@@ -427,7 +427,7 @@ internal fun mapFromPaymentMethod(paymentMethod: PaymentMethod): WritableMap {
       )
       it.putString("last4", paymentMethod.usBankAccount?.last4)
       it.putString("bankName", paymentMethod.usBankAccount?.bankName)
-      it.putString("linkedAccount", paymentMethod.usBankAccount?.linkedAccount)
+      it.putString("linkedAccount", paymentMethod.usBankAccount?.financialConnectionsAccount)
```
samer-stripe, line 430: This maps to a type that still uses `linkedAccount` instead of `financialConnectionsAccount`. I think this is fine since they are intended to be the same value but should we look at making a future API update to rename or at least deprecate the `linkedAccount` attribute name in favor of `financialConnectionsAccount`? @tillh-stripe

tillh-stripe: Yes, we should eventually rename the field to `financialConnectionsAccount`, but I feel ok leaving it as `linkedAccount` for now.

janicduplessis (dismissed):
Great that this now works!

janicduplessis (dismissed):

porter-stripe (approved):

jaynewstrom-stripe (approved):

PR #1854, opened 2025-03-10
Fix E2E tests

samer-stripe:
## Summary
Fix E2E tests

## Motivation
Ensures E2E tests pass successfully!

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 


example/package.json
```diff
@@ -5,8 +5,8 @@
   "private": true,
   "scripts": {
     "android": "react-native run-android",
-    "build:android": "npm run mkdist && react-native bundle --entry-file index.js --platform android --dev true --bundle-output dist/main.android.jsbundle --assets-dest dist/res",
-    "build:ios": "npm run mkdist && react-native bundle --entry-file index.js --platform ios --dev true --bundle-output dist/main.ios.jsbundle --assets-dest dist",
+    "build:android": "npm run mkdist && react-native bundle --entry-file index.js --platform android --dev false --bundle-output dist/main.android.jsbundle --assets-dest dist/res",
+    "build:ios": "npm run mkdist && react-native bundle --entry-file index.js --platform ios --dev false --bundle-output dist/main.ios.jsbundle --assets-dest dist",
```
samer-stripe, line 9: Generating the bundles as dev resulted in:
- iOS being unable to load the bundle in release mode, resulting in a blank screen.
- Android loading the bundle but acting in dev mode, expecting a local server to have been launched rather than rely on the Glitch server.

example/app.json
```diff
@@ -32,6 +32,20 @@
...
+        "storePassword": "android"
+      },
+      "release": {
+        "keyAlias": "androiddebugkey",
+        "keyPassword": "android",
+        "storeFile": "debug.keystore",
+        "storePassword": "android"
+      }
+    }
   }
```
samer-stripe, line 50: Added the `signingConfigs` and re-added the deleted `debug.keystore` to allow for generating a release Android build.

.github/workflows/e2e-tests.yml
```diff
@@ -140,7 +140,13 @@ jobs:
...
 
       - name: Install Dependencies
-        run: yarn bootstrap
+        run: yarn bootstrap-no-pods
+
+      - name: Generate bundle
+        run: yarn example build:ios
+
+      - name: Install Pods
+        run: yarn pods
```
samer-stripe, line 149: Since `react-native-test-app` doesn't generate the bundle by default, we need to do on our own and it has to happen before we install/generate the pods. If we don't, the bundle and assets we defined in `app.json` are not taken into account when creating the Pod files since they don't exist.

.github/workflows/e2e-tests.yml
```diff
@@ -83,6 +83,7 @@ jobs:
         uses: reactivecircus/android-emulator-runner@v2
         with:
           api-level: 33
+          emulator-build: 11237101
```
samer-stripe, line 86: Specifying specific build number to get a stable emulator. There are issues with some of the other emulator builds that result in a `Failed to find: ColorBuffer` error.

tillh-stripe: Should we add a note about this inline?

tillh-stripe (commented):
Thanks for getting these fixed again 🙏

tillh-stripe (approved):

PR #1865, opened 2025-03-14
Update react-native deep imports for 0.79 compatibility

huntie:
## Summary

In the incoming React Native 0.79 release, we've converted a number of internal modules to `export` syntax — affecting usages with `require()`. This PR attempts to update these imports ahead of time, to work with both versions.

💡 Secondary note: In a future release, we are intending to omit each of these deep-imported modules from the public API of React Native. There will be a consultation period on opening up new APIs via the root `'react-native'` exports — e.g. `TextInputState` here.

## Test plan

**Needs testing**. CI should pass for integration with current RN versions.

- [ ] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 


janicduplessis:
@huntie Thanks for the PR! Would switching to using import works here? Not sure why it was done this way in the first place, but we're using import syntax through the codebase already.

huntie:
@janicduplessis Yes, `import` will absolutely work too in this case 👍🏻

src/helpers.ts
```diff
@@ -2,7 +2,8 @@ import type React from 'react';
 import type { StripeError } from './types';
 
 import { Platform, NativeModules } from 'react-native';
-const TextInputState = require('react-native/Libraries/Components/TextInput/TextInputState');
+// @ts-ignore TextInputState has no type definition
```
huntie, line 5: FYI: We're actively working on improved TypeScript defs across `react-native`. A future RN version should export a typed version of this export from the root.

janicduplessis (approved):

samer-stripe (approved):

src/helpers.ts
huntie: Update on this:

- Went ahead with https://github.com/facebook/react-native/pull/50321 initially.
**However**, after some team discussion we think this is in fact an internal API of `TextInput`.
    - If used directly, `TextInputState` has shortcomings — e.g. doesn't consider a multi-window scenario.
    - **Suggested**: The public API is focusing and blurring from a text input ref.

janicduplessis: I will investigate our usage to see if it would be possible to use public api

PR #1877, opened 2025-04-09
Fix broken parcelize on Android

tillh-stripe:
## Testing

- [ ] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -22,6 +22,7 @@ if (project == rootProject) {
 
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
+apply plugin: 'kotlin-parcelize'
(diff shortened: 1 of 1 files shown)


davidme-stripe (approved):

android/build.gradle
```diff
@@ -22,6 +22,7 @@ if (project == rootProject) {
 
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
+apply plugin: 'kotlin-parcelize'
```
jaynewstrom-stripe, line 25: We've gotten reports about similar issues from merchants. We should root cause this. Do we need to `api` a dependency in the repo or something?

PR #1889, opened 2025-04-14
Re-enable FC tests on CI

tillh-stripe:
## Summary

This pull request re-enables the Financial Connections tests on CI.

We now use a different merchant to guarantee the native flow, and made some other changes to make the test reliable.

## Testing

- [ ] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 


e2e-tests/ios-only/financial-connections-session.yml
```diff
@@ -9,16 +9,22 @@ appId: ${APP_ID}
     text: "Collect session"
     retryTapIfNoChange: false
 - assertVisible:
-    text: "Manually verify instead"
-- tapOn: "Manually verify instead"
```
tillh-stripe, line 13: Seems like Maestro is unable to hit this on iOS. We switch to longer flow of actually selecting an institution and account.

example/.env.example
```diff
@@ -14,6 +14,9 @@ STRIPE_SECRET_KEY_AU=sk_test
 STRIPE_PUBLISHABLE_KEY_WECHAT=pk_test
 STRIPE_SECRET_KEY_WECHAT=sk_test
 
+STRIPE_PUBLISHABLE_KEY_FINANCIAL_CONNECTIONS=pk_test
+STRIPE_SECRET_KEY_FINANCIAL_CONNECTIONS=sk_test
```
tillh-stripe, line 18: We're using a separate merchant for Financial Connections, so that we can guarantee a native flow (and not a webview).

PR #1893, opened 2025-04-17
Implement new architecture

janicduplessis:
## Testing

- [ ] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 


.clang-format
```diff
@@ -0,0 +1,90 @@
+---
```
janicduplessis, line 1: I added this config based on the react native one to help format objc++ code. I am thinking we could do a better pass at code formatting on iOS later.

android/src/androidTest/java/com/reactnativestripesdk/addresssheet/AddressSheetViewTest.kt
```diff
@@ -178,7 +178,7 @@ class AddressSheetViewTest {
         ),
       )
 
-    val expected =
+    val result =
```
janicduplessis, line 181: I had to change the shape of a few events so the payload is an object to be able to use UnsafeMixed codegen. This doesn't change the public api.

android/src/main/java/com/reactnativestripesdk/AuBECSDebitFormViewManager.kt
```diff
@@ -1,35 +1,44 @@
...
+import com.facebook.react.viewmanagers.AuBECSDebitFormManagerDelegate
+import com.facebook.react.viewmanagers.AuBECSDebitFormManagerInterface
+import com.reactnativestripesdk.utils.asMapOrNull
 
 @ReactModule(name = AuBECSDebitFormViewManager.REACT_CLASS)
-class AuBECSDebitFormViewManager : SimpleViewManager<AuBECSDebitFormView>() {
+class AuBECSDebitFormViewManager :
+  SimpleViewManager<AuBECSDebitFormView>(),
+  AuBECSDebitFormManagerInterface<AuBECSDebitFormView> {
+  private val delegate = AuBECSDebitFormManagerDelegate(this)
```
janicduplessis, line 16: For android this is a simple example of the changes needed:

- Implement the codegen interface
- Add the delegate
- UnsafeMixed generates Dynamic type instead of ReadableMap, so I made a extension method to easily convert between the two.

android/src/main/java/com/reactnativestripesdk/CardChangeEvent.kt
```diff
@@ -0,0 +1,56 @@
+package com.reactnativestripesdk
```
janicduplessis, line 1: This is a rename, not sure why git didn't pick it up. CardChangedEvent -> CardChangeEvent, for consistency.

android/src/main/java/com/reactnativestripesdk/CardFieldViewManager.kt
```diff
@@ -1,127 +1,137 @@
...
-    val stripeSdkModule: StripeSdkModule? =
-      reactContext.getNativeModule(StripeSdkModule::class.java)
-    val view = CardFieldView(reactContext)
-
-    reactContextRef = reactContext
+  // Native commands
 
-    stripeSdkModule?.cardFieldView = view
-    return view
+  override fun blur(view: CardFieldView) {
```
janicduplessis, line 125: View commands are now just overridden methods.

android/src/main/java/com/reactnativestripesdk/CollectBankAccountLauncherFragment.kt
```diff
@@ -38,10 +38,10 @@ class CollectBankAccountLauncherFragment(
     super.onCreate(savedInstanceState)
 
     val stripeSdkModule: StripeSdkModule? = context.getNativeModule(StripeSdkModule::class.java)
-    if (stripeSdkModule != null && stripeSdkModule.eventListenerCount > 0) {
+    if (stripeSdkModule != null) {
       FinancialConnections.setEventListener { event ->
         val params = mapFromFinancialConnectionsEvent(event)
-        stripeSdkModule.sendEvent(context, "onFinancialConnectionsEvent", params)
+        stripeSdkModule.emitOnFinancialConnectionsEvent(params)
```
janicduplessis, line 44: Events are now emitted from codegened methods on the native module.

android/src/main/java/com/reactnativestripesdk/CollectBankAccountLauncherFragment.kt
```diff
@@ -38,10 +38,10 @@ class CollectBankAccountLauncherFragment(
     super.onCreate(savedInstanceState)
 
     val stripeSdkModule: StripeSdkModule? = context.getNativeModule(StripeSdkModule::class.java)
-    if (stripeSdkModule != null && stripeSdkModule.eventListenerCount > 0) {
```
janicduplessis, line 41: I dropped tracking the listener count, as there is no api for it with the new codegened event emitters and doesn't seem to be needed.

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
```diff
@@ -176,17 +171,18 @@ class StripeSdkModule(
     )
   }
 
-  override fun getConstants(): MutableMap<String, Any> =
-    hashMapOf(
+  @SuppressLint("RestrictedApi")
+  override fun getTypedExportedConstants() =
+    mapOf(
       "API_VERSIONS" to
-        hashMapOf(
```
janicduplessis, line 182: Deprecated api

example/src/screens/CustomerSheetScreen.tsx
```diff
@@ -68,7 +68,7 @@ export default function CustomerSheetScreen() {
       setupIntentClientSecret,
       customerEphemeralKeySecret,
       customerId,
-      returnURL: 'stripe-example://stripe-redirect',
+      returnURL: 'com.stripe.react.native://stripe-redirect',
```
janicduplessis, line 71: I missed some when updating the example app, the new linking scheme should be com.stripe.react.native and not stripe-example.

example/src/screens/MultilineWebhookPaymentScreen.tsx
```diff
@@ -113,6 +115,22 @@ export default function MultilineWebhookPaymentScreen() {
           countryCode: 'US',
         }}
       />
+      <View style={[styles.row, styles.rowWithGap]}>
```
janicduplessis, line 118: Added this to test view commands.

ios/AddressSheet/AddressSheetView.swift
```diff
@@ -9,33 +9,33 @@ import Foundation
...
-    @objc var appearance: NSDictionary? = nil
-    @objc var defaultValues: NSDictionary? = nil
-    @objc var additionalFields: NSDictionary? = nil
-    @objc var allowedCountries: [String] = []
-    @objc var autocompleteCountries: [String] = []
-    @objc var primaryButtonTitle: String? = nil
-    @objc var sheetTitle: String? = nil
-    @objc var onSubmitAction: RCTDirectEventBlock?
-    @objc var onErrorAction: RCTDirectEventBlock?
+public class AddressSheetView: UIView {
```
janicduplessis, line 12: Methods / classes need to be public to be usable from objc.

ios/ApplePayButtonManager.m
```diff
@@ -3,7 +3,6 @@
 #import <React/RCTViewManager.h>
 
 @interface RCT_EXTERN_MODULE(ApplePayButtonManager, RCTViewManager)
-RCT_EXPORT_VIEW_PROPERTY(onPressAction, RCTDirectEventBlock)
```
janicduplessis, line 6: This prop wasn't used.

ios/ApplePayButtonManager.swift
```diff
@@ -3,9 +3,7 @@ import Foundation
 @objc(ApplePayButtonManager)
 class ApplePayButtonManager: RCTViewManager {
     override func view() -> UIView! {
-        let view = ApplePayButtonView(frame: CGRect.init())
-        view.stripeSdk = bridge.module(forName: "StripeSdk") as? StripeSdk
-        return view
+        return ApplePayButtonView(frame: CGRect.init())
```
janicduplessis, line 6: Since the stripe sdk implementation is now a singleton it doesn't need to be passed around from the native module.

ios/CardFieldManager.swift
```diff
@@ -3,10 +3,7 @@ import Foundation
 @objc(CardFieldManager)
 class CardFieldManager: RCTViewManager {
     override func view() -> UIView! {
-        let cardField = CardFieldView()
-        let stripeSdk = bridge.module(forName: "StripeSdk") as? StripeSdk
-        stripeSdk?.cardFieldView = cardField;
```
janicduplessis, line 8: This is now set by the view class

janicduplessis:
The iOS e2e tests are failing for grabpay, but it also seems to fail on a fresh pr with no changes so I don’t think it’s related. From what I saw from the test failure screenshot the webview fails to load the stripe website.

jwaldrip:
Does this enable support for bridgeless mode?

janicduplessis:
Yes, I tested with bridgeless mode enabled.

android/src/main/java/com/reactnativestripesdk/AuBECSDebitFormViewManager.kt
```diff
@@ -1,35 +1,44 @@
...
   ) {
     view.setCompanyName(name)
   }
 
   @ReactProp(name = "formStyle")
-  fun setFormStyle(
+  override fun setFormStyle(
     view: AuBECSDebitFormView,
-    style: ReadableMap,
+    style: Dynamic,
```
samer-stripe, line 39: Is `Dynamic` here a compat type for the legacy `ReadableMap` and the new architecture?

android/src/main/java/com/reactnativestripesdk/CardFieldView.kt
```diff
@@ -99,10 +99,10 @@ class CardFieldView(
   private fun onChangeFocus() {
     UIManagerHelper
       .getEventDispatcherForReactTag(context, id)
-      ?.dispatchEvent(CardFocusEvent(context.surfaceId, id, currentFocusedField))
+      ?.dispatchEvent(CardFocusChangeEvent(context.surfaceId, id, currentFocusedField))
   }
 
-  fun setCardStyle(value: ReadableMap) {
+  fun setCardStyle(value: ReadableMap?) {
```
samer-stripe, line 105: Any risks with these being nullable?

android/src/main/java/com/reactnativestripesdk/CardFieldViewManager.kt
```diff
@@ -1,127 +1,137 @@
...
+    return view
+  }
+
+  override fun onDropViewInstance(view: CardFieldView) {
+    super.onDropViewInstance(view)
+
+    val stripeSdkModule: StripeSdkModule? =
+      reactContextRef?.getNativeModule(StripeSdkModule::class.java)
+    stripeSdkModule?.cardFieldView = null
+    reactContextRef = null
```
samer-stripe, line 47: Not related to this PR but I really don't understand why we are setting the field view in the native module. I'm hoping there is some way we can avoid this.

android/src/main/java/com/reactnativestripesdk/CardFieldViewManager.kt
samer-stripe: This is a much better solution!

android/src/main/java/com/reactnativestripesdk/CollectBankAccountLauncherFragment.kt
samer-stripe: Is `EventEmitter` abstracted away now behind the codegen components? Is there React Native documentation I view on this?

android/src/oldarch/java/com/facebook/react/viewmanagers/AddressSheetViewManagerDelegate.java
```diff
@@ -0,0 +1,64 @@
...
+      case "sheetTitle":
+        mViewManager.setSheetTitle(view, value == null ? null : (String) value);
+        break;
+      case "googlePlacesApiKey":
+        mViewManager.setGooglePlacesApiKey(view, value == null ? null : (String) value);
+        break;
+      default:
+        super.setProperty(view, propName, value);
+    }
+  }
```
samer-stripe, line 63: For my understanding, this delegate is to maintain compatibility with the legacy architecturre?

android/src/main/java/com/reactnativestripesdk/CardFieldViewManager.kt
janicduplessis: Looks like it is being used to get the info from the field later when creating a payment method (https://github.com/stripe/stripe-react-native/blob/f5e7cfc20ed08d1cb855781b6cb2637c8736484e/android/src/main/java/com/reactnativestripesdk/PaymentMethodCreateParamsFactory.kt). I feel like there should be a better way to do this.

android/src/oldarch/java/com/facebook/react/viewmanagers/AddressSheetViewManagerDelegate.java
janicduplessis: Yes, old architecture does not run codegen, so it is just a copy of what is generated by codegen.

I was thinking it might be good to automate copying those over in a follow up to make sure we don't forget when making changes to the spec files.

android/src/main/java/com/reactnativestripesdk/CardFieldView.kt
janicduplessis: This was mostly to please the compiler since after changing some methods to use Dynamic instead of ReadableMap. The conversion method I made always returns nullable ReadableMap from Dynamic. I could make a version that asserts that the value is not null and return ReadableMap if the prop can never be null, but supporting passing null was usually pretty simple and seemed safer.

android/src/main/java/com/reactnativestripesdk/CardFieldViewManager.kt
samer-stripe: Agreed! We should file a ticket for someone to fix this later.

src/components/CardField.tsx
```diff
@@ -124,11 +124,11 @@ export const CardField = forwardRef<CardFieldInput.Methods, Props>(
     );
 
     const onFocusHandler = useCallback(
-      (event: CardFieldInput.OnFocusChangeEvent) => {
+      (event: NativeSyntheticEvent<FocusChangeEvent>) => {
         const { focusedField } = event.nativeEvent;
         if (focusedField) {
           focusInput(inputRef.current);
-          onFocus?.(focusedField);
+          onFocus?.(focusedField as CardFieldInput.FieldName);
```
samer-stripe, line 131: Why do we need to TS cast here? Is `event.nativeEvent` not typed properly?

android/src/main/java/com/reactnativestripesdk/CollectBankAccountLauncherFragment.kt
janicduplessis: Yes, we still need some code to support the old arch, but once we don't need to support it events are 100% codegened.

Here's the doc: https://reactnative.dev/docs/next/the-new-architecture/native-modules-custom-events

src/components/CardField.tsx
janicduplessis: I had to type this as string instead of an enum, otherwise iOS codegen will type it as a c++ enum, which requires extra conversions. Since for now we still need to support old arch it is simpler to keep it typed like this.

src/types/components/CardFieldInput.ts
```diff
@@ -3,12 +3,7 @@ import type { CardBrand } from '../Token';
 
 export type FieldName = 'CardNumber' | 'Cvc' | 'ExpiryDate' | 'PostalCode';
 
-export enum ValidationState {
-  Valid = 'Valid',
-  Invalid = 'Invalid',
-  Incomplete = 'Incomplete',
-  Unknown = 'Unknown',
-}
+export type ValidationState = 'Valid' | 'Invalid' | 'Incomplete' | 'Unknown';
```
samer-stripe, line 6: I agree that this is the way we should be doing it but I don't think we can make this change without an API review.

src/types/components/CardFieldInput.ts
```diff
@@ -55,14 +50,13 @@ export type OnFocusChangeEvent = NativeSyntheticEvent<{
   focusedField: FieldName | null;
 }>;
 
-export type OnCardChangeEvent = NativeSyntheticEvent<Details>;
+export type OnCardChangeEvent = NativeSyntheticEvent<{ card: Details }>;
```
samer-stripe, line 53: This is an internal event?

src/types/CustomerSheet.ts
```diff
@@ -56,6 +56,10 @@ export type CustomerSheetInitParams = {
   cardBrandAcceptance?: PaymentSheet.CardBrandAcceptance;
 };
 
+export type CustomerSheetInitResult = {
+  error?: StripeError<CustomerSheetError>;
+};
+
```
samer-stripe, line 62: I also think we cannot introduce this type without an API review. I agree though it should be a type instead of the function returning `Promise<{error?: StripeError<CustomerSheetError>}`.

samer-stripe:
This looks mostly good to me! There are a couple API changes we cannot make in this PR without a proper API review. I'm hoping we can progress without needing to make API changes.

src/types/components/CardFieldInput.ts
janicduplessis: Good catch, I think I can revert these changes

src/types/components/CardFieldInput.ts
janicduplessis: It is only used internally, this is the type for the NativeCardField onCardChange prop, but the CardField component uses a different type.

However the type is exported as CardFieldInput.OnCardChangeEvent, so technically it is public api, although I doubt it would be used since the component its for is not public.

janicduplessis:
@samer-stripe I reverted the type changes that were not needed. Do you think it is ok now?

samer-stripe (approved):
LGTM on the Android side! Did a review on iOS as well but would be more confident if someone can have a look there as well.

porter-stripe (commented):
iOS side looks good! 

Could we get this up to date w/ master, I just merged a PR that may need some small changes https://github.com/stripe/stripe-react-native/pull/1899 🙏

janicduplessis:
I will wait for https://github.com/stripe/stripe-react-native/pull/1901 to land and then port the new component to new arch too.

porter-stripe:
@janicduplessis thank you! After the JS PR lands I'll hold off on more changes until this lands.

janicduplessis:
@porter-stripe Sounds good! If you get to add the example in the example app that would help too when testing.

samer-stripe (approved):

porter-stripe:
@janicduplessis thanks for letting us know! Please tag me in the follow up PR w/ embedded

samer-stripe (approved):

janicduplessis:
@samer-stripe Do we still need review for iOS part?

porter-stripe (approved):

PR #1894, opened 2025-04-17
Embedded API

porter-stripe:
## Testing

- [ ] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 


src/types/EmbeddedPaymentElement.tsx
```diff
@@ -0,0 +1,427 @@
...
+  | { status: 'completed' }
+  | { status: 'canceled' }
+  | { status: 'failed'; error: Error };
+
+/**
+ * The result of an update call.
+ */
+export type EmbeddedPaymentElementUpdateResult =
+  | { status: 'succeeded' }
+  | { status: 'canceled' }
```
jaynewstrom-stripe, line 51: Android doesn't have canceled as this would only happen if the merchant cancelled via a kotlin coroutine cancelation.

src/types/EmbeddedPaymentElement.tsx
```diff
@@ -0,0 +1,427 @@
...
+   * Stripe values: https://stripe.com/docs/api/payment_methods/object#payment_method_object-type
+   * External methods: https://stripe.com/docs/payments/external-payment-methods?platform=ios#available-external-payment-methods
+   * Apple Pay: "apple_pay"
+   */
+  paymentMethodType: string;
+  /**
+   * Optional base64-encoded regulatory mandate text.
+   * If `embeddedViewDisplaysMandateText = false`, this must be shown near the buy button in a tappable view that supports links.
+   */
+  mandateText?: string;
```
jaynewstrom-stripe, line 81: Android mostly can't expose this as a string, because we have to be able to support links and such.

porter-stripe: What is it exposed as?

src/types/EmbeddedPaymentElement.tsx
```diff
@@ -0,0 +1,427 @@
...
+    };
+
+/**
+ * Describes how row selections are handled in the `EmbeddedPaymentElement` view.
+ *
+ * When customers select a payment option (like a card), your app can either:
+ * - Wait for the customer to manually confirm by tapping a button (`default` behavior), or
+ * - Immediately perform an action as soon as the option is selected (`immediateAction` behavior).
+ */
+export type RowSelectionBehavior =
```
jaynewstrom-stripe, line 118: I _think_ this will need to be passed in `init` for Android, rather than configuration. API is still WIP, so won't know for sure for a little bit. Might be worth keeping this out of the initial implementation to unblock.

src/types/EmbeddedPaymentElement.tsx
porter-stripe: I suppose it's reasonable to never pass back canceled when used on Android then.

src/types/EmbeddedPaymentElement.tsx
```diff
@@ -0,0 +1,427 @@
...
+   * By default, EmbeddedPaymentElement will accept all supported cards by Stripe.
+   * You can specify card brands EmbeddedPaymentElement should block or allow payment for by providing an array of those card brands.
+   * Note: This is only a client-side solution.
+   * Note: Card brand filtering is not currently supported in Link.
+   */
+  cardBrandAcceptance?: PaymentSheetTypes.CardBrandAcceptance;
+  /** (Private Preview) This parameter is expected to be removed once we GA this feature
+   * When using customerSessions, allow users to update their saved cards
+   */
+  updatePaymentMethodEnabled?: boolean;
```
jaynewstrom-stripe, line 210: This should be removed. I _think_ it's also getting removed from iOS.

src/types/EmbeddedPaymentElement.tsx
```diff
@@ -0,0 +1,427 @@
...
+   */
+  formSheetAction?: EmbeddedFormSheetAction;
+  /** Controls whether the view displays mandate text at the bottom for payment methods that require it. If set to `false`, your integration must display `PaymentOptionDisplayData.mandateText` to the customer near your “Buy” button to comply with regulations.
+   * Note: This doesn't affect mandates displayed in the form sheet.
+   */
+  embeddedViewDisplaysMandateText?: boolean;
+  /**
+   * Determines the behavior when a row  is selected. Defaults to `.default`.
+   */
+  rowSelectionBehavior?: RowSelectionBehavior;
```
jaynewstrom-stripe, line 222: As mentioned above, I don't think this will work on Android, and will need to be passed into init. Probably worth not doing it as a first implementation.

src/types/EmbeddedPaymentElement.tsx
porter-stripe: What's the ETA on this?

src/types/EmbeddedPaymentElement.tsx
```diff
@@ -0,0 +1,427 @@
...
+// -----------------------------------------------------------------------------
+
+class EmbeddedPaymentElement {
+  /**
+   * Call this when the intent configuration changes (e.g., amount or currency).
+   * Cancels any in-progress update. Ensures the correct payment methods are shown and fields are collected.
+   * If the selected payment option becomes invalid, it may be cleared.
+   * Returns the final result of the update; earlier in-flight updates will return `{ status: 'canceled' }`.
+   */
+  async update(
```
jaynewstrom-stripe, line 236: Android calls this configure (not a big deal on the naming difference).

But we also pass in configuration here. https://github.com/stripe/stripe-android/blob/4633fa71bb77649bbbc61bd51a50f319796fb68c/paymentsheet/src/main/java/com/stripe/android/paymentelement/EmbeddedPaymentElement.kt#L78

src/types/EmbeddedPaymentElement.tsx
```diff
@@ -0,0 +1,427 @@
...
+ */
+export type EmbeddedFormSheetAction =
+  | {
+      /**
+       * The button says “Pay” or “Set up”. When tapped, it confirms the payment or setup directly within the form sheet.
+       * @param result - Callback invoked with the result of the confirmation. You can use this to show a success message or handle errors.
+       */
+      type: 'confirm';
+      onFormSheetConfirmComplete?: (
+        result: EmbeddedPaymentElementResult
```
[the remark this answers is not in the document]
porter-stripe: I think that's ok, in the native android code we should be able to look inside the overall result callback, and call this callback when the user is using formSheetAction.confirm.

src/types/EmbeddedPaymentElement.tsx
```diff
@@ -0,0 +1,427 @@
...
+  }
+
+  /**
+   * Confirm the payment or setup intent.
+   * Waits for any in-progress `update()` call to finish before proceeding.
+   * May present authentication flows (e.g., 3DS) if required.
+   * Requires the most recent `update()` call to have succeeded.
+   * Returns the final result: success, failure, or cancellation.
+   */
+  async confirm(): Promise<EmbeddedPaymentElementResult> {
```
jaynewstrom-stripe, line 251: Android doesn't return the result, but rather delivers it via a callback set at initialization time.

This is to support things like process death, as confirm might redirect to other applications.

src/types/EmbeddedPaymentElement.tsx
```diff
@@ -0,0 +1,427 @@
...
+
+// -----------------------------------------------------------------------------
+// JS Factory: createEmbeddedPaymentElement
+// -----------------------------------------------------------------------------
+let confirmHandlerCallback: EmitterSubscription | null = null;
+let formSheetActionConfirmCallback: EmitterSubscription | null = null;
+let rowSelectionImmediateActionCallback: EmitterSubscription | null = null;
+
+async function createEmbeddedPaymentElement(
+  intentConfig: PaymentSheetTypes.IntentConfiguration,
```
jaynewstrom-stripe, line 270: At least on Android we can't initialize with the intent config or the configuration. This is to support process death and activity recreation.

src/types/EmbeddedPaymentElement.tsx
porter-stripe: Given iOS doesn't allow you to pass in a config on update, I think we should take the subset here and not pass in the configuration

src/types/EmbeddedPaymentElement.tsx
porter-stripe: Should be something we can handle on the bridging layer I'd assume, e.g. just hold onto the intentConfig and configuration until we can init.

src/types/EmbeddedPaymentElement.tsx
jaynewstrom-stripe: Tian is on run this week, has a few things to do next week, then will start on the Android API review. So probably 2 ish weeks.

src/types/EmbeddedPaymentElement.tsx
jaynewstrom-stripe: I think we'll have to rework the implementation for iOS RN. We need to be able to initialize with no arguments in order to support Android.

So the iOS impl for update might be init + update where android would just be configure.

src/types/EmbeddedPaymentElement.tsx
porter-stripe: Yeah should be removed

src/types/EmbeddedPaymentElement.tsx
jaynewstrom-stripe: I'd rather omit for now, and add later if someone needs it and make a call on it then.

porter-stripe: Yeah I think to support this we'd actually need to vend back a view w/ the mandate :(

porter-stripe: I'm opting to removing mandate stuff from this API, I think in RN merchants will need to let embedded show the mandate.

example/src/screens/EmbeddedPaymentElementScreen.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+      setElementConfig(uiConfig);
+      setIntentConfig(newIntentConfig);
+    },
+    [customerKeyType]
+  );
+
+  // Hook into Stripe element
+  const { view, paymentOption, confirm, clear } = useEmbeddedPaymentElement(
+    intentConfig!,
```
janicduplessis, line 220: Can we allow passing null / undefined here? I think generally non-null assertion is seen as bad.

nickw:
@porter-stripe is the Embedded Element for React Native still planned? We were really hoping to use this feature in our app and were bummed to see this one closed!

porter-stripe:
@nickw this was split up into a few PRs! It is currently in public preview: https://docs.stripe.com/payments/mobile/accept-payment-embedded?platform=react-native&type=payment

PR #1899, opened 2025-04-23
Implement bridge for RN Android

porter-stripe:
## Summary
- Implements the bridge layer for embedded payment element on Android
- Based on this prototype https://github.com/stripe/stripe-react-native/pull/1894/files?file-filters%5B%5D=.kt&show-viewed-files=true
- Adds compose to our library

## Motivation
- Embedded payment element

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -12,6 +12,7 @@ buildscript {
     // noinspection DifferentKotlinGradleVersion
     classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     classpath 'com.diffplug.spotless:spotless-plugin-gradle:6.25.0'
+    classpath "org.jetbrains.kotlin:compose-compiler-gradle-plugin:$kotlin_version"
   }
 }
 
@@ -23,6 +24,7 @@ if (project == rootProject) {
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-parcelize'
+apply plugin: 'org.jetbrains.kotlin.plugin.compose'
 //apply plugin: 'kotlin-android-extensions'
 
 def getExtOrDefault(name) {
@@ -60,6 +62,7 @@ android {
 
   buildFeatures {
     viewBinding true
+    compose true
   }
 }
 
@@ -164,4 +167,9 @@ dependencies {
   androidTestImplementation "androidx.test:core:1.4.0"
   androidTestImplementation 'androidx.test:runner:1.1.0'
   androidTestImplementation "org.mockito:mockito-core:3.+"
+
+  implementation "androidx.compose.ui:ui:1.7.8"
+  implementation "androidx.compose.material:material:1.7.8"
+  implementation "androidx.compose.ui:ui-tooling-preview:1.7.8"
+  debugImplementation "androidx.compose.ui:ui-tooling:1.7.8"
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -18,7 +18,7 @@ buildscript {
     classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     classpath 'com.diffplug.spotless:spotless-plugin-gradle:6.25.0'
     // only use this old compose-compiler plugin when Kotlin <= 2.0
-    if (kotlinMajor <= 2) {
+    if (kotlinMajor >= 2) {
       classpath "org.jetbrains.kotlin:compose-compiler-gradle-plugin:$kotlin_version"
     }
   }
@@ -38,7 +38,7 @@ apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-parcelize'
 // Only apply the compose plugin if we have the old compose-compiler on the classpath
-if (kotlinMajor <= 2) {
+if (kotlinMajor >= 2) {
   apply plugin: 'org.jetbrains.kotlin.plugin.compose'
 }
 //apply plugin: 'kotlin-android-extensions'
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,7 +1,8 @@
 buildscript {
+  // Buildscript is evaluated before everything else so we can't use getExtOrDefault
   def kotlin_version = rootProject.ext.has('kotlinVersion')
-  ? rootProject.ext.get('kotlinVersion')
-  : project.properties['StripeSdk_kotlinVersion']
+    ? rootProject.ext.get('kotlinVersion')
+    : project.properties['StripeSdk_kotlinVersion']
 
   def kotlinMajor = kotlin_version.tokenize('\\.')[0].toInteger()
 
@@ -28,7 +29,9 @@ if (project == rootProject) {
     return
 }
 
-def kotlin_version = getExtOrDefault('kotlinVersion')
+def kotlin_version = rootProject.ext.has('kotlinVersion')
+    ? rootProject.ext.get('kotlinVersion')
+    : project.properties['StripeSdk_kotlinVersion']
 def kotlinMajor = kotlin_version.tokenize('\\.')[0].toInteger()
 
 apply plugin: 'com.android.library'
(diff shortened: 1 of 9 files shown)


android/build.gradle
```diff
@@ -164,4 +181,9 @@ dependencies {
   androidTestImplementation "androidx.test:core:1.4.0"
   androidTestImplementation 'androidx.test:runner:1.1.0'
   androidTestImplementation "org.mockito:mockito-core:3.+"
+
+  implementation "androidx.compose.ui:ui:1.7.8"
+  implementation "androidx.compose.material:material:1.7.8"
+  implementation "androidx.compose.ui:ui-tooling-preview:1.7.8"
```
jaynewstrom-stripe, line 187: Is this necessary?

android/src/main/java/com/reactnativestripesdk/PaymentOptionDisplayDataMapper.kt
```diff
@@ -0,0 +1,39 @@
...
+ * that can be sent over the RN bridge.
+ */
+@OptIn(ExperimentalEmbeddedPaymentElementApi::class)
+fun EmbeddedPaymentElement.PaymentOptionDisplayData.toWritableMap(): WritableMap {
+  val map = Arguments.createMap()
+
+  map.putString("label", label)
+  map.putString("paymentMethodType", paymentMethodType)
+
+  val billingMap = Arguments.createMap()
```
jaynewstrom-stripe, line 17: Should there be a reusable place to do this logic? My guess is it already exists, and I'm scared we'd update one place and not the other.

android/build.gradle
jaynewstrom-stripe: Just the one the comment is on.

android/src/main/java/com/reactnativestripesdk/EmbeddedPaymentElementView.kt
```diff
@@ -0,0 +1,200 @@
...
+        reactContext
+          .getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter::class.java)
+          .emit("embeddedPaymentElementFormSheetConfirmComplete", map)
+      },
+    )
+  }
+
+  @Composable
+  override fun Content() {
+    val embedded = rememberEmbeddedPaymentElement(builder)
```
jaynewstrom-stripe, line 106: This makes me think we should create an initializer that doesn't require compose.

android/build.gradle
porter-stripe: Tested this and looks like we only need

implementation "androidx.compose.ui:ui:1.7.8"

android/src/main/java/com/reactnativestripesdk/EmbeddedPaymentElementView.kt
jaynewstrom-stripe: Although, it's probably fine.

android/src/main/java/com/reactnativestripesdk/EmbeddedPaymentElementView.kt
```diff
@@ -0,0 +1,200 @@
...
+        }
+      }
+    }
+
+    LaunchedEffect(embedded) {
+      embedded.paymentOption.collect { opt ->
+        val optMap = opt?.toWritableMap()
+        val payload =
+          Arguments.createMap().apply {
+            // TODO: image?
```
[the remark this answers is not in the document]
porter-stripe: Yeah, just left a TODO incase we ever want to do it, I think that's the right spot b/c it's in compose code. But I can remove it.

android/build.gradle
jaynewstrom-stripe: We shouldn't need to do this at all. I'll fix in a future stripe-android release. https://github.com/stripe/stripe-android/pull/10721

android/src/main/java/com/reactnativestripesdk/EmbeddedPaymentElementView.kt
samer-stripe: I thought about this too, but I think doesn't allow two-step to be possible.

samer-stripe: There is a somewhat equivalent API I've thought about using to make 2-step work here with [useId](https://react.dev/reference/react/useId) but then we would need to expose an API for RN to allow custom IDs to be passed.

android/src/main/java/com/reactnativestripesdk/PaymentOptionDisplayDataMapper.kt
porter-stripe: Surprisingly after looking around we do not have anything that turns a PaymentSheet.BillingDetails into a Map. We have something that turns a PaymentMethod.BillingDetails into a Map but nothing for PaymentSheet. I'll move this logic into a place it's easier to reuse in the future.

android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt
```diff
@@ -509,7 +509,7 @@ class PaymentSheetFragment(
       }
 
     @Throws(PaymentSheetException::class)
-    private fun buildIntentConfiguration(intentConfigurationParams: Bundle?): PaymentSheet.IntentConfiguration? {
+    internal fun buildIntentConfiguration(intentConfigurationParams: Bundle?): PaymentSheet.IntentConfiguration? {
```
samer-stripe, line 512: Should we move these out of the fragment?

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
```diff
@@ -1254,6 +1276,35 @@ class StripeSdkModule(
...
   }
 
+  private fun performOnEmbeddedView(
+    viewTag: Int,
+    promise: Promise,
+    action: EmbeddedPaymentElementView.() -> Unit,
+  ) {
+    val uiManager =
+      reactApplicationContext
+        .getNativeModule(UIManagerModule::class.java)
```
samer-stripe, line 1286: This has been deprecated in the new architecture. In fact, I believe this makes the SDK incompatible with the new architecture again. Can we use `UIManagerHelper` here instead?

android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt
porter-stripe: We could, but I'd like to avoid refactors in this PR. Could be a good follow up.

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
porter-stripe: Updated it, but I think I'll need to find a different workaround in iOS too?
https://github.com/stripe/stripe-react-native/pull/1898/files#diff-a74bbe3c70d6d07338126ceffec8d9db707058a3e49ad8c3162bac8d32635484R20


I guess we can test more once the new arch lands, it's hard to tell what works/won't :/

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
```diff
@@ -1254,6 +1277,36 @@ class StripeSdkModule(
...
+    viewTag: Int,
+    promise: Promise,
+    action: EmbeddedPaymentElementView.() -> Unit,
+  ) {
+    val uiManager =
+      UIManagerHelper
+        .getUIManagerForReactTag(
+          reactApplicationContext as ReactContext,
+          viewTag,
+        ) as? FabricUIManager
```
samer-stripe, line 1291: I'm surprised this works even though we don't have the new architecture enabled.

samer-stripe: Do we need to fetch the `UIManager` as `FabricUIManager`?

porter-stripe: I'm not 100% sure this works actually since I can't test w/ new architecture yet, but it's a start. I think we'll need to test this after new arch lands to confirm it works.

Issue #1871, opened 2025-03-23
Expo android prebuild failing


Alvi24:
when running  expo run:android with the latest react-native-stripe version 
("@stripe/stripe-react-native": "^0.43.0",)
prebuild fails with this error:
Execution failed for task ':stripe_stripe-react-native:compileDebugKotlin'.
> A failure occurred while executing org.jetbrains.kotlin.compilerRunner.GradleCompilerRunnerWithWorkers$GradleKotlinCompilerWorkAction

Jackmekiss:
Same here

jaynewstrom-stripe:
Hi, it looks like the dependencies in your project aren't aligned. We updated the kotlin version in a recent Android SDK version, could you update it in your project as well? 

https://github.com/stripe/stripe-android/releases/tag/v21.6.0 Describes the dependency versions we're using.

Alvi24:
i have installed the latest version of @stripe/stripe-react-native, and also i don't have stripe-android explicitly downloaded since i am using react native

jaynewstrom-stripe:
@Alvi24 you need to update your project to compile with kotlin 2.1.10 or later.

crrobinson14:
We have updated to the latest Stripe SDK (`@stripe/stripe-react-native` v0.43.0) and tried both the Expo 52 Kotlin default of 1.9.25 as well as 2.1.10 and are still having this issue ourselves. I can provide more detailed build logs but wanted to ask first what was most helpful because they're quite long.

Oddly, one of the errors we get in our build log is the old `isWhitespace is missing` error reported over a year ago at https://github.com/stripe/stripe-react-native/issues/1321. I don't see how that makes any sense because Expo 52 defaults to Kotlin 1.9.25 which is much later than what's reported there. Just including it for reference.

crrobinson14:
Update: We temporarily worked around this issue by downgrading Stripe to `0.38.6` which appears to be the last version compatible with stock Expo 52. We didn't feel upgrading Kotlin was appropriate here - that's not a simple change at all in Expo. There's a build-config plugin with a setting for it, but changing it to v2.1.10+ causes a ton of other issues such as complaints about `kotlin-compose-compiler-plugin-embeddable`. We put a few hours into trying to sort that out but it was a huge can of worms and downgrading this module was just easier.

Perhaps somebody from Stripe can correct me if I'm wrong but at first glance it appears only a few things are lost by downgrading here:
- New CustomerSheet functionality
[...]

PR #1900, opened 2025-04-24
Fix crash on fragment restoration

janicduplessis:
## Summary

- Introduce StripeFragment base class that makes it easier to handle fragment recreation without crashing.
- Remove arguments from all fragment constructors to prevent crashes.
- Try to dismiss stripe ui on activity re-creation to prevent apps from being in inconsistent states.

## Motivation

When activities are destroyed while the app is in background the react-native sdk crashes once they are restored. In general react-native apps do not support activity re-creation and just reset their ui state if it happens. This can cause apps to be in an inconsistent state if stripe ui is still displayed. 

## Testing

- [x] I tested this manually
- [ ] I added automated tests

To test use the "Don't keep Activities" developer setting in the emulator.

I tried testing all of the different fragments by opening them then backgrounding the app. Most of them would cause the app to crash, after this the app will re-open properly.

## Documentation

Select one: 


android/src/androidTest/java/com/reactnativestripesdk/paymentsheet/PaymentSheetFragmentTest.kt
```diff
@@ -8,6 +8,13 @@ import org.junit.Assert
 import org.junit.Test
 
 class PaymentSheetFragmentTest {
+  @Test
+  fun hasNoArgsConstructor() {
+    // This is just a type check to make sure to keep a no args constructor
+    // so that when the fragment is re-created it doesn't crash.
+    PaymentSheetFragment()
```
jaynewstrom-stripe, line 15: Should we add tests like this for the other fragments too?

janicduplessis: Would be good, I only set it up in that one since it already had a test file, but I could create it for all other ones too.

janicduplessis: Added tests for all fragments

android/src/main/java/com/reactnativestripesdk/PaymentLauncherFragment.kt
```diff
@@ -161,40 +187,35 @@ class PaymentLauncherFragment(
...
+            retrievePaymentIntent(handleNextActionPaymentIntentClientSecret!!, stripeAccountId)
           } else if (setupIntentClientSecret != null) {
-            retrieveSetupIntent(setupIntentClientSecret, stripeAccountId)
+            retrieveSetupIntent(setupIntentClientSecret!!, stripeAccountId)
           } else if (handleNextActionSetupIntentClientSecret != null) {
-            retrieveSetupIntent(handleNextActionSetupIntentClientSecret, stripeAccountId)
+            retrieveSetupIntent(handleNextActionSetupIntentClientSecret!!, stripeAccountId)
           } else {
             throw Exception("Failed to create Payment Launcher. No client secret provided.")
           }
```
samer-stripe, line 221: Should we unwrap these instead? I think that will be safer than asserting its not null
```kotlin
paymentIntentClientSecret?.let {
    retrievePaymentIntent(it, stripeAccountId)
} ?: handleNextActionPaymentIntentClientSecret?.let {
    retrievePaymentIntent(it, stripeAccountId)
} ?: setupIntentClientSecret?.let {
    retrievePaymentIntent(it, stripeAccountId)
} ?: handleNextActionSetupIntentClientSecret?.let {
    retrievePaymentIntent(it, stripeAccountId)
} ?: throw Exception("Failed to create Payment Launcher. No client secret provided.")
```

samer-stripe (dismissed):

janicduplessis:
@samer-stripe Should be good to go now, had a little issue with the refactor [here](https://github.com/stripe/stripe-react-native/pull/1900#discussion_r2064184986), it actually calls 2 different function in different branches.

samer-stripe (approved):
Pulled the branch and tested, thank you for fixing this!

marcshilling:
What release is this in?

janicduplessis:
0.46.0

marcshilling:
> 0.46.0

Unfortunately seems like I can't upgrade to that version because I'm on RN 0.76 :(
https://github.com/stripe/stripe-react-native/issues/1924#issuecomment-2867227374
https://github.com/stripe/stripe-react-native/issues/1871#issuecomment-2861152694

janicduplessis:
We're looking into these issues, and should have a fix in the next release.

marcshilling:
> > 0.46.0
> 
> Unfortunately seems like I can't upgrade to that version because I'm on RN 0.76 :( [#1924 (comment)](https://github.com/stripe/stripe-react-native/issues/1924#issuecomment-2867227374) [#1871 (comment)](https://github.com/stripe/stripe-react-native/issues/1871#issuecomment-2861152694)

For others - was able to get things building by following [this comment](https://github.com/stripe/stripe-react-native/issues/1924#issuecomment-2867227374)

PR #1904, opened 2025-04-28
Fix compose views inside react-native-screens fragments

janicduplessis:
## Summary

This fixes an issue with screens and compose. The issue is described here https://github.com/software-mansion/react-native-screens/issues/2098. Basically if a screen is pushed on top of a view using compose when the screen is dismissed the compose view will disappear.

We can reproduce this issue with our Embedded component in the example app with the "Push screen" button added in this PR. Tap on the button then go back and the Embedded view is no longer visible.

What happens is that react native screens reuses views when a fragment is destroyed, which composes doesn't support by default. So when pushing a new screen the previous screen fragment is destroyed causes the compose view to be disposed. When the fragment is re-created when going back, the disposed compose view is re-used which is not supported.

To fix this we need to avoid dispatching the fragment DESTROY event so the compose view is not disposed. To do this we can create a custom lifecycle that will follow the activity's lifecycle events instead of the fragment's. When the view is actually destroyed by react we can dispatch the DESTROY event manually.

## Motivation

Make embedded view work properly when using react native screens.

## Testing

- [x] I tested this manually
- [x] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/EmbeddedPaymentElementView.kt
+++ b/android/src/main/java/com/reactnativestripesdk/EmbeddedPaymentElementView.kt
@@ -3,7 +3,6 @@ package com.reactnativestripesdk
 import android.content.Context
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
-import androidx.compose.ui.platform.AbstractComposeView
 import androidx.lifecycle.findViewTreeLifecycleOwner
 import androidx.lifecycle.lifecycleScope
 import com.facebook.react.bridge.Arguments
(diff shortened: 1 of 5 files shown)


janicduplessis:
I think we could upstream a similar fix in react-native-screens to avoid having to do this, but this will allow it to work now without having to wait for the fix to land and users to update. If we rollout a fix in RN screens we can eventually remove this code.

samer-stripe (dismissed):

janicduplessis:
@samer-stripe Conflicts fixed!

samer-stripe (approved):

PR #1905, opened 2025-04-29
Fix Android `Embedded` calls to native functions

samer-stripe:
## Summary
Fix Android `Embedded` calls to native functions

## Motivation
While working on the height solution for Embedded RN, I noticed that the calls to `confirm` and `clearPaymentOption` did not execute. This PR fixes that so the calls go through.

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/src/specs/NativeEmbeddedPaymentElement.ts
+++ b/src/specs/NativeEmbeddedPaymentElement.ts
@@ -7,6 +7,7 @@ import codegenNativeComponent from 'react-native/Libraries/Utilities/codegenNati
 import { EmbeddedPaymentElementConfiguration } from '../types/EmbeddedPaymentElement';
 import { IntentConfiguration } from '../types/PaymentSheet';
 import type { UnsafeMixed } from './utils';
+import codegenNativeCommands from 'react-native/Libraries/Utilities/codegenNativeCommands';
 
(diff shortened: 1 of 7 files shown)


src/specs/NativeEmbeddedPaymentElement.ts
```diff
@@ -18,6 +19,17 @@ export interface NativeProps extends ViewProps {
...
+export interface NativeCommands {
+  confirm: (viewRef: React.ElementRef<HostComponent<NativeProps>>) => void;
+  clearPaymentOption: (
+    viewRef: React.ElementRef<HostComponent<NativeProps>>
+  ) => void;
+}
+
+export const Commands: NativeCommands = codegenNativeCommands<NativeCommands>({
+  supportedCommands: ['confirm', 'clearPaymentOption'],
+});
```
samer-stripe, line 31: Relying on commands to the view rather than directly calling the module. There's a path to also do the same for `configure` but we can do that as a follow-up if needed.

porter-stripe (approved):

porter-stripe (dismissed):

porter-stripe (approved):

PR #1907, opened 2025-04-29
Update the CHANGELOG for v0.45

davidme-stripe:
## Summary
Update changelog for v0.45, which was just released.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # CHANGELOG
-## Unreleased
+## 0.45.0 - 2025-04-29
 
 ** Features **
+- Support for the New Architecture in React Native v0.68 or later.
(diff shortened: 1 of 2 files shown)


toluo-stripe (approved):

CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
 # CHANGELOG
-## Unreleased
+## 0.45.0 - 2025-04-29
 
 ** Features **
+- Support for the New Architecture in React Native v0.68 or later.
```
janicduplessis, line 5: I actually missed this, I think we should mention a higher RN version, 0.68 is really old and didn't test that low. I did test on 0.76 so that could be the lowest, although it might still work on something lower than that, but I had some issues running the example app on lower versions.

PR #1964, opened 2025-06-05
Add CustomPaymentMethods API

porter-stripe:
## Summary
- Implements Custom Payment Methods for React Native.
- Adds E2E tests
- [API review](https://docs.google.com/document/d/1gjbQJcFxase1SnAUkqr5vmtpykW1AbOa6WarH65jp0M/edit?tab=t.0#heading=h.570u4leviueu)

## Motivation
- CPMs in RN

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/CustomPaymentMethodActivity.kt
+++ b/android/src/main/java/com/reactnativestripesdk/CustomPaymentMethodActivity.kt
@@ -3,6 +3,7 @@ package com.reactnativestripesdk
 import android.os.Bundle
 import android.view.MotionEvent
 import com.facebook.react.ReactActivity
+import java.lang.ref.WeakReference
 
 /**
  * A transparent activity that is launched when the Payment Element requests the
@@ -62,7 +63,7 @@ class CustomPaymentMethodActivity : ReactActivity() {
 
   companion object {
     @Volatile
-    private var currentActivityRef: java.lang.ref.WeakReference<CustomPaymentMethodActivity>? = null
+    private var currentActivityRef: WeakReference<CustomPaymentMethodActivity>? = null
 
     fun finishCurrent() {
       currentActivityRef?.get()?.let { activity ->
@@ -75,6 +76,6 @@ class CustomPaymentMethodActivity : ReactActivity() {
 
   override fun onStart() {
     super.onStart()
-    currentActivityRef = java.lang.ref.WeakReference(this)
+    currentActivityRef = WeakReference(this)
   }
 }


samer-stripe (dismissed):

android/src/main/java/com/reactnativestripesdk/CustomPaymentMethodActivity.kt
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * Its only purpose is to bring the app back to the foreground (the Stripe
+ * SDK launches its own proxy activity which pauses the host React Native
+ * activity). Having a React (transparent) activity on top ensures that React
+ * Native can display UI elements such as `Alert` dialogs coming from
+ * JavaScript.
+ *
+ * The activity uses a completely translucent theme so it is not visible to the
+ * end-user and is excluded from recents.
+ */
```
samer-stripe, line 21: Comment here doesn't make sense since we do show the activity here.

android/src/main/java/com/reactnativestripesdk/EmbeddedPaymentElementView.kt
```diff
@@ -56,9 +65,91 @@ class EmbeddedPaymentElementView(
...
+              CustomPaymentMethodActivity.finishCurrent()
+              return@remember
+            }
+
+          // Keep JS awake while React Native is backgrounded by Stripe SDK.
+          val keepJsAwakeTask =
+            KeepJsAwakeTask(reactContext.reactApplicationContext).apply { start() }
+
+          // Run on main coroutine scope.
+          kotlinx.coroutines.CoroutineScope(kotlinx.coroutines.Dispatchers.Main).launch {
```
samer-stripe: Can use `rememberCoroutineScope` above
```kotlin
val coroutineScope = rememberCoroutineScope()
val confirmCustomPaymentMethodCallback = ....
```

and re-use here.

android/src/main/java/com/reactnativestripesdk/EmbeddedPaymentElementView.kt
```diff
@@ -56,9 +65,91 @@ class EmbeddedPaymentElementView(
...
+            try {
+              // Give the CustomPaymentMethodActivity a moment to fully initialize
+              kotlinx.coroutines.delay(100)
```
samer-stripe: Can we simplify the imports here?

android/src/main/java/com/reactnativestripesdk/EmbeddedPaymentElementView.kt
porter-stripe: added!

android/src/main/java/com/reactnativestripesdk/EmbeddedPaymentElementView.kt
```diff
@@ -56,9 +67,92 @@ class EmbeddedPaymentElementView(
...
+              // Give the CustomPaymentMethodActivity a moment to fully initialize
+              delay(100)
+
+              // Emit event so JS can show the Alert and eventually respond via `customPaymentMethodResultCallback`.
+              stripeSdkModule.emitOnCustomPaymentMethodConfirmHandlerCallback(
+                mapFromCustomPaymentMethod(customPaymentMethod, billingDetails),
+              )
+
+              // Await JS result.
+              val resultFromJs = stripeSdkModule.customPaymentMethodResultCallback.await()
```
samer-stripe, line 119: I actually think this might break on config changes (screen rotation), can you check that CPMs here work with config changes?

porter-stripe: Yeah good flag, looks like it does. I think we have the same issue with `val resultFromJavascript = stripeSdkModule.embeddedIntentCreationCallback.await()`

unsure if theres a viable fix though.

samer-stripe: This looks like an SDK problem! Probably something we can have solved by in a separate issue!

android/src/main/java/com/reactnativestripesdk/CustomPaymentMethodActivity.kt
```diff
@@ -0,0 +1,80 @@
...
+        activity.runOnUiThread {
+          activity.finish()
+        }
+      }
+    }
+  }
+
+  override fun onStart() {
+    super.onStart()
+    currentActivityRef = java.lang.ref.WeakReference(this)
```
samer-stripe: Can we import `WeakReference` directly?

PR #1966, opened 2025-06-09
Add Form Insets and Primary Button height API

tjclawson-stripe:
## Summary

Add `formInsetValues` and `PrimaryButtonConfig.shapes.height` to `AppearanceParams`
Update `stripe-ios` to version `24.15.0`
Update `stripe-android` to version `21.17.+`

## Motivation

https://jira.corp.stripe.com/browse/MOBILESDK-3610

## Testing

- [X] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/PaymentSheetAppearance.kt
+++ b/android/src/main/java/com/reactnativestripesdk/PaymentSheetAppearance.kt
@@ -8,6 +8,7 @@ import android.os.Bundle
 import com.reactnativestripesdk.utils.PaymentSheetAppearanceException
 import com.stripe.android.paymentelement.ExperimentalEmbeddedPaymentElementApi
 import com.stripe.android.paymentsheet.PaymentSheet
+import com.stripe.android.uicore.StripeThemeDefaults
 
 @SuppressLint("RestrictedApi")
 fun buildPaymentSheetAppearance(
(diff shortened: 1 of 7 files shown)


src/types/PaymentSheet.ts
```diff
@@ -206,6 +206,9 @@ export type AppearanceParams = RecursivePartial<{
 
   /** Describes the appearance of the Embedded Mobile Payment Element */
   embeddedPaymentElement: EmbeddedPaymentElementAppearance;
+
+  /** Describes the inset values applied to Mobile Payment Element forms */
+  formInsetValues: EdgeInsetsConfig;
```
samer-stripe, line 211: Is this an internal type?

android/src/main/java/com/reactnativestripesdk/PaymentSheetAppearance.kt
```diff
@@ -403,6 +407,24 @@ private fun buildEmbeddedAppearance(
   return PaymentSheet.Appearance.Embedded(style = rowStyle)
 }
 
+private fun buildFormInsets(insetParams: Bundle?): PaymentSheet.Insets {
+  val defaultLeft = 20f
+  val defaultTop = 0f
+  val defaultRight = 20f
+  val defaultBottom = 40f
```
samer-stripe: Should we programmatically figure this out here so that we don't have to update this in case we ever update out defaults? I'm okay with creating an empty `Appearance` instance to pull out the default values.

ios/PaymentSheetAppearance.swift
```diff
@@ -314,6 +321,20 @@ internal class PaymentSheetAppearance {
             return defaultColor
           })
       }
+    
+    private class func buildFormInsets(params: NSDictionary) throws -> NSDirectionalEdgeInsets {
+        let top = params[PaymentSheetAppearanceKeys.TOP] as? CGFloat ?? PaymentSheet.Appearance.default.formInsets.top
+        let leading = params[PaymentSheetAppearanceKeys.LEFT] as? CGFloat ?? PaymentSheet.Appearance.default.formInsets.leading
+        let bottom = params[PaymentSheetAppearanceKeys.BOTTOM] as? CGFloat ?? PaymentSheet.Appearance.default.formInsets.bottom
+        let trailing = params[PaymentSheetAppearanceKeys.RIGHT] as? CGFloat ?? PaymentSheet.Appearance.default.formInsets.trailing
```
samer-stripe, line 329: As mentioned in the other comment, let's do something like this with an empty `Appearance` object.

android/src/main/java/com/reactnativestripesdk/PaymentSheetAppearance.kt
tjclawson-stripe: Because `formInsetValues` is internal as well as it's companion object, we would have to use `StripeThemeDefaults.formInsets` to get the defaults from `stripe-android` I can update to do so if you're alright with that.

ios/PaymentSheetAppearance.swift
tjclawson-stripe: Meaning something like 
```
let appearance = PaymentSheet.Appearance()
let top = params[PaymentSheetAppearanceKeys.TOP] as? CGFloat ?? apperance.formInsets.top
...
```

samer-stripe: Since we expose the defaults on iOS, what you have here is fine. This was more of a comment to do this for Android as well.

android/src/main/java/com/reactnativestripesdk/PaymentSheetAppearance.kt
samer-stripe: Not blocking for this PR, but would be great if we can do it eventually.

tjclawson-stripe: Updated

src/types/PaymentSheet.ts
tjclawson-stripe: `EdgeInsetsConfig`? It's an [exported type](https://github.com/stripe/stripe-react-native/blob/a8b18831752d/src/types/PaymentSheet.ts#L338) that was added for embedded insets

samer-stripe (approved):

PR #1985, opened 2025-06-26
Fixing E2E tests and Sofort Deprecation and Glitch Migration 

tianzhao-stripe:
## Summary
Use a code sandbox project for publicly available backend.

Use sandboxApps to host our demo which has to be on all the time.

Tests were failing because the 

- email field could not be found
- the link bank account info entry page has new button text for filling accepting stripe terms.
- the chrome welcome page is appearing after the institution picker page
- the country was being set to Austria
- the primary button was being hidden by the keyboard
- Sofort is no longer supported

## Motivation
RIP Glitch & CI is better when it works

## Testing

- [X] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/e2e-tests/android-only/becs-payment.yml
+++ b/e2e-tests/android-only/becs-payment.yml
@@ -26,7 +26,6 @@ appId: ${APP_ID}
 - tapOn:
     text: "Account number"
 - inputText: "000123456"
-- hideKeyboard
 - scrollUntilVisible:
     element:
       text: "Pay"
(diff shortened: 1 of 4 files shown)


gbirch-stripe:
So to check my understanding - the reason the tests started failing is because we hadn't migrated off of Glitch and it went down? What's the reason that the other changes were needed to the tests? What changed that the flows that we were working fine before no longer do?

PR #1991, opened 2025-07-02
0.49.0 proposal

porter-stripe:
Propose 0.48.1


gbirch-stripe (changes requested):
Shouldn't it be a minor version (0.49.0) given the amount of changes? Also typo on third bullet

Issue #1965, opened 2025-06-06
confirmPlatformPayPayment times out but succeeds


codan84:
**Describe the bug**
When using `confirmPlatformPayPayment` on iPhone it seems to take a very long time and eventually causes the payment to time-out. However, the call eventually resolves and succeeds. Thus I suspect it just takes too long and the phone itself displays 'payment not completed' message. I can see payment successful in Stripe dashboard and in the response from the `confirmPlatformPayPayment` call.

**To Reproduce**
Code I am running, with some debug logs to see how long the call takes:
```
    console.log(`>>> ${new Date().toISOString()} >>> about to pay`)
    const { error, paymentIntent } = await confirmPlatformPayPayment(
      clientSecret,
      {
        applePay: {
          cartItems: [
            {
              label: "Goods",
              amount: order.amount, //ie '22.50'
              paymentType: PlatformPay.PaymentType.Immediate,
            }
          ],
          merchantCountryCode: "GB",
          currencyCode: "GBP",
        },
        googlePay: {
         /***/
        },
      },
    );
    console.log(`>>> ${new Date().toISOString()} >>> paid: error=${error} & status=${paymentIntent?.status}`)
```

Example logs from the above:
```
 LOG  >>> 2025-06-06T09:05:46.291Z >>> about to pay
 LOG  >>> 2025-06-06T09:06:26.139Z >>> paid: error=undefined & status=Succeeded
```

And another:
```
 LOG  >>> 2025-06-06T08:58:46.044Z >>> about to pay
 LOG  >>> 2025-06-06T08:59:22.496Z >>> paid: error=undefined & status=Succeeded
```

(I pressed the power button twice to confirm payment immediately in both cases, so it is not me delaying)

**Expected behavior**
Payment processing is faster and does not time out on the phone.

**Screenshots**

![Image](https://github.com/user-attachments/assets/3fa35b3f-5dbc-491d-8459-0eb81cb6b750)

**Smartphone (please complete the following information):**
 - Device: iPhone 12
 - OS: iOS 18.5
 - @stripe/stripe-react-native: 0.45.0
 - expo: 53.0.0
 - react-native: 0.79.2

**Additional context**

This is running on test data.

ulrikem:
This happens in my app as well with confirmPlatformPaySetupIntent().

- Device: iPhone 16 Pro (real device) and iPhone 15 Pro (simulator)
- OS: iOS 18.5
[...]

PR #1993, opened 2025-07-02
Fix Apple Pay sheet hanging issues in new architecture

joyceqin-stripe:
## Summary

Added boolean props to NativeApplePayButton to indicate whether or not a callback was truly passed through the prop. When tapping the Apple Pay button, it checks whether the callback was passed in, otherwise set the callback to nil.
## Motivation

Issue reported here: https://github.com/stripe/stripe-react-native/issues/1965
The PaymentSheet would successfully confirm a PaymentIntent, but the Apple Pay sheet would hang and eventual say "Payment not completed". This is because we [rely on callbacks being nil](https://github.com/stripe/stripe-react-native/blob/master/ios/ApplePayViewController.swift#L183) when the callback is not passed through the prop, but with the new architecture, we [always set the callback](https://github.com/stripe/stripe-react-native/blob/master/ios/NewArch/ApplePayButtonComponentView.mm#L80). Therefore, we always enter the clause assuming the existence of the callback, and when it's not implemented, the completion handler never gets called.
## Testing

- [X] I tested this manually
- [x] I added automated tests

## Documentation
Added comments
Select one: 

### DIFF
--- a/e2e-tests/ios-only/local-only/applepay-buy.yml
+++ b/e2e-tests/ios-only/local-only/applepay-buy.yml
@@ -1,0 +1,12 @@
+appId: ${APP_ID}
+---
+- launchApp
+- tapOn: "Wallets"
+- tapOn: "Apple Pay"
+- tapOn: "Buy with Apple\_Pay"
+- tapOn: "Pay with Passcode"
+- assertVisible:
+    text: "Done"
+- assertVisible:
+    text: "Success"
+- tapOn: "OK"


porter-stripe:
Can we add an E2E test for this? https://github.com/stripe/stripe-react-native/tree/master/e2e-tests

gbirch-stripe:
> Can we add an E2E test for this? https://github.com/stripe/stripe-react-native/tree/master/e2e-tests

Partially related, do we currently have a plan for keeping CI times reasonable? It seems like maestro is pretty slow — we're already up to ~40min. I guess we could start splitting up the tests into smaller jobs?

porter-stripe:
@gbirch-stripe yes, see here: https://docs.google.com/document/d/1jsPHkTYg4VaYTwSArtc7wdimPqdSOCvcBVAGx5iFJDc/edit?usp=sharing

joyceqin-stripe:
Added an e2e test, but it has to be run locally because it doesn't seem like the CI can handle Apple Pay when it isn't local

porter-stripe (approved):

PR #1994, opened 2025-07-08
Update tests to fix CI

tianzhao-stripe:
## Summary
Tests were failing because the bank account info entry page has new button text for filling accepting stripe terms.

Tests were failing because the country was being set to Austria

## Motivation
CI is better when it's not broken.

## Testing
- [x] I tested this manually
- [x] I added automated tests

## Documentation

Select one: 


tianzhao-stripe:
fixing ci tests requires glitch migration

PR #2006, opened 2025-07-18
Fix single color support EMPE

porter-stripe:
## Summary
- We allow passing in a single color rather than just a light and dark. This fixes that.

## Motivation
- Bug fix


ios/PaymentSheetAppearance.swift
```diff
@@ -340,13 +335,32 @@ internal class PaymentSheetAppearance {
             return defaultColor
           })
       }
-    
+
```
tianzhao-stripe, line 338: Nit: Include a comment describing the function similar to Android

```suggestion
/**
 * Parses a themedColor from [params] at [key]. Supports both:
 * - Single hex string: "#RRGGBB"
 * - Light/dark object: { "light": "#RRGGBB", "dark": "#RRGGBB" }
 * For light/dark objects, chooses the appropriate color based on current UI mode.
 * Falls back to [defaultColor] if no color is provided.
 */
```

tianzhao-stripe (approved):

PR #2032, opened 2025-08-04
Add image to embedded payment option display data

porter-stripe:
## Testing

- [ ] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/PaymentOptionDisplayDataMapper.kt
+++ b/android/src/main/java/com/reactnativestripesdk/PaymentOptionDisplayDataMapper.kt
@@ -1,7 +1,11 @@
+import android.graphics.drawable.Drawable
 import com.facebook.react.bridge.Arguments
 import com.facebook.react.bridge.WritableMap
+import com.reactnativestripesdk.getBase64FromBitmap
+import com.reactnativestripesdk.getBitmapFromDrawable
 import com.reactnativestripesdk.utils.mapFromPaymentSheetBillingDetails
 import com.stripe.android.paymentelement.EmbeddedPaymentElement
+import kotlinx.coroutines.withTimeout
 
 /**
  * Serialize Stripe's PaymentOptionDisplayData into a WritableMap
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/reactnativestripesdk/PaymentOptionDisplayDataMapper.kt
```diff
@@ -12,4 +16,29 @@ fun EmbeddedPaymentElement.PaymentOptionDisplayData.toWritableMap(): WritableMap
     putString("label", label)
     putString("paymentMethodType", paymentMethodType)
     putMap("billingDetails", mapFromPaymentSheetBillingDetails(billingDetails))
+
+    // Direct access to imageLoader and call it synchronously with timeout
+    val imageBase64 =
+      try {
+        val imageLoaderField = this@toWritableMap.javaClass.getDeclaredField("imageLoader")
```
jaynewstrom-stripe: This might not work if the merchant uses progaurd/R8. 

That being said, I'm not convinced this is the right approach to begin with!

What are you trying to do here?

porter-stripe: Trying to access the imageLoader, which is private, any chance we can make it like RESTRICT_TO ilb?

android/src/main/java/com/reactnativestripesdk/PaymentOptionDisplayDataMapper.kt
```diff
@@ -12,4 +15,24 @@ fun EmbeddedPaymentElement.PaymentOptionDisplayData.toWritableMap(): WritableMap
     putString("label", label)
     putString("paymentMethodType", paymentMethodType)
     putMap("billingDetails", mapFromPaymentSheetBillingDetails(billingDetails))
+
+    // Direct access to imageLoader and call it synchronously with timeout
+    val imageBase64 =
+      try {
+        // Load synchronously with 5 second timeout
+        val drawable =
+          kotlinx.coroutines.runBlocking {
```
jaynewstrom-stripe: What thread does this happen on?

android/src/main/java/com/reactnativestripesdk/PaymentOptionDisplayDataMapper.kt
```diff
@@ -12,4 +15,24 @@ fun EmbeddedPaymentElement.PaymentOptionDisplayData.toWritableMap(): WritableMap
...
+            withTimeout(5000L) {
```
jaynewstrom-stripe: Is there a way to fallback if something fails?

android/src/main/java/com/reactnativestripesdk/PaymentOptionDisplayDataMapper.kt
porter-stripe: Looks like it is running on the main thread which isn't great! However, it is [consistent](https://github.com/stripe/stripe-ios/blob/5ff542f3c59a85f86de387fe8eb6ea24cdf2e308/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement%2BInternal.swift#L361) with the iOS native lib 🙈.

I pushed up a change to run it on a different thread than main, however my Android threading knowledge is limited so lmk what you think. Verified it still works too.

android/src/main/java/com/reactnativestripesdk/PaymentOptionDisplayDataMapper.kt
porter-stripe: We could fallback, but I think we'd need to bundle some resource in the RN SDK. My preference would be that the imageLoader resolves with some default/placeholder image if loading fails to keep the logic out of the RN SDK. Unsure what "generic" payment method icon we could use. My immediate reaction is a card icon, but that may be inaccurate for most payment method types.

PR #2035, opened 2025-08-05
Add mandate text to payment option

porter-stripe:
## Summary
- Adds the ability for merchants to display the mandate themselves in their app
- We vend an HTML string that can be rendered, ensuring links can be tapped.

## Motivation
https://docs.google.com/document/d/1OaY8oTGsO2dPS7HhZTCKB-tLoCN2m_S1ag0kJ83eD_E/edit?tab=t.0#heading=h.570u4leviueu

## Documentation
- Doc changes to follow after the next deploy

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -21,11 +21,18 @@ jobs:
     name: e2e-android-test
     runs-on: ubuntu-24.04
     strategy:
+      fail-fast: false
       matrix:
         react_arch: [old_arch, new_arch]
+        shard_index: [0, 1, 2]
     env:
       MAESTRO_CLI_ANALYSIS_NOTIFICATION_DISABLED: true
       NEW_ARCH_ENABLED: ${{ matrix.react_arch == 'new_arch' }}
+      RCT_NEW_ARCH_ENABLED: ${{ matrix.react_arch == 'new_arch' && '1' || '0' }}
+      SHARD_COUNT: 3
+      SHARD_INDEX: ${{ matrix.shard_index }}
+      RETRY_SLEEP_FAST: 10
+      RETRY_SLEEP_SLOW: 45
     steps:
       - name: Delete unnecessary tools 🔧
         uses: jlumbroso/free-disk-space@v1.3.1
@@ -113,7 +120,7 @@ jobs:
         if: always()
         uses: actions/upload-artifact@v4
         with:
-          name: maestro-artifacts-android-${{ matrix.react_arch }}
+          name: maestro-artifacts-android-${{ matrix.react_arch }}-shard-${{ matrix.shard_index }}
           path: e2e-artifacts
           include-hidden-files: true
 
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/reactnativestripesdk/PaymentOptionDisplayDataMapper.kt
```diff
@@ -38,3 +48,8 @@ suspend fun EmbeddedPaymentElement.PaymentOptionDisplayData.toWritableMap(): Wri
       }
     putString("image", imageBase64)
   }
+
+fun AnnotatedString.toHtmlString(): String {
```
jaynewstrom-stripe, line 60: Is it possible to add unit tests for something like this?

porter-stripe: I think it may be possible! However I'm on the fence as we'd be unit testing two system APIs (`toSpannable` and `toHtml`). I suppose we could test that `toSpannable` and `toHtml` preserve the attributes we care about. If we think there's value (maybe some Android quirk I'm overlooking) I'm happy to add some unit tests as I think it's feasible.

jaynewstrom-stripe: I think it's probably a good idea to write a unit test that ensures we're preserving links from the original, as well as a simple text based example.

android/src/androidTest/java/com/reactnativestripesdk/mappers/PaymentOptionDisplayDataMapperTest.kt
```diff
@@ -0,0 +1,65 @@
...
+      "Result should contain the text content",
+      result.contains("By continuing, you agree"),
+    )
+    Assert.assertTrue(
+      "Result should contain link text",
+      result.contains("Privacy Policy"),
+    )
+    Assert.assertTrue(
+      "Result should contain link text",
+      result.contains("Terms of Service"),
```
jaynewstrom-stripe: I'd expect this to assert there's an a href in here.

porter-stripe: Thanks for probing, during my tests manually I can see the links being preserved but during tests there are some cases not handled. Going to beef up the implementation of toHtml.

porter-stripe: After testing more and throwing more test cases at the spannable API, it turns out it drops some stuff like links. So I think the best path forward is to recreate a basic HTML string based on the attributes in the string, took some inspiration from [here](https://github.com/coderforlife/MarkdownEditor/blob/3ac7ce05f180e278bc0c440d31e1f0b26b3af375/notes/html-decode.kt#L2282) and help from Claude 🙈 Added more tests this time around.

PR #2073, opened 2025-08-20
Propose v0.50.4

porter-stripe:
- [x] Ensure the CHANGELOG is up to date with all relevant commits since the last release
- [x] Add the version number for this release & the date to the CHANGELOG, underneath "## Unreleased" 
  - e.g. "## 1.2.3 - 2022-02-14"
- [x] Update the README if necessary (this is only required when there are breaking changes in the release, such as dropping support for an iOS || Android version)

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -5,9 +5,6 @@ on:
     branches: [master]
   pull_request:
     branches: ['**']
-    paths-ignore:
-      - '**/*.md'
-      - '**/*.MD'
   schedule:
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -5,9 +5,6 @@ on:
     branches: [master]
   pull_request:
     branches: ['**']
-    paths-ignore:
-      - '**/*.md'
-      - '**/*.MD'
```
porter-stripe, line 10: These weren't working as expected, will need to try out a new approach in a later PR.

jaynewstrom-stripe (approved):

PR #2079, opened 2025-08-28
Add PaymentSheet FlowController didCancel presentation APIs to React Native SDK

joyceqin-stripe:
## Summary

iOS and Android added a boolean for when a user Canceled/dismissed the Payment Sheet within a Flow Controller custom flow. A live user needs this functionality in React Native as well.
## Motivation

https://jira.corp.stripe.com/browse/RUN_MOBILESDK-4599
## Testing

- [X] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt
+++ b/android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt
@@ -132,7 +132,9 @@ class PaymentSheetFragment :
             val option: WritableMap = WritableNativeMap()
             option.putString("label", it.label)
             option.putString("image", imageString)
-            createResult("paymentOption", option)
+            val result = createResult("paymentOption", option)
+            result.putBoolean("didCancel", paymentOptionResult.didCancel)
+            result
           }
             ?: run {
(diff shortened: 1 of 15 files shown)


android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt
```diff
@@ -132,7 +132,9 @@ class PaymentSheetFragment :
             val option: WritableMap = WritableNativeMap()
             option.putString("label", it.label)
             option.putString("image", imageString)
-            createResult("paymentOption", option)
+            val result = createResult("paymentOption", option)
```
tianzhao-stripe: can you update the createResult function in Mappers.kt instead of doing this

it'll look something like this

```

internal fun createResult(
  key: String,
  value: WritableMap,
  additionalFields: Map<String, Any>?
): WritableMap {
  val map = WritableNativeMap()
  map.putMap(key, value)
  
  additionalField?.forEach{map.put(it.key, it.value)}
  return map
}
```

tianzhao-stripe (approved):

PR #2080, opened 2025-08-29
Fix config plugin INVALID_PLUGIN_IMPORT error

### LINKED ISSUE #1915: INVALID_PLUGIN_IMPORT on EAS build
**Describe the bug**
I get this error when creating a build via Expo EAS. I am on the newest version of Expo (`53.0.7`) and @stripe/stripe-react-native (`0.45.0`). I have installed the library as indicated in the docs https://github.com/stripe/stripe-react-native?tab=readme-ov-file#expo

```
 PluginError: Package "@stripe/stripe-react-native" does not contain a valid
    config plugin.
    Learn more: https://docs.expo.dev/guides/config-plugins/#creating-a-plugin

    Cannot use import statement outside a module
    Code: INVALID_PLUGIN_IMPORT
```

janicduplessis:
## Summary

Some users have been getting a INVALID_PLUGIN_IMPORT error when building on expo, the error also happens when using the expo vscode extension.

Fixes #1915

This also updates the version of `react-native-builder-bob` as I wanted to see if the latest version also creates those package.json files, but it does. I kept the version upgrade, but can revert if you prefer as it doesn't fix the issue.

## Motivation

`react-native-builder-bob` generates package.json files under the lib folder, which seems to confuse how expo resolves the config plugin. I went looking at other libraries that have an expo plugin and build using bob and found this https://github.com/zoontek/react-native-bootsplash/blob/master/package.json#L50. 

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "test": "jest",
     "typescript": "tsc --noEmit",
     "lint": "eslint \"**/*.{js,ts,tsx}\" --ignore-pattern \"docs/api-reference/*\" --ignore-path .gitignore",
-    "prepare": "bob build && husky",
+    "prepare": "bob build && rm -rf lib/*/package.json && husky",
     "release": "./scripts/publish",
     "example": "yarn --cwd example",
     "pods": "cd example && npx pod-install --quiet",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -130,12 +130,7 @@
           "configFile": true
         }
       ],
-      [
```
janicduplessis, line 133: The config validation is stricter now and the configFile option doesn't exist for typescript.

samer-stripe (approved):

Issue #1938, opened 2025-05-20
In Android app crash in release mode


akshSekhon:
**Describe the bug**

Android app is crash while initialising stripe

**To Reproduce**
  useEffect(() => {
        initializeStripe()
    }, [])

export const initializeStripe = async () => {
    const publishableKey = "km2Vp...."
    if (!publishableKey) {
        console.warn('Stripe key is missing!');
        return;
    }
    try {
        const resStripeInit = await initStripe({
            publishableKey,
            merchantIdentifier: 'merchant.identifier',
        });
        console.log('resStripeInit resStripeInit:-- ', resStripeInit);
    } catch (error) {
        console.log('error initializeStripe raised', error);
    }
}
**Expected behavior**

**Smartphone (please complete the following information):**
 - Device: Android 11

**Additional context**
 "react-native": "0.78.1",
  "@stripe/stripe-react-native": "^0.45.0", also on  "^0.46.0"

porter-stripe:
@akshSekhon can you provide the full stack track and the compose version you are using?

joacub:
> [@akshSekhon](https://github.com/akshSekhon) can you provide the full stack track and the compose version you are using?

This issue occurs due to a JSON parsing error combined with the new Android task manager behavior on many devices. Additionally, handleNextAction is being called with 3 arguments, but it expects 4

tjclawson-stripe:
Hey @akshSekhon can you provide your kotlin version as well?

sfriedman-stripe:
Hi @akshSekhon, just following up again here. Could you please share the full stack trace, Compose version, and Kotlin version?

akshSekhon:
@sfriedman-stripe @tjclawson-stripe  @porter-stripe  Sorry for late response please check below information

**Android Configrations**

 ext {
       buildToolsVersion = "35.0.0"
       minSdkVersion = 24
       compileSdkVersion = 34
       targetSdkVersion = 34
       ndkVersion = "27.1.12297006"
       kotlinVersion = "2.0.21"
    }

  "react-native": "0.78.1",
  "@stripe/stripe-react-native": "^0.47.1",


**System Config**

System:
  OS: macOS 15.3.1
  CPU: (6) x64 Intel(R) Core(TM) i5-8500B CPU @ 3.00GHz
  Memory: 62.96 MB / 8.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 23.10.0
[...]

Issue #1935, opened 2025-05-14
Android Release mode crash when calling initPaymentSheet


PierreCapo:
**Describe the bug**

**To Reproduce**
1. Build on Android Release Mode
2. Call the function `initPaymentSheet`
3. It should crashes the app

**Expected behavior**
App should not crash

**Smartphone (please complete the following information):**
 - Device: Samsung Galaxy A13
 - OS: Android 14

**Additional context**
- I have a bare react-native app (some expo libraries are used).
- iOS is perfectly fine.
- Android works well in debug build on a real device, but crashes in release mode.
- I am using:
```
react-native: "0.79.2"
@stripe/stripe-react-native: "0.46.0"
```
I also tried with `@stripe/stripe-react-native: "0.45.0"`, same crash 
I could circumvent the issue by using `@stripe/stripe-react-native: "0.43.0"` ✅

- I see this error when using logcat:
`android Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR)`

PierreCapo:
Looks like "@stripe/stripe-react-native: "0.43.0" is not compatible with Android new arch though

porter-stripe:
@PierreCapo can you provide the full stack track and the compose version you are using?

tjclawson-stripe:
@PierreCapo can you also provide your kotlin version

PierreCapo:
@tjclawson-stripe kotlinVersion used is "2.0.21".

@porter-stripe excuse-me I don't understand your question:
- Given I am using react-native, I haven't registered anywhere in my project any jetpack compose version (I am not super used to Android native dev, so let me know if I am missing something):

This is my `android/build.gradle`:

```
// Top-level build file where you can add configuration options common to all sub-projects/modules.

buildscript {
    ext {
        buildToolsVersion = "35.0.0"
        minSdkVersion = 24
        compileSdkVersion = 35
        targetSdkVersion = 35
        ndkVersion = "27.1.12297006"
        kotlinVersion = "2.0.21"
    }
    repositories {
        google()
        mavenCentral()

        maven { url 'https://maven.google.com' } // required by react-native-image-crop-picker
        maven { url "https://www.jitpack.io" } // required by react-native-image-crop-picker
    }
    
    dependencies {
        classpath("com.android.tools.build:gradle")
        classpath("com.facebook.react:react-native-gradle-plugin")
        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin")
[...]

Issue #2027, opened 2025-07-28
UnsupportedModulePropertyParserError: Module NativeStripeSdkModule: TypeScript interfaces extending TurboModule must only contain 'FunctionTypeAnnotation's or non nullable 'EventEmitter's. Property 'onConfirmHandlerCallback' refers to a 'TSTypeReference'.


erjondev:
**Describe the bug**
Pod installation fails with this message : 

`UnsupportedModulePropertyParserError: Module NativeStripeSdkModule: TypeScript interfaces extending TurboModule must only contain 'FunctionTypeAnnotation's or non nullable 'EventEmitter's. Property 'onConfirmHandlerCallback' refers to a 'TSTypeReference'.
    at throwIfModuleTypeIsUnsupported (/Users/erjon/Devs/mobile-app-upgrade/node_modules/@react-native/codegen/lib/parsers/error-utils.js:205:11)
    at buildPropertySchema (/Users/erjon/Devs/mobile-app-upgrade/node_modules/@react-native/codegen/lib/parsers/parsers-commons.js:523:3)
    at /Users/erjon/Devs/mobile-app-upgrade/node_modules/@react-native/codegen/lib/parsers/parsers-commons.js:856:22
    at guard (/Users/erjon/Devs/mobile-app-upgrade/node_modules/@react-native/codegen/lib/parsers/utils.js:31:14)
    at /Users/erjon/Devs/mobile-app-upgrade/node_modules/@react-native/codegen/lib/parsers/parsers-commons.js:836:14
    at Array.map (<anonymous>)
    at buildModuleSchema (/Users/erjon/Devs/mobile-app-upgrade/node_modules/@react-native/codegen/lib/parsers/parsers-commons.js:818:6)
    at /Users/erjon/Devs/mobile-app-upgrade/node_modules/@react-native/codegen/lib/parsers/parsers-commons.js:633:9
    at guard (/Users/erjon/Devs/mobile-app-upgrade/node_modules/@react-native/codegen/lib/parsers/utils.js:31:14)
    at buildSchemaFromConfigType (/Users/erjon/Devs/mobile-app-upgrade/node_modules/@react-native/codegen/lib/parsers/parsers-commons.js:632:22) {
  nodes: [ undefined ]
}`

**To Reproduce**
1. install stripe-react-native
2. pod install

**Expected behavior**
Pod install complete

cttsai-stripe:
[...]

PR #2092, opened 2025-09-05
Use compat layer events on new arch

janicduplessis:
## Summary

This removes codegen EventEmitters since it causes some issues with older RN versions. For now we will use the compat layer that we use for old arch instead on both new and old arch.

## Motivation

Fixes https://github.com/stripe/stripe-react-native/issues/1935
Fixes https://github.com/stripe/stripe-react-native/issues/1938

Better fix for https://github.com/stripe/stripe-react-native/issues/2027

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 


porter-stripe:
Thanks! Could we get the CHANGELOG updated with this fix?

PR #2106, opened 2025-09-10
Crypto Onramp Example Cleanup

Twigz:
## Summary
* Removes VerifyCryptoLinkUserScreen (this functionality already exists in the flow)
* Moves UI to a subfolder specific to onramp
* Separates UI from the main CryptoOnrampFlow to subsections that are easier to follow.
* Separates some logic/styling into separate files to be added to later.

## Motivation
This is a step towards making the RN Example for CryptoOnramp a bit clearer and easier to follow, rather than having a singular file that contains all of the code. This PR is primarily geared towards the UI.

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 


example/src/screens/Onramp/styles.ts
```diff
@@ -0,0 +1,48 @@
+import { StyleSheet } from 'react-native';
+import { colors } from '../../colors';
+
+export const styles = StyleSheet.create({
```
lng-stripe: I think having some global generic styles is OK, but a component's style should be private and declared in the same file as the component.

Twigz: For RN, this isn't an uncommon way of doing things to have a `styles.ts` inside a particular section of components (Onramp) delineated by folder like this, **but** there was not as much being used here as I thought by various pieces, so I've put it back in the CryptoOnrampScreen, and removed some things that were unused.

PR #2113, opened 2025-09-11
fix: add missing await to NativeStripeSdk.initialise calls

mmv08:
## Summary
The initialise method is async and returns a Promise, but it wasn't being awaited. 

Changes:
- Added await to initStripe function call to NativeStripeSdk.initialise
- Wrapped useEffect initialization in async IIFE 
- Both Android and iOS initialization paths now properly await completion

## Motivation
This could cause race conditions where subsequent Stripe operations might be called before initialization completes.

## Testing

- [ ] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/src/components/StripeProvider.tsx
+++ b/src/components/StripeProvider.tsx
@@ -45,7 +45,7 @@ export const initStripe = async (params: InitStripeParams): Promise<void> => {
   }
 
   const extendedParams: InitialiseParams = { ...params, appInfo };
-  NativeStripeSdk.initialise(extendedParams);
+  await NativeStripeSdk.initialise(extendedParams);
 };
 
 /**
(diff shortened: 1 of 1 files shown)


imjakechapman:
@janicduplessis Does this mean there's no way initialization could error? That once initStripe is called, we'll always be initialized successfully? Only reason I'm asking is I'm working on a small refactor right now and was looking how the initialization works and appears registering `StripeKeepJsAwakeTask` is the only thing really happening for Android and can't see internals of `NativeStripeSdk.initialise(extendedParams)` so guessing that's not doing anything that can actually throw?

janicduplessis:
Yes, currently there is nothing that can throw in the initialization code

PR #2118, opened 2025-09-12
Embedded invisible Android fix

porter-stripe:
## Summary
Fixes intermittent invisibility of EmbeddedPaymentElementView by correcting height measurement and state updates. Previously, the component stored raw pixel heights as dp and mutated state inside layout {}, leading to zero-height frames and invisible but still tappable UI. This change introduces a custom measure path that:

- Calculates child height in px during measure and lays out accordingly.
- Updates requiredHeight in dp via onSizeChanged instead of mutating state in measure.
- Ensures a non-zero initial height to prevent disappearing frames.

## Motivation
- Rendering issues for embedded on React Native

## Testing

- [x] I tested this manually
- [ ] I added automated tests

Verified it fixed the issue with the user

## Documentation

Select one: 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 **Fixes**
 - Fixed Android EmbeddedPaymentElement not respecting Link configuration. Now properly honors `LinkDisplay.NEVER` to hide Link payment method when configured.
+- Fixed Android EmbeddedPaymentElement layout measurement issues by improving height calculation and switching from `onPlaced` to `onSizeChanged` for better layout synchronization.
 
 ## 0.52.0 - 2025-09-08
 **Features**


android/src/main/java/com/reactnativestripesdk/EmbeddedPaymentElementView.kt
```diff
@@ -287,33 +287,56 @@ class EmbeddedPaymentElementView(
...
+      measuredEmbeddedElement(
+        reportHeightChange = { h -> reportHeightChange(h) },
+      ) {
+        embedded.Content()
+      }
+    }
+  }
+
+  @Composable
+  private fun measuredEmbeddedElement(
```
samer-stripe, line 300: Nit: `Composable` methods should start with a capital letter, can follow-up since this fix is urgent

android/src/main/java/com/reactnativestripesdk/EmbeddedPaymentElementView.kt
```diff
@@ -287,33 +287,56 @@ class EmbeddedPaymentElementView(
...
+    reportHeightChange: (Float) -> Unit,
+    content: @Composable () -> Unit,
+  ) {
```
samer-stripe, line 303: Nit: `Composable` parameters should start with a capital letter, can follow-up since this fix is urgent

android/src/main/java/com/reactnativestripesdk/EmbeddedPaymentElementView.kt
porter-stripe: Our linter actually fails if I do that :(

PR #2128, opened 2025-09-16
Set up headless task when using StripeProvider

janicduplessis:
## Summary

When initializing the app via `StripeProvider` we never end up registering the headless task. To fix it I changed the Provider to call `initStripe` so we now have only one initialization code path. I also added a check to make sure the task only gets registered once, as I assume it could happen that multiple `StripeProvider` are rendered.

## Motivation

Task should be registered when using `StripeProvider`

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 


src/components/StripeProvider.tsx
```diff
@@ -83,18 +86,16 @@ export function StripeProvider({
...
           stripeAccountId,
           threeDSecureParams,
           urlScheme,
           setReturnUrlSchemeOnAndroid,
         });
       } else {
-        await NativeStripeSdk.initialise({
+        await initStripe({
           publishableKey,
-          appInfo,
```
janicduplessis, line 97: App info is added by `initStripe` already.

PR #2133, opened 2025-09-18
Fix a potential bug in the stop method.

jaynewstrom-stripe:
## Summary

A user wrote in suggesting a fix for this, it doesn't seem harmful, so I'm adding it in.

## Motivation

https://jira.corp.stripe.com/browse/RUN_MOBILESDK-4693

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/utils/KeepJsAwakeTask.kt
+++ b/android/src/main/java/com/reactnativestripesdk/utils/KeepJsAwakeTask.kt
@@ -33,7 +33,14 @@ internal class KeepJsAwakeTask(
   fun stop() {
     val taskId = taskId ?: return
     val headlessJsTaskContext = HeadlessJsTaskContext.getInstance(context)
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/reactnativestripesdk/utils/KeepJsAwakeTask.kt
```diff
@@ -33,7 +33,14 @@ internal class KeepJsAwakeTask(
...
     val taskId = taskId ?: return
     val headlessJsTaskContext = HeadlessJsTaskContext.getInstance(context)
-    headlessJsTaskContext.finishTask(taskId)
-    this.taskId = null
+
+    try {
+      headlessJsTaskContext.finishTask(taskId)
+    } catch (e: AssertionError) {
+      // Ignore if task already finished
+      // Log.w("KeepJsAwakeTask", "Tried to stop a non-existent task (id=$taskId)")
```
janicduplessis: Shoud we uncomment this?

PR #2147, opened 2025-09-27
Fix stripe ui errors when rotating screen

janicduplessis:
## Summary

The issue was related to the code we have to dismiss stripe UI when the react-native activity is re-created. The problem is that stripe activities are re-created on screen rotation, while the react-native one is not, leading to the error about not being able to retrieve result.

To fix this I changed the approach a little bit to make it more reliable and make sure it is only triggered when the react-native activity is re-created, and not any activity.

The idea is to set the flag only if activity is a ReactActivity, and reset that flag when stripe is initialized again. This is safe since no stripe UI can be presented before the module is initialized.

I also saw issues where the lifecycle callback was being called more and more since we never unregistered lifecycle listeners.

## Motivation

Fix error on screen rotation.

## Testing

- [x] I tested this manually
- [ ] I added automated tests

- Payment sheet works fine on screen rotation.
- Make sure that we don't regress on activity re-creation behaviour from https://github.com/stripe/stripe-react-native/pull/1900. Make sure that the app doesn't crash when backgrounded with "Don't keep Activities" developer setting and that any stripe ui is dismissed.

https://github.com/user-attachments/assets/671c5b4d-81e1-4b32-a623-a5350e5b5a4d

## Documentation

Select one: 


tjclawson-stripe:
Tested the 3DS2 flow and this resolves the error, thanks!

PR #2166, opened 2025-10-14
Add ConfirmationTokens API

porter-stripe:
## Summary
- Adds ConfirmationToken APIs for Embedded and PaymentSheet
- Updates our deferred and embedded tests screens to use confirmation tokens, so they are now covered by tests.

## Motivation
- ConfirmationTokens
- https://docs.google.com/document/d/1zL6KkOaOM2vVqJ3PBM_o2qHSNej1B7Dhx09Fu0y1jCs/edit?tab=t.0#heading=h.570u4leviueu

## Testing

- [x] I tested this manually
- [x] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/EmbeddedPaymentElementView.kt
+++ b/android/src/main/java/com/reactnativestripesdk/EmbeddedPaymentElementView.kt
@@ -217,10 +217,9 @@ class EmbeddedPaymentElementView(
 
                 stripeSdkModule.eventEmitter.emitOnConfirmationTokenHandlerCallback(params)
 
-                // Save reference and reset BEFORE awaiting to prevent race conditions
-                val callback = stripeSdkModule.embeddedConfirmationTokenCreationCallback
+                val resultFromJavascript = stripeSdkModule.embeddedConfirmationTokenCreationCallback.await()
+                // reset the completable
                 stripeSdkModule.embeddedConfirmationTokenCreationCallback = CompletableDeferred()
-                val resultFromJavascript = callback.await()
 
                 keepJsAwakeTask.stop()
 
--- a/android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt
+++ b/android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt
@@ -231,10 +231,9 @@ class PaymentSheetFragment :
 
         stripeSdkModule?.eventEmitter?.emitOnConfirmationTokenHandlerCallback(params)
 
-        // Save reference and reset BEFORE awaiting to prevent race conditions
-        val callback = paymentSheetConfirmationTokenCreationCallback
+        val resultFromJavascript = paymentSheetConfirmationTokenCreationCallback.await()
+        // reset the completable
         paymentSheetConfirmationTokenCreationCallback = CompletableDeferred<ReadableMap>()
-        val resultFromJavascript = callback.await()
 
         return@CreateIntentWithConfirmationTokenCallback resultFromJavascript.getString("clientSecret")?.let {
           CreateIntentResult.Success(clientSecret = it)


android/gradle.properties
```diff
@@ -3,4 +3,4 @@ StripeSdk_compileSdkVersion=30
 StripeSdk_targetSdkVersion=28
 StripeSdk_minSdkVersion=21
 # Keep StripeSdk_stripeVersion in sync with https://github.com/stripe/stripe-identity-react-native/blob/main/android/gradle.properties
-StripeSdk_stripeVersion=21.26.+
+StripeSdk_stripeVersion=21.28.0
```
cttsai-stripe: I think this should be 21.29.0, if we need the code complete version.

android/src/main/java/com/reactnativestripesdk/utils/Mappers.kt
```diff
@@ -1116,3 +1116,67 @@ internal fun mapFromCustomPaymentMethod(
...
+  token.putString("setupFutureUsage", mapFromSetupFutureUsage(confirmationToken.setupFutureUsage))
+
+  // PaymentMethodPreview
+  confirmationToken.paymentMethodPreview?.let { preview ->
+    val paymentMethodPreview = WritableNativeMap()
+    paymentMethodPreview.putString("type", mapPaymentMethodType(preview.type))
+    paymentMethodPreview.putMap("billingDetails", mapFromBillingDetails(preview.billingDetails))
+    paymentMethodPreview.putString("allowRedisplay", mapFromAllowRedisplay(preview.allowRedisplay))
+    paymentMethodPreview.putString("customerId", preview.customerId)
+    paymentMethodPreview.putMap("allResponseFields", preview.allResponseFields?.toReadableMap() ?: WritableNativeMap())
```
cttsai-stripe: Curious if it's ok to be exposed in RN? I think we keep this field internal/restricted on Android per [API review discussion](https://docs.google.com/document/d/1zL6KkOaOM2vVqJ3PBM_o2qHSNej1B7Dhx09Fu0y1jCs/edit?disco=AAABrLaUbR4). cc. @jaynewstrom-stripe

android/src/main/java/com/reactnativestripesdk/utils/Mappers.kt
```diff
@@ -1116,3 +1116,67 @@ internal fun mapFromCustomPaymentMethod(
...
+    } ?: run {
+      shipping.putMap("address", WritableNativeMap())
+    }
+
+    token.putMap("shipping", shipping)
+  } ?: run {
+    token.putNull("shipping")
+  }
+
+  token.putMap("allResponseFields", confirmationToken.allResponseFields?.toReadableMap() ?: WritableNativeMap())
```
cttsai-stripe: Same for this field

android/src/main/java/com/reactnativestripesdk/utils/Mappers.kt
jaynewstrom-stripe: I think we should omit it, at least until there's a use case that someone asks for. It's much easier to add a field than it is to remove one.

android/src/main/java/com/reactnativestripesdk/utils/Mappers.kt
```diff
@@ -1118,3 +1120,65 @@ internal fun mapFromCustomPaymentMethod(
...
+  } ?: run {
+    token.putNull("shipping")
+  }
+
+  return token
+}
+
+private fun mapFromSetupFutureUsage(setupFutureUsage: Any?): String? {
+  // This will need to be implemented based on the actual SetupFutureUsage enum when available
+  return setupFutureUsage?.toString()
```
cttsai-stripe: **🚨 CRITICAL:** Incorrect enum mapping. Using `.toString()` will return "OnSession" instead of "on_session".

**Fix:**
```kotlin
private fun mapFromSetupFutureUsage(setupFutureUsage: ConfirmPaymentIntentParams.SetupFutureUsage?): String? {
  return setupFutureUsage?.code  // Use code property, not toString()
}
```

The Android SDK `SetupFutureUsage` enum has an internal `code` property that contains the correct API values ("on_session", "off_session").

android/src/main/java/com/reactnativestripesdk/utils/Mappers.kt
```diff
@@ -1118,3 +1120,65 @@ internal fun mapFromCustomPaymentMethod(
...
+}
+
+private fun mapFromAllowRedisplay(allowRedisplay: Any?): String? {
+  // This will need to be implemented based on the actual AllowRedisplay enum when available
+  return allowRedisplay?.toString()
```
cttsai-stripe: **🚨 CRITICAL:** Incorrect enum mapping. Using `.toString()` will return "UNSPECIFIED" instead of "unspecified".

**Fix:**
```kotlin
@SuppressLint("RestrictedApi")
private fun mapFromAllowRedisplay(allowRedisplay: PaymentMethod.AllowRedisplay?): String? {
  return allowRedisplay?.value  // Use value property, not toString()
}
```

The Android SDK `AllowRedisplay` enum has a `value` property that contains the correct API values ("unspecified", "limited", "always").

android/src/main/java/com/reactnativestripesdk/EmbeddedPaymentElementView.kt
```diff
@@ -153,73 +161,128 @@ class EmbeddedPaymentElementView(
...
-                      putString("status", "canceled")
-                    }
-
-                    is EmbeddedPaymentElement.Result.Failed -> {
-                      putString("status", "failed")
-                      putString("error", result.error.message ?: "Unknown error")
-                    }
+                stripeSdkModule.eventEmitter.emitOnConfirmationTokenHandlerCallback(params)
+
+                val resultFromJavascript = stripeSdkModule.embeddedConfirmationTokenCreationCallback.await()
```
cttsai-stripe, line 220: **⚠️ HIGH PRIORITY:** Threading race condition. The callback reset happens AFTER awaiting, which means another confirmation could start between `await()` and the reset, causing the new callback to be lost.

**Fix:**
```kotlin
// Reset BEFORE awaiting to prevent race condition
val callback = stripeSdkModule.embeddedConfirmationTokenCreationCallback
stripeSdkModule.embeddedConfirmationTokenCreationCallback = CompletableDeferred()
val resultFromJavascript = callback.await()

keepJsAwakeTask.stop()
```

android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt
```diff
@@ -212,6 +221,32 @@ class PaymentSheetFragment :
...
+      CreateIntentWithConfirmationTokenCallback { confirmationToken ->
+        val stripeSdkModule: StripeSdkModule? = context.getNativeModule(StripeSdkModule::class.java)
+        val params =
+          Arguments.createMap().apply {
+            putMap("confirmationToken", mapFromConfirmationToken(confirmationToken))
+          }
+
+        stripeSdkModule?.eventEmitter?.emitOnConfirmationTokenHandlerCallback(params)
+
+        val resultFromJavascript = paymentSheetConfirmationTokenCreationCallback.await()
```
cttsai-stripe, line 234: **⚠️ HIGH PRIORITY:** Same threading race condition as in EmbeddedPaymentElementView. Reset the callback BEFORE awaiting.

**Fix:**
```kotlin
val callback = paymentSheetConfirmationTokenCreationCallback
paymentSheetConfirmationTokenCreationCallback = CompletableDeferred<ReadableMap>()
val resultFromJavascript = callback.await()
```

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
```diff
@@ -323,6 +325,22 @@ class StripeSdkModule(
     promise?.resolve(null)
   }
 
+  @ReactMethod
+  override fun confirmationTokenCreationCallback(
+    params: ReadableMap,
+    promise: Promise,
+  ) {
+    embeddedConfirmationTokenCreationCallback.complete(params)
```
cttsai-stripe, line 333: **⚠️ HIGH PRIORITY:** This method completes ALL callbacks regardless of which component (Embedded or PaymentSheet) initiated the request. This could cause:
1. `IllegalStateException` if a callback is already completed
2. Wrong component receiving results meant for another

**Fix:**
```kotlin
@ReactMethod
override fun confirmationTokenCreationCallback(params: ReadableMap, promise: Promise) {
  // Only complete if not already completed
  if (!embeddedConfirmationTokenCreationCallback.isCompleted) {
    embeddedConfirmationTokenCreationCallback.complete(params)
  }
  if (!paymentSheetConfirmationTokenCreationCallback.isCompleted) {
    paymentSheetConfirmationTokenCreationCallback.complete(params)
  }
  paymentSheetFragment?.paymentSheetConfirmationTokenCreationCallback?.let { callback ->
    if (!callback.isCompleted) {
      callback.complete(params)
    }
  }
  promise.resolve(null)
}
```

porter-stripe: This is ok, embedded and payment sheet are never used side-by-side.

cttsai-stripe:
I reviewed the PR, and also asked claude to do another pass. I think the ones marked "critical" are real issues. For those marked "HIGH PRIORITY", I am not sure, and feel free to ignore them if they are noise.

cttsai-stripe (dismissed):

android/src/main/java/com/reactnativestripesdk/PaymentSheetFragment.kt
porter-stripe: Had to revert the sgugestion, it broke the payment flow. I tink this race condition isn't very realistic.

PR #2174, opened 2025-10-18
Remove fragments

janicduplessis:
## Summary

Use new apis from stripe android that do not require being initialized while the activity is being created. This allows removing the fragments hack that we’re currently using.

This introduces a new base class StripeUIManager that is used instead of the fragment and manages the lifecycle of the stripe ui. These can just be instantiated normally, passing options as constructor arguments, leaving the previous requirement of fragments having no argument constructors. This also avoid having to use Bundle and can use ReadableMap directly.

This also does a bit of cleanup in the parsing code to use extension methods instead of adhoc helpers methods we had.

## Motivation

Remove fragment usage

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 **Changes**
 - CustomerSession is now generally available. The `customerSessionClientSecret` parameter is no longer experimental.
 - ConfirmationTokens are now generally available. The `confirmationTokenConfirmHandler` parameter and `ConfirmationToken.Result` type are no longer experimental.
+- [Added] Added support for `CustomerSession` to CustomerSheet. The CustomerSession object grants the SDK temporary access to the Customer and provides additional configuration options. These configuration options allow you to customize the behavior of CustomerSheet. A complete list of features exposed on the CustomerSession are [in our API docs](https://docs.corp.stripe.com/api/customer_sessions/create#create_customer_session-components-customer_sheet).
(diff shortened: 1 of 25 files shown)


porter-stripe (commented):
changes look good to me, @samer-stripe any thoughts? Some of this is a bit out of my knowledge realm

android/src/main/java/com/reactnativestripesdk/EmbeddedPaymentElementViewManager.kt
```diff
@@ -134,20 +133,20 @@ class EmbeddedPaymentElementViewManager :
...
-        buildCustomerConfiguration(toBundleObject(map))
+        buildCustomerConfiguration(map)
       } catch (error: PaymentSheetException) {
         throw Error() // TODO handle error
       }
 
-    val googlePayConfig = buildGooglePayConfig(toBundleObject(map.getMap("googlePay")))
-    val linkConfig = PaymentSheetFragment.buildLinkConfig(toBundleObject(map.getMap("link")))
+    val googlePayConfig = buildGooglePayConfig(map.getMap("googlePay"))
+    val linkConfig = PaymentSheetManager.buildLinkConfig(map.getMap("link"))
```
samer-stripe, line 142: As a follow-up, we should move shared elements between Embedded & Payment Sheet to `PaymentElementConfigExtensions` file.

porter-stripe:
@janicduplessis I can stamp this once the conflicts are resolved! :)

samer-stripe (dismissed):

janicduplessis:
Do you know if the changes in https://github.com/stripe/stripe-react-native/commit/8aca5f2b078546e7431472c56c83a0677d9f1a1e is covered by e2e tests, the merge was not trivial so want to make sure I didn't break anything if manual testing is needed.

porter-stripe:
@janicduplessis will check with the author and double check it still works

PR #2180, opened 2025-10-22
Update SDKs for Identity

Twigz:
## Summary
Updated iOS/Android SDKs, mainly to support the new Identity changes made by the Identity team.

This update in particular handles icon display during Identity Verification, as well as auto-dismissal of verification completion.

## Motivation
Changes to the Identity framework were necessary to support some Crypto Onramp features.

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

## Screenshots
| Android | iOS |
| --- | --- |
|  |  |

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -1685,12 +1685,12 @@ PODS:
     - ReactCommon/turbomodule/core
     - Yoga
   - SocketRocket (0.7.1)
-  - Stripe (24.24.1):
-    - StripeApplePay (= 24.24.1)
-    - StripeCore (= 24.24.1)
-    - StripePayments (= 24.24.1)
-    - StripePaymentsUI (= 24.24.1)
-    - StripeUICore (= 24.24.1)
+  - Stripe (24.24.4):
+    - StripeApplePay (= 24.24.4)
+    - StripeCore (= 24.24.4)
+    - StripePayments (= 24.24.4)
+    - StripePaymentsUI (= 24.24.4)
+    - StripeUICore (= 24.24.4)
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/reactnativestripesdk/utils/Mappers.kt
```diff
@@ -594,6 +594,12 @@ internal fun mapNextAction(
         nextActionMap.putString("qrCodeUrl", it.qrCodeUrl)
       }
     }
+    NextActionType.DisplayPromptPayDetails -> {
```
lng-stripe, line 597: Curious why iOS doesn't need a similar change?

Twigz: That, is a good question. The answer appears to be, it probably should, but hasn't kept up to date for a while. The reason it's not flagged is because it uses a `default` case in Swift.

```swift
default: // .useStripeSDK, .BLIKAuthorize, .unknown
    return nil
}
```

Currently the only implemented cases are:
```swift
verifyWithMicrodeposits
redirectToURL
weChatPayRedirectToApp
alipayHandleRedirect
OXXODisplayDetails
boletoDisplayDetails
konbiniDisplayDetails
```

Based on comparing the available iOS cases to implemented Android cases, likely the following need to be added for functionality
```swift
    case payNowDisplayQrCode
    case promptpayDisplayQrCode
    case swishHandleRedirect
    case multibancoDisplayDetails
```

with a few others being defined but being a no-op (`upiAwaitNotification` and `cashAppRedirectToApp`)

I can handle that as part of this PR if what I've explained is correct that this is an oversight due to a default switch case. Let me know @lng-stripe / @mats-stripe.

PR #2087, opened 2025-09-04
Add support for Account Onboarding

gimenete-stripe:
## Summary

This PR adds support for the Account Onboarding embedded component.

## Motivation

We need it for the new React Native dashboard application.

## Testing

- [ ] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 
- [ ] I have added relevant documentation for my changes.
- [ ] This PR does not result in any developer-facing changes.

PR #2187, opened 2025-10-27
Fix account onboarding from PR 2087

maxvaljan:
Follow-up to #2087 addressing Guido's review comments by removing the unused AccountOnboarding client secret wiring across TS, iOS, Android, and the example app.

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/StripeSdkContentProvider.kt
+++ b/android/src/main/java/com/reactnativestripesdk/StripeSdkContentProvider.kt
@@ -7,35 +7,28 @@ import android.content.ContentValues
 import android.database.Cursor
 import android.net.Uri
 import android.os.Bundle
-import android.util.Log
 import androidx.activity.ComponentActivity
 import com.stripe.android.connect.EmbeddedComponentManager
 
 class StripeSdkContentProvider : ContentProvider() {
   override fun onCreate(): Boolean {
-    Log.d("StripeReactNative", "StripeSdkContentProvider $context")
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/reactnativestripesdk/StripeSdkContentProvider.kt
maragues-stripe, line 1: Let's cleanup logs in this file. I don't think they are valuable.

maragues-stripe, line 1: Let's cleanup all logs in this file, and the `log` function. I don't think they are valuable.

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
maragues-stripe, line 1: Same. Please cleanup logs, unless @gimenete-stripe wants to keep some of them.

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
```diff
@@ -277,6 +282,36 @@ class StripeSdkModule(
     }
   }
 
+  @ReactMethod
+  override fun presentAccountOnboardingScreen(
+    options: ReadableMap,
+    promise: Promise,
+  ) {
+    getCurrentActivityOrResolveWithError(promise)?.let { activity ->
+      UiThreadUtil.runOnUiThread {
```
maragues-stripe: Does this whole block need to run on UI thread? Or maybe only `accountOnboardingController.show()`?

Issue #2173, opened 2025-10-18
Android: Affirm is not showing - Is Affirm not available in Android?


takosunil:
**Describe the bug**
Affirm payment type is not showing in android.

**To Reproduce**
Using latest version 0.54.1

**Expected behavior**
Affirm payment type should display in PaymentSheet same as in iOS.
<img width="346" height="745" alt="Image" src="https://github.com/user-attachments/assets/144b19b2-d20b-4ca3-af80-2127a8b21df4" />

**Screenshots: Android**
<img width="278" height="596" alt="Image" src="https://github.com/user-attachments/assets/2aefd945-1b26-40df-8a32-c52d2b5c8e9a" />

**Question**
Is Affirm not available in Android?

davidme-stripe:
Affirm is supported on Android, so I'm not sure why it isn't showing up in the second screenshot. Can you share a `req_` request ID or PaymentIntent ID for us to investigate? You could also reach out to [Stripe support](https://support.stripe.com/), they can help you debug the integration.

suniltakorugby:
@davidme-stripe Here is the `PaymentIntent ID: pi_3SKoHiGiWn6VaB4z0powWqaz`

Here is the payment flow that I have:
1. Frontend - Payment Screen:

 const result = await fetchRequest(() => createPaymentIntent(data)); // API call to create PaymentIntent

  const clientSecret = result?.paymentIntent?.client_secret;
  if (!clientSecret) {
    console.warn('No client_secret returned from PaymentIntent');
    return;
  }

  const { error } = await initPaymentSheet({
    paymentIntentClientSecret: clientSecret,
    merchantDisplayName: 'Test',
    style: 'automatic', // automatic dark/light mode
    paymentMethodLayout: PaymentMethodLayout.Horizontal, // layout style
    customFlow: true,
    returnURL: 'test-app://registration/payment-details', // your app's return URL
  });
  
  if (error) {
    console.log('initialisePaymentSheet ~ error:', error);
  } else {
    console.log('PaymentSheet initialized successfully');
  }

2. Create payment intent from backend. Only adding 'affirm' as payment method types.
`const paymentIntent = await stripe.paymentIntents.create({payment_method_types: ['affirm'], ...paymentIntentReq}); `

3. When I tried to show the payment sheet. I'm getting this error only in Android (No error in iOS):
const { error, didCancel, paymentOption } = await presentPaymentSheet();
[...]

PR #2191, opened 2025-10-30
Fix patch releases not automatically being used from `stripe-android`

samer-stripe:
## Summary
Fix patch releases not automatically being used from `stripe-android`

## Motivation
We should be automatically pulling in patch releases for `stripe-android`. Was accidentally removed during confirmation tokens implementation for React Native.

Resolves https://github.com/stripe/stripe-react-native/issues/2173 since 21.29.2 has removed the Affirm requirement.

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/example/src/screens/EmbeddedPaymentElementScreen.tsx
+++ b/example/src/screens/EmbeddedPaymentElementScreen.tsx
@@ -95,7 +95,7 @@ function PaymentElementView({ intentConfig, elementConfig }: any) {
 
       <Button
         variant="primary"
-        title="Pay"
+        title="Complete payment"
         onPress={handlePay}
         loading={loading}
         disabled={!paymentOption}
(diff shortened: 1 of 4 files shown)


example/src/screens/EmbeddedPaymentElementScreen.tsx
```diff
@@ -95,7 +95,7 @@ function PaymentElementView({ intentConfig, elementConfig }: any) {
 
       <Button
         variant="primary"
-        title="Pay"
+        title="Complete payment"
```
samer-stripe, line 98: I was noticing failures where Embedded CPM tests would fail to click "Pay". When I ran locally, I saw it would try to click any LPM that had "Pay" in it (Cash App Pay, BufoPay).

Thought maybe something unintentional changed between 21.29.0 and 21.29.2 but seems like nothing has (this after clean builds):
| Before | After |
| ---- | ----- |
| before | after |

samer-stripe: I updated these tests to use `Complete payment`

PR #2193, opened 2025-10-31
Embedded components

gimenete-stripe:
## Summary

This PR adds support for Connect embedded components. In particular it adds support for 3 components:

- Account onboarding
- Payments
- Payouts

New components can easily be added.

Most of the work is done using just React Native but I need to use native code for a couple of things:

- Handling "secure webviews"
- Rendering the navigation bar to match the look and feel of the system (e.g. liquid glass) when opening the onboarding embedded component, which needs to be shown in a modal.

Question: this API is not ready for GA yet. How do we prevent users to think it's production ready? Should I add an `unstable_` prefix or a `PrivatePreview` suffix to the API or something like that?

## Motivation

https://docs.google.com/document/d/1tEm_mtsbHgbyhBoctYGLjmmLth0nI4b24osTIxN8_gM/edit?usp=sharing

## Testing

- [x] I tested this manually
- [ ] I added automated tests

I updated the backend to have an additional endpoint to return a client secret for the embedded components and added 3 screens to the example app.

## Documentation

Select one: 


src/connect/EmbeddedComponent.tsx
```diff
@@ -0,0 +1,265 @@
+import type { LoadError, LoaderStart } from '@stripe/connect-js';
+import React, { useRef } from 'react';
+import { Platform, StyleProp, ViewStyle } from 'react-native';
+import { WebView } from 'react-native-webview';
+import { useConnectComponents } from './ConnectComponentsProvider';
+import pjson from '../../package.json';
+
+const DEVELOPMENT_MODE = true;
```
gimenete-stripe: Reminder: change to false

example/src/screens/ConnectScreen.tsx
```diff
@@ -0,0 +1,51 @@
...
+    >
+      {children}
+    </ConnectComponentsProvider>
+  );
+};
+
+export default ConnectScreen;
+
+const fetchClientSecret = async () => {
+  // TODO: use a different endpoint
```
gimenete-stripe: I think we need a new endpoint just for this in the demo project. Will investigate.

src/connect/EmbeddedComponent.tsx
```diff
@@ -0,0 +1,265 @@
...
+  const platform = Platform.OS.toLowerCase();
+  const sdkVersion = pjson.version;
+
+  const baseURL = DEVELOPMENT_MODE
+    ? Platform.select({
+        ios: 'http://localhost:3001',
+        android: 'http://10.0.2.2:3001',
+      })
+    : 'https://connect-js.stripe.com';
+  const connectURL = `${baseURL}/v1.0/${platform}_webview.html#${hash}`;
```
gimenete-stripe: TODO: what if the platform is not iOS/Android. Could also be windows/macos/web.

gimenete-stripe: We have a specific react native endpoint now.

example/src/screens/ConnectScreen.tsx
gimenete-stripe: See PR description.

android/src/main/java/com/reactnativestripesdk/NavigationBarManager.kt
```diff
@@ -0,0 +1,38 @@
...
+  NavigationBarManagerInterface<NavigationBarView> {
+  private val delegate = NavigationBarManagerDelegate(this)
+
+  override fun getName() = REACT_CLASS
+
+  override fun getDelegate() = delegate
+
+  override fun getExportedCustomDirectEventTypeConstants() =
+    mutableMapOf(
+      "onCloseButtonPress" to mutableMapOf("registrationName" to "onCloseButtonPress"),
```
maragues-stripe: `onCloseButtonPress` should be a constant.

android/src/main/java/com/reactnativestripesdk/NavigationBarView.kt
```diff
@@ -0,0 +1,119 @@
...
+import android.widget.TextView
+import androidx.appcompat.widget.Toolbar
+import com.facebook.react.bridge.Arguments
+import com.facebook.react.uimanager.ThemedReactContext
+import com.facebook.react.uimanager.UIManagerHelper
+import com.facebook.react.uimanager.events.Event
+
+@SuppressLint("ViewConstructor")
+class NavigationBarView(
+  private val context: ThemedReactContext,
```
maragues-stripe:
```suggestion
  context: ThemedReactContext,
```

android/src/main/java/com/reactnativestripesdk/NavigationBarView.kt
```diff
@@ -0,0 +1,119 @@
...
+    titleTextView.text = title
+  }
+
+  private fun dispatchCloseButtonPress() {
+    val event =
+      CloseButtonPressEvent(
+        context.surfaceId,
+        id,
+      )
+    UIManagerHelper.getEventDispatcherForReactTag(context, id)?.dispatchEvent(event)
```
maragues-stripe:
```suggestion
        getContext().surfaceId,
        id,
      )
    UIManagerHelper.getEventDispatcherForReactTag(getContext(), id)?.dispatchEvent(event)
```

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
```diff
@@ -1329,6 +1329,39 @@ class StripeSdkModule(
...
+        val uri = android.net.Uri.parse(url)
+        val builder =
+          androidx.browser.customtabs.CustomTabsIntent
+            .Builder()
+
+        // Set toolbar color for better UX
+        builder.setShowTitle(true)
+        builder.setUrlBarHidingEnabled(true)
+
+        val customTabsIntent = builder.build()
```
maragues-stripe, line 1356: nit: You can probably move this outside of `runOnUiThread`

example/android/build.gradle
```diff
@@ -42,3 +42,41 @@ allprojects {
...
+    subproject.afterEvaluate {
+        subproject.tasks.matching { task ->
+            task.name == 'generateAndroidManifest'
+        }.each { task ->
+            task.doLast {
+                def manifestFile = file("${subproject.buildDir}/generated/rnta/src/main/AndroidManifest.xml")
+                if (manifestFile.exists()) {
+                    def manifest = manifestFile.text
+
+                    // Add launchMode="singleTask"
```
maragues-stripe, line 57: MainActivity with launchMode=singleTask will destroy all activities that previously existed on top of it.

Is that what we want?

In any case, can't we define a standard AndroidManifest.xml file, instead of doing this?

example/android/build.gradle
```diff
@@ -42,3 +42,41 @@ allprojects {
         google()
     }
 }
+
+// Fix MainActivity launchMode and add deep link for stripe-connect
+subprojects { subproject ->
+    subproject.afterEvaluate {
+        subproject.tasks.matching { task ->
+            task.name == 'generateAndroidManifest'
+        }.each { task ->
```
maragues-stripe:
```suggestion
        }.configureEach { task ->
```

src/connect/Components.tsx
```diff
@@ -0,0 +1,207 @@
...
+  LoaderStart,
+  PaymentsListDefaultFilters,
+  StepChange,
+} from './connectTypes';
+import { CommonComponentProps, EmbeddedComponent } from './EmbeddedComponent';
+import { NavigationBar } from './NavigationBar';
+
+// Export NavigationBar for external use
+export { NavigationBar } from './NavigationBar';
+export type { NavigationBarProps } from './NavigationBar';
```
maragues-stripe, line 24: Are these needed? I've found that claude loves to export types like this, when it's actually not needed, since they are already exported by the other file.

src/connect/Components.tsx
```diff
@@ -0,0 +1,207 @@
...
+    return {
+      onExit: onExitCallback,
+      onStepChange,
+      onCloseWebView: onExitCallback,
+    };
+  }, [onExitCallback, onStepChange]);
+
+  const onLoaderStartCallback = useCallback(
+    (event: LoaderStart) => {
+      setLoading(false);
```
maragues-stripe, line 95: Should this be `setLoading(true)`?

I guess I'm wrong, since there's no other `setLoading(false)`. But the name is counter intuitive.

maragues-stripe (commented):
I focused on the Android part

android/src/main/java/com/reactnativestripesdk/NavigationBarManager.kt
gimenete-stripe: Done here https://github.com/stripe/stripe-react-native/pull/2193/commits/c3c69f9c3b47a38e81cd2782f615d1c1dd3021e9 👍

example/android/build.gradle
gimenete-stripe: launchMode=singleTask is necessary for deep linking with React Navigation:
https://reactnavigation.org/docs/deep-linking/#setup-on-android

And deep linking is necessary for using "secure webviews" (implemented with chrome tabs on Android) to return back to the app.

src/connect/Components.tsx
gimenete-stripe: If you don't export the type, and users need to use it, the'd be forced to extract them with `React.ComponentProps<typeof TheComponent>`.

src/connect/Components.tsx
gimenete-stripe: We show a native loader when we start loading the webview and we hide it when the webview shows its own loader, which is when `onLoaderStart` is fired.

example/android/build.gradle
maragues-stripe: And we can't use `AndroidManifest.xml` file, instead of modifying the manifest in the build.gradle?

android/src/main/java/com/reactnativestripesdk/NavigationBarView.kt
cttsai-stripe, line 1: nit: Could we add tests to prevent regression in a follow-up PRs? I think unit tests and e2e tests would be helpful.

src/connect/EmbeddedComponent.tsx
```diff
@@ -0,0 +1,388 @@
...
+        } else if (setter === 'setOnLoadError') {
+          onLoadError?.(value as LoadError);
+        } else {
+          // remove the 'set' prefix and lowercase the first letter
+          const functionName =
+            setter.charAt(3).toLowerCase() + setter.substring(4);
+          callbacks?.[functionName]?.(value);
+        }
+      } else if (message.type === 'openAuthenticatedWebView') {
+        const { url, id } = message.data as { id: string; url: string };
```
cttsai-stripe, line 309: Should we validate the url?

src/connect/EmbeddedComponent.tsx
```diff
@@ -0,0 +1,388 @@
...
+        }
+      }
+
+      appState.current = nextAppState;
+    });
+
+    return () => {
+      subscription.remove();
+    };
+  }, []);
```
cttsai-stripe, line 167: How does it work? Some opinions from claude:
- Race condition: AppState change might fire before deep link listener captures URL
- Always resolves with null: Doesn't actually pass the deep link URL to the callback
- No timeout: Promise could hang indefinitely if deep link never arrives

cttsai-stripe: Also:
The AppState subscription in useEffect is removed on cleanup, but:
  - If EmbeddedComponent unmounts during auth flow, pendingAuthWebViewPromise.current is orphaned
  - No cleanup of authenticationSession in iOS on unmount

  Recommendation: Add cleanup in useEffect return:
 ``` return () => {
    subscription.remove();
    // Cancel pending auth sessions
    if (Platform.OS === 'ios') {
      NativeStripeSdk.cancelAuthenticatedWebView?.();
    }
    pendingAuthWebViewPromise.current = null;
  };
```

Comment (cttsai-stripe) [reply src/connect/EmbeddedComponent.tsx:167]: Feel free to ignore comments from claude if those do not make sense.

File: src/connect/EmbeddedComponent.tsx [hunk 14]
```diff
@@ -0,0 +1,388 @@
...
+  ConnectComponentsPayload,
+  useConnectComponents,
+} from './ConnectComponentsProvider';
+import type { LoadError, LoaderStart } from './connectTypes';
+
+const DEVELOPMENT_MODE = false;
+const DEVELOPMENT_URL =
+  Platform.OS === 'android' ? 'http://10.0.2.2:3001' : 'http://localhost:3001';
+const PRODUCTION_URL = 'https://connect-js.stripe.com';
+const BASE_URL = DEVELOPMENT_MODE ? DEVELOPMENT_URL : PRODUCTION_URL;
```
cttsai-stripe, line 28: Looks like `DEVELOPMENT_MODE` is always false

src/connect/EmbeddedComponent.tsx
gimenete-stripe: That can be a good sanity check. Done [here](https://github.com/stripe/stripe-react-native/pull/2193/commits/184d03083adb824b59c079af91769ddfe09f56fc).

src/connect/EmbeddedComponent.tsx
gimenete-stripe: Cleaning up the promise is a good idea. Done [here](https://github.com/stripe/stripe-react-native/pull/2193/commits/22f5f5132d92f9ac99605319e129bc0695cb3aa8) 👍

android/src/main/java/com/reactnativestripesdk/NavigationBarView.kt
gimenete-stripe: Sure. Although it's hard to test these things because they depend on an external service I can add tests in a follow-up PR.

Also, I'm building an example app focused on embedded components that uses a different backend that allows testing multiple merchants [in this other PR](https://github.com/stripe/stripe-react-native/pull/2236). It's for manual testing but it covers everything.

src/connect/EmbeddedComponent.tsx
gimenete-stripe: Yep, this can be set to true while developing to switch to using a local server for the webview URLs and for enabling the debug mode in the webview.

Happy to leave a comment or make this configurable in a different way.

gimenete-stripe:
@cttsai-stripe thanks for the review!! 🙏 

> I'll discuss with our team to see how we should deal with private preview on RN.

Cool 👍  Thanks again! 🙇

PR #2210, opened 2025-11-07
Update iOS SDK to 25.0.0 and Android SDK to 22.0.0

yuki-stripe:
## Summary
Update iOS SDK to 25.0.0 and Android SDK to 22.0.0.

## Testing
- Manually tested deferred flow iOS
- Added tests for Android Appearance bridge refactor

TODO Probably won't actually pass tests or compile until Android releases 22.0.1 so we need to wait for that to land. Update Nov 10: They pass when I point to Android master.

## Documentation

Select one: 


porter-stripe:
Looks like you may need to run `pod install` or `pod update` in the example/ios directory

android/src/main/java/com/reactnativestripesdk/utils/Extensions.kt
```diff
@@ -48,6 +48,9 @@ fun ReadableMap?.getDoubleOr(
 fun ReadableMap?.getFloatOrNull(key: String): Float? =
   if (this?.hasKey(key) == true && this.getType(key) == ReadableType.Number) this.getDouble(key).toFloat() else null
 
+fun ReadableMap?.getBooleanOrNull(key: String): Boolean? =
```
tianzhao-stripe: Doesn't appear that you're using this function

PR #2223, opened 2025-11-14
Obo for paymentsheet

tianzhao-stripe:
## Summary
Added Example app screen

Parsing OBO in PS

## Motivation
Parity w. iOS and Android

## Testing

- [ ] I tested this manually
- [X] I added automated tests

## Documentation

Select one: 
- [X] TBD I have added relevant documentation for my changes. 


example/src/screens/PaymentSheetWithIntentConfigurationScreen.tsx
```diff
@@ -0,0 +1,152 @@
+import React, { useCallback, useState } from 'react';
```
porter-stripe: I don't think we need a whole new screen for this feature. What's the reason we can't just use one of the existing ones?

example/src/Config.ts
```diff
@@ -19,7 +20,7 @@ const CUSTOM_BACKEND_URL = 'https://your-custom-backend.com';
 // ACTIVE CONFIGURATION
 // =============================================================================
 
-export const API_URL = DEMO_BACKEND_URL;
+export const API_URL = LOCAL_URL;
```
porter-stripe: is this right?

example/server/index.ts
```diff
@@ -972,6 +974,113 @@ app.post('/detach-payment-method', async (req, res) => {
   });
 });
 
+app.get('/on-behalf-of-accounts', async (_, res) => {
```
porter-stripe: I don't think we should be updating this server, we should just remove it, and prefer all our server code lives in sandbox-apps and we run that server locally, otherwise we have two servers with two different implementations that are not sync'd.

tianzhao-stripe: agreed

example/src/screens/PaymentSheetWithIntentConfigurationScreen.tsx
tianzhao-stripe: I'm going to revert this last commit, I'll put in a demo video that shows it working

tianzhao-stripe:
https://github.com/user-attachments/assets/41c3337d-b506-445a-9b3a-745c5d93cf28

PR #2224, opened 2025-11-17
Add affirm shipping details.

jaynewstrom-stripe:
## Summary

https://admin.corp.stripe.com/request-log/req_scYkTDQseaaLyi

## Motivation

https://jira.corp.stripe.com/browse/RUN_MOBILESDK-4866

## Testing

- [x] I tested this manually
- [ ] I added automated tests

### DIFF
--- a/ios/PaymentMethodFactory.swift
+++ b/ios/PaymentMethodFactory.swift
@@ -366,8 +366,10 @@ class PaymentMethodFactory {
     }
 
     private func createAffirmPaymentMethodParams() throws -> STPPaymentMethodParams {
-        let params = STPPaymentMethodAffirmParams()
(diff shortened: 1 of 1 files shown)


.node-version
```diff
@@ -0,0 +1,2 @@
+20.19.0
```
tjclawson-stripe: Was this intentional?

jaynewstrom-stripe: Yeah, but it causes CI to break. So I probably need to revert it.

But I also need to land the iOS PR first, since this is also broken on iOS right now.

tjclawson-stripe (approved):

tjclawson-stripe (approved):

PR #2229, opened 2025-11-20
Make SDK compile with Android 22.2.+

tianzhao-stripe:
## Summary
Making RN SDK compatible with breaking changes to Appearance.Colors.Builder 
https://github.com/stripe/stripe-android/pull/11991/files

## Testing

- [ ] I tested this manually
- [X] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/PaymentSheetAppearance.kt
+++ b/android/src/main/java/com/reactnativestripesdk/PaymentSheetAppearance.kt
@@ -34,8 +34,8 @@ fun buildPaymentSheetAppearance(
 
   val builder = PaymentSheet.Appearance.Builder()
   builder.typography(buildTypography(userParams?.getMap(PaymentSheetAppearanceKeys.FONT), context))
-  builder.colorsLight(buildColorsBuilder(true, lightColorParams).build())
-  builder.colorsDark(buildColorsBuilder(false, darkColorParams).build())
+  builder.colorsLight(buildColorsBuilder(isLightMode = true, lightColorParams).build())
+  builder.colorsDark(buildColorsBuilder(isLightMode = false, darkColorParams).build())
   builder.shapes(buildShapes(userParams?.getMap(PaymentSheetAppearanceKeys.SHAPES)))
   builder.primaryButton(
     buildPrimaryButton(


android/src/androidTest/java/com/reactnativestripesdk/PaymentSheetAppearanceTest.kt
```diff
@@ -107,8 +108,8 @@ class PaymentSheetAppearanceTest {
 
     val primaryButton =
       PaymentSheet.PrimaryButton(
-        colorsLight = primaryButtonColorsBuilder.buildLight(),
-        colorsDark = primaryButtonColorsBuilder.buildDark(),
+        colorsLight = primaryButtonColorsBuilder.build(),
+        colorsDark = primaryButtonColorsBuilder.build(),
```
cttsai-stripe, line 112: Should this use a dark mode builder?

android/src/androidTest/java/com/reactnativestripesdk/PaymentSheetAppearanceTest.kt
```diff
@@ -120,8 +121,8 @@ class PaymentSheetAppearanceTest {
 
     val appearanceBuilder = PaymentSheet.Appearance.Builder()
     appearanceBuilder.typography(typographyBuilder.build())
-    appearanceBuilder.colorsLight(colorsBuilder.buildLight())
-    appearanceBuilder.colorsDark(colorsBuilder.buildDark())
+    appearanceBuilder.colorsLight(colorsBuilder.build())
+    appearanceBuilder.colorsDark(colorsBuilder.build())
```
cttsai-stripe, line 125: Same here

android/src/androidTest/java/com/reactnativestripesdk/PaymentSheetAppearanceTest.kt
```diff
@@ -173,25 +174,38 @@ class PaymentSheetAppearanceTest {
...
+    lightColorsBuilder.onSurface(testColor)
+    lightColorsBuilder.placeholderText(testColor)
+    lightColorsBuilder.error(testColor)
+
+    darkColorsBuilder.primary(testColor)
+    darkColorsBuilder.component(testColor)
+    darkColorsBuilder.componentDivider(testColor)
+    darkColorsBuilder.onSurface(testColor)
+    darkColorsBuilder.placeholderText(testColor)
+    darkColorsBuilder.error(testColor)
```
cttsai-stripe, line 192: nit: I think it would be better to place this part closer to where `build()` is called.

android/src/main/java/com/reactnativestripesdk/PaymentSheetAppearance.kt
```diff
@@ -34,8 +34,8 @@ fun buildPaymentSheetAppearance(
 
   val builder = PaymentSheet.Appearance.Builder()
   builder.typography(buildTypography(userParams?.getMap(PaymentSheetAppearanceKeys.FONT), context))
-  builder.colorsLight(buildColorsBuilder(lightColorParams).buildLight())
-  builder.colorsDark(buildColorsBuilder(darkColorParams).buildDark())
+  builder.colorsLight(buildColorsBuilder(true, lightColorParams).build())
+  builder.colorsDark(buildColorsBuilder(false, darkColorParams).build())
```
cttsai-stripe: nit:
```suggestion
  builder.colorsLight(buildColorsBuilder(isLightMode = true, lightColorParams).build())
  builder.colorsDark(buildColorsBuilder(isLightMode = false, darkColorParams).build())
```
could be more readable

android/src/main/java/com/reactnativestripesdk/PaymentSheetAppearance.kt
```diff
@@ -162,9 +170,9 @@ private fun buildPrimaryButton(
 
   return PaymentSheet.PrimaryButton(
     colorsLight =
-      buildPrimaryButtonColors(lightColorParams, context).buildLight(),
+      buildPrimaryButtonColors(true, lightColorParams, context).build(),
     colorsDark =
-      buildPrimaryButtonColors(darkColorParams, context).buildDark(),
+      buildPrimaryButtonColors(false, darkColorParams, context).build(),
```
cttsai-stripe, line 175: Same here

android/src/androidTest/java/com/reactnativestripesdk/PaymentSheetAppearanceTest.kt
cttsai-stripe, line 572: nit: I feel a bit uncomfortable that we always use a single value for all the assignable colors in a test. Would it make more sense to separate them in a follow-up PR?

cttsai-stripe: And I am also a bit surprised that we don't differentiate light and dark mode on RN while we do on Android.

android/src/androidTest/java/com/reactnativestripesdk/PaymentSheetAppearanceTest.kt
tianzhao-stripe: we're replacing all the colors here, so no real point

android/src/androidTest/java/com/reactnativestripesdk/PaymentSheetAppearanceTest.kt
tianzhao-stripe: This test only tests that we parse the color correctly when building PaymentSheet.Appearance. 

I don't think it matters that every color is the same.

cttsai-stripe: Having the same color would fail to verifying that! If we accidentally parse `separatorColor` as `disclosure` color, we wouldn't know. But again, this is out of scope for this PR.

PR #2239, opened 2025-12-03
fix: confirm platform pay on Android

janicduplessis:
## Summary

I noticed an issue with `confirmPlatformPay` that caused it to never resolve. This is a regression from the fragment removal PR, that one manager was missing calls to actually present its ui.

This also adds e2e tests for the wallet screen on Android. While adding those I noticed some other issues that required fixing:

- Stripe functions being called before initialization, in an effect. This caused the app to crash. They need to be call in the Payment screen init callback instead `onInit={...}`.
- Passing null to `token` prop would cause an error. This is because the `asMap` function will throw if the value is null. Instead use the extension function `asMapOrNull`. Also updated other places we used `asMap` to be safe.

## Motivation

Fix and prevent regression in android google pay methods.

## Testing

- [x] I tested this manually
- [x] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/e2e-tests/android-only/google-pay.yml
+++ b/e2e-tests/android-only/google-pay.yml
@@ -9,8 +9,7 @@ appId: ${APP_ID}
 
 # Donate with G Pay
 - tapOn:
-    id: 'com.google.android.gms:id/pay_button_view'
-    index: 0
+    id: 'donate-with-g-pay'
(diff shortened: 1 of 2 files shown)


janicduplessis:
Looking into why the new e2e test works locally on my emulator, but not on bitrise

PR #2258, opened 2025-12-15
Remove local server

porter-stripe:
## Summary
- This server is not used, we use the one hosted on stripe demos

## Motivation
- https://stripe.slack.com/archives/C08EWMFBWN6/p1764978263130459

## Testing
- CI 

## Documentation
N/A

### DIFF
--- a/example/src/Config.ts
+++ b/example/src/Config.ts
@@ -1,3 +1,13 @@
 // Demo backend - no configuration needed
 export const API_URL =
   'https://rigorous-heartbreaking-cephalopod.stripedemos.com';
+
+/*
+🛠️ To use Custom Backend:
+
+Remote/Codesandbox:
+1. Fork this codesandbox: https://codesandbox.io/p/devbox/rigorous-heartbreaking-cephalopod-m358cz
+2. Deploy your fork or get the preview URL
+3. Update CUSTOM_BACKEND_URL above with your URL
+4. Update API_URL above to: CUSTOM_BACKEND_URL
+*/


tianzhao-stripe:
Now that we're removing the local server, we should ensure the sandbox apps server is consistent with this implementation first.

The sandbox apps server (rigorous-heartbreaking-cephalopod/server.js) is missing:
- `/setup-without-webhooks` endpoint
- `revolut_pay` support in `getKeys()` and `/create-setup-intent`

Should we update the sandbox server before merging this PR, or file a follow-up task?

porter-stripe:
@tianzhao-stripe these are good callouts, and there is usage of setup-without-webhooks; however, these never worked anyway unless you pointed to the local server, which I don't think anyone would have thought of if they tried to use either of these. So I think the current status quo is that neither of these work since we moved off the local server, so removing the setup-without-webhooks endpoint doesn't really change anything. https://rigorous-heartbreaking-cephalopod.stripedemos.com/ doesn't have a setup-without-webhooks endpoint yet our example app was trying to access it.

I think it's better as a separate task to fix these, as they're broken today.

example/src/Config.ts
```diff
@@ -1,45 +1,3 @@
...
-
-🛠️ To use Custom Backend:
-
-Local Development:
-1. Create .env file: cp example/.env.example example/.env
-2. Add your Stripe keys to the .env file
-3. Start backend server: yarn example start:server
-4. Update API_URL above to: LOCAL_URL
-
-Remote/Codesandbox:
```
tianzhao-stripe, line 40: I think we should leave this callout in for 3rd party devs to be able to run their own version of the sandbox apps. 

Otherwise lgtm

porter-stripe: Good idea!

PR #2291, opened 2026-01-30
Add example for custom fonts

gimenete-stripe:
## Summary

Adds an example for custom fonts.

simulator_screenshot_195D73AB-E0F7-48F1-BE61-5BC449805ECF

## Motivation

Embedded components for React Native project.

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 


example-stripe-connect/src/utils/fontLoader.ts
```diff
@@ -0,0 +1,129 @@
...
+      display: 'swap', // Better UX: show fallback font while custom font loads
+    };
+  }
+
+  try {
+    // Map font paths to their static require statements
+    // Metro bundler requires static requires for asset bundling
+    const fontModules: Record<string, number> = {
+      'Handjet-Regular.woff2': require('../../assets/fonts/Handjet-Regular.woff2'),
+    };
```
maragues-stripe, line 63: Can/Should this be moved outside of the function?

gimenete-stripe: Thanks for the review!

No strong opinions. The way it's done catches the errors if require() fails. For that it needs to be surrounded by a try/catch and I feel it looks more readable if the try/catch is inside a function than in the root level.

PR #2298, opened 2026-02-09
feat: Add native CSV export with share sheet for Connect components

gimenete-stripe:
## Summary
Implements native file download and sharing for Stripe CSV exports in EmbeddedComponent. When users export data from Connect components (e.g., transactions list), the SDK now downloads the CSV file and presents a native share sheet (iOS UIActivityViewController, Android Intent) instead of displaying the file content in the WebView.

https://github.com/user-attachments/assets/a1053b76-b68e-4406-9680-d042aa424afd

## Motivation
Currently, when users export CSV files from Connect embedded components, the WebView navigates to display the raw CSV content, which provides a poor user experience. Users cannot easily save or share these exports.

This PR addresses that by:
- Intercepting CSV export URLs (matching `stripe-data-exports` domain)
- Downloading the file natively using URLSession (iOS) and OkHttp (Android)
- Presenting native share sheets for better UX
- Automatically cleaning up temporary files after sharing

This matches the behavior of the iOS Connect SDK (`stripe-ios`).

## Testing
- [x] I tested this manually on iOS simulator (iPhone 16)
- [x] I tested this manually on Android emulator
- [ ] I added automated tests

Manual testing verified:
- CSV export detection and download
- Native share sheet presentation on both platforms
- File sharing to various apps (Messages, Files, Drive)
- Temp file cleanup
- Error handling for network failures

## Documentation
- [ ] I have added relevant documentation for my changes.
- [x] This PR does not result in any developer-facing changes.

The feature is transparent to developers - CSV exports automatically work better without any API changes.

PR #2306, opened 2026-02-16
fix(android): uses unique authorities for file provider

huextrat:
## Summary
- Use a library-specific FileProvider subclass (`StripeFileProvider`) in the Android manifest instead of the generic `androidx.core.content.FileProvider`. Giving this library's provider a unique `android:name` prevents the manifest merger from merging it with another library's FileProvider, which would cause an "Attribute provider@authorities value=... is also present at..." conflict.
- Authority remains `${applicationId}.fileprovider`; no changes to `StripeSdkModule.kt` or CSV export sharing.

fixes: https://github.com/stripe/stripe-react-native/issues/2307

## Motivation
When an app uses both `@stripe/stripe-react-native` and another library that declares a `FileProvider` (e.g. for document viewing), the Android manifest merger can report a conflict on the `android:authorities` attribute because both components use the same `FileProvider` class and similar or competing authorities. Using a dedicated subclass (`StripeFileProvider`) gives this library's provider a distinct `android:name`, so the merger keeps it separate and the conflict is resolved without requiring app-level `tools:replace` workarounds.

## Testing
- [x] I tested this manually
- [ ] I added automated tests

## Documentation

Select one:

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -27,8 +27,8 @@
         </activity>
 
         <provider
-            android:name="androidx.core.content.FileProvider"
-            android:authorities="${applicationId}.stripe.fileprovider"
+            android:name=".StripeFileProvider"
+            android:authorities="${applicationId}.fileprovider"
             android:exported="false"
             android:grantUriPermissions="true">
             <meta-data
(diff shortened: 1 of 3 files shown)


huextrat:
FileProvider was added by @gimenete-stripe with https://github.com/stripe/stripe-react-native/pull/2298

PR #2319, opened 2026-02-23
Update stripe-android

tjclawson-stripe:
## Summary

Update PMME `infoIconColor` to `linkTextColor`
Remove Onramp `authenticateUser`

## Motivation

Update to latest version of `stripe-android`

## Testing

- [ ] I tested this manually
- [ ] I added automated tests

## Documentation
CHANGELOG entry
```
**Changes**
* [Breaking] `authenticateUser` has been removed in favor of `onrampAuthorize` in `useOnramp` hook.
* [Changed] Updated Stripe Android from 22.7.+ to 22.8.+
```

Select one: 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # CHANGELOG
 
-## x.x.x - 20xx-xx-xx
+## 0.59.1 - 2026-02-25
+**Fixes**
+* [Fixed] A build issue when using Xcode 16 ([#2323](https://github.com/stripe/stripe-react-native/issues/2323))
 
 **Changes**
 * [Breaking] `authenticateUser` has been removed in favor of `onrampAuthorize` in `useOnramp` hook.
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -27,8 +27,8 @@
         </activity>
 
         <provider
-            android:name="androidx.core.content.FileProvider"
-            android:authorities="${applicationId}.stripe.fileprovider"
+            android:name="com.reactnativestripesdk.StripeFileProvider"
+            android:authorities="${applicationId}.fileprovider"
             android:exported="false"
             android:grantUriPermissions="true">
             <meta-data
--- a/android/src/main/java/com/reactnativestripesdk/StripeFileProvider.kt
+++ b/android/src/main/java/com/reactnativestripesdk/StripeFileProvider.kt
@@ -1,0 +1,17 @@
+package com.reactnativestripesdk
+
+/**
+ * A custom FileProvider subclass for the Stripe React Native SDK.
+ *
+ * This class extends androidx.core.content.FileProvider to provide a unique
+ * android:name in the manifest, preventing conflicts with other libraries that
+ * also use FileProvider (e.g., react-native-document-viewer, expo-file-system).
+ *
+ * By using a library-specific class name, the Android manifest merger keeps
+ * this provider separate from other FileProviders, avoiding the
+ * "Attribute provider@authorities value=... is also present at..." error.
+ *
+ * No additional implementation is needed - this simply inherits all functionality
+ * from the parent FileProvider class.
+ */
+class StripeFileProvider : androidx.core.content.FileProvider()
--- a/android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
+++ b/android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
@@ -1518,7 +1518,7 @@ class StripeSdkModule(
       val uri =
         androidx.core.content.FileProvider.getUriForFile(
           reactApplicationContext,
-          "${reactApplicationContext.packageName}.stripe.fileprovider",
+          "${reactApplicationContext.packageName}.fileprovider",
           file,
         )
 
--- a/ios/PaymentMethodMessagingElementConfig.swift
+++ b/ios/PaymentMethodMessagingElementConfig.swift
@@ -103,12 +103,12 @@ internal class PaymentMethodMessagingElementConfig {
             }
         }
 
-        var configuration = PaymentMethodMessagingElement.Configuration(
+        let configuration = PaymentMethodMessagingElement.Configuration(
             amount: amount,
             currency: currency,
             locale: locale,
             countryCode: country,
-            paymentMethodTypes: paymentMethodTypes,
+            paymentMethodTypes: paymentMethodTypes
         )
 
         return (nil, configuration)


Twigz (changes requested):
Hey @tjclawson-stripe. Looks like a few things might be off for the Onramp changes. There's an existing branch that updates both Android and iOS here: https://github.com/stripe/stripe-react-native/compare/master...mliberatore/crypto-onramp-api-review-updates-2. I'd suggest either reviewing that for onramp code changes, or using that as a base for your changes.

For things that I saw:
1. `onrampCallbacks` has a new `onrampSessionClientSecretProvider` closure it builds with that's required.
2. The `displayData` `icon` is built differently than what was recommended for interoperability. We'll have to either switch to what was tested in the branch I linked, or test this out for functionality.
3. Looks like a lot of the appearance builders for `Colors`, `PrimaryButton`, etc. have an `else` case in this PR that goes to `null`, where now it should just return the default builder, eg. `Colors()`, `PrimaryButton()`. I think that's what's causing the oddity around `704-710`.

tjclawson-stripe:
Hey @Twigz, Thanks for the reference PR! I cherry-picked the android update commit.

android/src/onramp/java/com/reactnativestripesdk/OnrampSdkModule.kt
```diff
@@ -81,6 +81,7 @@ class OnrampSdkModule(
   private var verifyKycPromise: Promise? = null
 
   private var checkoutClientSecretDeferred: CompletableDeferred<String>? = null
+  private val rnScope = CoroutineScope(SupervisorJob() + Dispatchers.Main.immediate)
```
jaynewstrom-stripe, line 79: What's the purpose of SupervisorJob? Do we ever close this scope?

tjclawson-stripe: @Twigz can you provide context here?

android/src/onramp/java/com/reactnativestripesdk/OnrampSdkModule.kt
```diff
@@ -136,20 +160,20 @@ class OnrampSdkModule(
...
       val cryptoCustomerId = config.getString("cryptoCustomerId")
 
       val configuration =
-        OnrampConfiguration(
-          merchantDisplayName = displayName,
-          publishableKey = publishableKey,
-          appearance = appearance,
-          cryptoCustomerId = cryptoCustomerId,
-        )
+        OnrampConfiguration()
```
jaynewstrom-stripe: Do we not have any tests around this? Shouldn't we have them?

android/src/onramp/java/com/reactnativestripesdk/OnrampSdkModule.kt
Twigz: The purpose of the SupervisorJob was to maintain an active state for other children in the event one of them caused a failure. It's sort of future proofing against exceptions, as I think the existing ones are all caught. Happy to change to a different approach though. If we're keeping it, for closing scope, yeah that does still need to be done in an `invalidate()` function like so:

```
override fun invalidate() {
    super.invalidate()
    rnScope.cancel()
}
```

android/src/onramp/java/com/reactnativestripesdk/OnrampSdkModule.kt
Twigz: I don't believe there are existing tests for this, though they could certainly be added. 

I'm not 100% sure this is the appropriate place for them though. Presumably the Android SDK would be treated like it had test coverage for this functionality, otherwise we'd be adding the same unit tests across Android and RN, unless I'm misunderstanding the request?

src/hooks/useOnramp.tsx
mliberatore: Hello! From the branch that @Twigz linked, we also have a change to this documentation to remove the reference to `authenticateUser`. I’ll suggest it in this comment here for ease of incorporation, but it can be cherry-picked from bc2cd3d if desired.
```suggestion
     * Requires the user to be authenticated with prior calls to either `authenticateUserWithToken` or `authorize`,
     * and also requires prior KYC info attachement via `attachKycInfo`.
```

src/hooks/useOnramp.tsx
```diff
@@ -247,14 +242,6 @@ export function useOnramp() {
...
     updatePhoneNumber: _updatePhoneNumber,
 
-    /**
-     * Presents Link UI to authenticate an existing Link user.
-     * `hasLinkAccount` must be called before this.
-     *
-     * @returns Promise that resolves to an object with customerId or error
-     */
-    authenticateUser: _authenticateUser,
-
```
mliberatore, line 257: Also from the branch that @Twigz linked, it looks like we're missing the change to remove `authenticateUser: jest.fn(),` from `jest/setup.js`. Could that please be removed as well? This change can also be seen / cherry picked from bc2cd3d.

mliberatore (commented):
Two minor comments after reviewing the changes here against https://github.com/stripe/stripe-react-native/compare/master...mliberatore/crypto-onramp-api-review-updates-2.

src/hooks/useOnramp.tsx
tjclawson-stripe: Updated!

mliberatore (dismissed):
The updates to the changes I requested look good :+1:

android/src/onramp/java/com/reactnativestripesdk/OnrampSdkModule.kt
tjclawson-stripe: @Twigz can you update to add the invalidate function?

android/src/onramp/java/com/reactnativestripesdk/OnrampSdkModule.kt
tjclawson-stripe: We generally add unit tests to verify that the passed in config is being parsed correctly. See `PaymentElementConfigTest` for reference.

Twigz: Which does make sense in that case, as `PaymentElementConfig` is a model owned by the RN repo, and it's initialized via a ReadableMap of keys and values.

`OnrampConfiguration` is a model owned by the Android SDK, and is a class that has properties. No parsing occuring. We could test the build functions, but I would make the case for that to exist in the Android SDK tests, not in React Native.

android/src/onramp/java/com/reactnativestripesdk/OnrampSdkModule.kt
Twigz: Done!

android/src/onramp/java/com/reactnativestripesdk/OnrampSdkModule.kt
jaynewstrom-stripe: My ask would be to test the parsing logic from RN. Which seems in scope.

Twigz: I've extracted the `OnrampConfiguration` and `LinkAppearance` mapping to a new file, and added unit tests (primarily around `LinkAppearace`) to validate the duck type parsing from RN here https://github.com/stripe/stripe-react-native/pull/2319/commits/1d0801f8ae2c8d5147b69b8ec3fe8e87b7875289.

> [!NOTE]
> The `OnrampConfiguration` test purely validate creation (which doesn't currently fail anywhere). This is due to the differences between how `LinkAppearance` and `OnrampConfiguration` are scoped (OnrampConfiguration is currently largely marked as `internal` including `build`). We could update it to be available using `RestrictTo` in order to aid in testing here, but that's currently what it is marked as.

Given the above, I'm not sure on the timing of making these changes during an SDK update PR, but it's in a singular commit so if you want to keep it or back it out, up to you all.

JayNewstrom (approved):

jaynewstrom-stripe (approved):

PR #2322, opened 2026-02-24
feat: expose termsDisplay as a public PaymentSheet configuration option

guido-stripe:
## Summary
- Adds a new `termsDisplay` option to `PaymentSheet.Configuration`, allowing merchants to control whether legal agreements (e.g. card mandate disclaimers) are shown per payment method type
- Replaces the reverted internal auto-hide approach (#2309, reverted in #2317) with an explicit, user-configurable setting that matches the native iOS and Android Stripe SDK APIs
- Supports a map of payment method type strings to `'automatic'` or `'never'` (e.g. `{ card: 'never' }`)

## Changes
- **TypeScript**: Added `TermsDisplay` enum and optional `termsDisplay` property on `SetupParamsBase`
- **Android**: Added `mapToTermsDisplay()` in `PaymentElementConfig.kt`, wired into `PaymentSheetManager.kt` via `configurationBuilder.termsDisplay()`
- **iOS**: Added `mapToTermsDisplay(params:)` in `StripeSdkImpl+PaymentSheet.swift`, wired into `buildPaymentSheetConfiguration`
- **Tests**: 6 new Android unit tests, 5 new iOS unit tests covering null/missing/valid/invalid/empty cases

## Test plan
- [x] `yarn typescript` passes
- [x] `yarn lint` passes
- [x] `yarn test` — 100/100 pass
- [x] `yarn test:unit:android` — all pass
- [x] `yarn test:unit:ios` — run on CI
- [x] Manual test on iOS with `{ card: 'never' }` — card mandate disclaimer hidden
- [x] Manual test on Android with `{ card: 'never' }` — card mandate disclaimer hidden
- [x] Verify default behavior (`termsDisplay` not set) is unchanged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # CHANGELOG
 
-## x.x.x - 20xx-xx-xx
+## 0.59.1 - 2026-02-25
+**Fixes**
+* [Fixed] A build issue when using Xcode 16 ([#2323](https://github.com/stripe/stripe-react-native/issues/2323))
 
 ## 0.59.0 - 2026-02-20
 
--- a/ios/PaymentMethodMessagingElementConfig.swift
+++ b/ios/PaymentMethodMessagingElementConfig.swift
@@ -103,12 +103,12 @@ internal class PaymentMethodMessagingElementConfig {
             }
         }
 
-        var configuration = PaymentMethodMessagingElement.Configuration(
+        let configuration = PaymentMethodMessagingElement.Configuration(
             amount: amount,
             currency: currency,
             locale: locale,
             countryCode: country,
-            paymentMethodTypes: paymentMethodTypes,
+            paymentMethodTypes: paymentMethodTypes
         )
 
         return (nil, configuration)


src/types/PaymentSheet.ts
```diff
@@ -97,6 +97,17 @@ export type SetupParamsBase = IntentParams & {
...
+  /**
+   * A map of payment method types to their terms display configuration.
+   * Controls whether legal agreements (e.g. card mandate disclaimers) are shown for each payment method type.
+   * Keys are payment method type strings (e.g. "card"), values are 'automatic' or 'never'.
+   * If not set, defaults to 'automatic' for all payment method types.
+   *
+   * Example: `{ card: 'never' }` to hide the card mandate disclaimer.
+   */
+  termsDisplay?: {
+    [key: string]: TermsDisplay;
```
guido-stripe, line 111: I investigated if it made sense to use a more-typed argument for `key`, but it seems that `string` is generally used and that's what's recommended

android/src/test/java/com/reactnativestripesdk/PaymentElementConfigTest.kt
```diff
@@ -1098,6 +1099,67 @@ class PaymentElementConfigTest {
...
+    val params = readableMapOf("someOtherKey" to "value")
+    val result = mapToTermsDisplay(params)
+    assertNull(result)
+  }
+
+  @Test
+  fun mapToTermsDisplay_CardNever_ReturnsSingleEntry() {
+    val params =
+      readableMapOf(
+        "termsDisplay" to readableMapOf("card" to "never"),
```
jaynewstrom-stripe, line 1123: Can you add a test with an invalid payment method type? ie - 
```suggestion
        "termsDisplay" to readableMapOf("foobar" to "never"),
```

guido-stripe: done, both on android and ios!

yuki-stripe: I see we have tests for the bridging logic, which is good. Do we have any existing tests that use the React Native configuration API that we could add to?

src/types/PaymentSheet.ts
```diff
@@ -97,6 +97,17 @@ export type SetupParamsBase = IntentParams & {
    * the card entry form will initialize with the card scanner already open.
    * Defaults to false. */
   opensCardScannerAutomatically?: boolean;
+  /**
+   * A map of payment method types to their terms display configuration.
+   * Controls whether legal agreements (e.g. card mandate disclaimers) are shown for each payment method type.
+   * Keys are payment method type strings (e.g. "card"), values are 'automatic' or 'never'.
+   * If not set, defaults to 'automatic' for all payment method types.
+   *
+   * Example: `{ card: 'never' }` to hide the card mandate disclaimer.
```
yuki-stripe: I think the example values ('automatic', 'never') aren't right, it should use the enum?

android/src/test/java/com/reactnativestripesdk/PaymentElementConfigTest.kt
guido-stripe: According to Claude Code, there are no tests for this as it just passes things through and it's only a type interface. I manually checked that it worked as expected for our use case

src/types/PaymentSheet.ts
```diff
@@ -97,6 +97,17 @@ export type SetupParamsBase = IntentParams & {
    * the card entry form will initialize with the card scanner already open.
    * Defaults to false. */
   opensCardScannerAutomatically?: boolean;
+  /**
+   * A map of payment method types to their terms display configuration.
+   * Controls whether legal agreements (e.g. card mandate disclaimers) are shown for each payment method type.
+   * Keys are payment method type strings (e.g. "card"), values are `TermsDisplay.AUTOMATIC` or `TermsDisplay.NEVER`.
```
yuki-stripe: Let's also specify the format of the pm type strings in the docstring - snake case - and call out you can find the full list of values here https://docs.stripe.com/api/payment_methods/object#payment_method_object-type

guido-stripe: Done!

yuki-stripe (approved):

Issue #2245, opened 2025-12-10
Implement Android setup instructions in the expo config plugin


Upsilon-Iridani:
**Is your feature request related to a problem? Please describe.**
https://github.com/stripe/stripe-react-native/issues/355

**Describe the solution you'd like**
The [stripe config plugin](https://github.com/stripe/stripe-react-native/blob/master/src/plugin/withStripe.ts) should (optionally?) implement the setup instructions outlined in [docs/android-chrome-tab-closes-on-background.md](https://github.com/stripe/stripe-react-native/blob/master/docs/android-chrome-tab-closes-on-background.md)

**Describe alternatives you've considered**
Writing the config myself.

**Additional context**
I'm guessing every expo user of the library deals with the same setup hassle?

Upsilon-Iridani:
I made a draft, but I think more eyes are needed on this one. Thoughts?

gbirch-stripe:
@Upsilon-Iridani thanks for opening a pull request! We will be sure to take a look at it

Upsilon-Iridani:
@gbirch-stripe, I think @huextrat's #2341 is the better option. Is it possible for you to prioritise this?

PR #2341, opened 2026-03-10
feat: expo plugin to enable launch activity

huextrat:
## Summary

Following this PR https://github.com/stripe/stripe-react-native/pull/2246 we are using something similar so here is the combination of both

It allows you to have the same theme as your MainApplication as `expo-web-browser` was doing https://github.com/expo/expo/blob/1f163e3d8ea3ad9ff7c2cfeb84851261e2cacdb0/packages/expo-web-browser/plugin/src/withWebBrowserAndroid.ts before the removal of `experimentalLauncherActivity`: https://github.com/expo/expo/pull/41535 which was useful to fix this Stripe issue this plugin is no longer present in latest Expo versions

## Motivation

fixes: https://github.com/stripe/stripe-react-native/issues/2245

## Testing

- [X] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
   },
   "homepage": "https://github.com/stripe/stripe-react-native/#readme",
   "devDependencies": {
-    "@expo/config-plugins": "^9.0.16",
+    "@expo/config-plugins": "^55.0.6",
     "@react-native/babel-preset": "^0.81.0",
     "@react-native/eslint-config": "^0.81.1",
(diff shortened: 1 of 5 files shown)


Upsilon-Iridani:
Nice! I ended up using the `experimentalLauncherActivity ` from `'expo-web-browser'` too, but just like you mention it isn't a viable solution any longer.

PR #2347, opened 2026-03-17
Refactor PaymentSheetManager

tjclawson-stripe:
## Summary

- Refactor `PaymentSheetManager` to separate configuration into separate method
- Update `StripeSdkModule.initPaymentSheet` to only call `PaymentSheetManager.onCreate` if class instance is null

## Motivation

Fix error where calling `initPaymentSheet` multiple times causes multiple instances of PaymentSheet/FlowController to be instantiated, which results in confirmation callbacks being delivered to incorrect instance when attempting multiple payments. Non-compose stripe-android PaymentSheet/FlowController APIs do not support multiple instances

## Testing

- [X] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/PaymentSheetManager.kt
+++ b/android/src/main/java/com/reactnativestripesdk/PaymentSheetManager.kt
@@ -283,7 +283,6 @@ class PaymentSheetManager(
               .confirmCustomPaymentMethodCallback(this)
               .build(activity, signal)
           }
-        initPromise.resolve(Arguments.createMap())
       }
     }
     configure(arguments, initPromise)
@@ -300,7 +299,7 @@ class PaymentSheetManager(
   ) {
     val merchantDisplayName = args.getString("merchantDisplayName").orEmpty()
     if (merchantDisplayName.isEmpty()) {
-      initPromise.resolve(
+      promise.resolve(
         createError(ErrorType.Failed.toString(), "merchantDisplayName cannot be empty or null."),
       )
       return
@@ -322,7 +321,7 @@ class PaymentSheetManager(
       try {
         buildIntentConfiguration(args.getMap("intentConfiguration"))
       } catch (error: PaymentSheetException) {
-        initPromise.resolve(createError(ErrorType.Failed.toString(), error))
+        promise.resolve(createError(ErrorType.Failed.toString(), error))
         return
       }
 
@@ -330,7 +329,7 @@ class PaymentSheetManager(
       try {
         buildPaymentSheetAppearance(args.getMap("appearance"), context)
       } catch (error: PaymentSheetAppearanceException) {
-        initPromise.resolve(createError(ErrorType.Failed.toString(), error))
+        promise.resolve(createError(ErrorType.Failed.toString(), error))
         return
       }
 
@@ -338,7 +337,7 @@ class PaymentSheetManager(
       try {
         buildCustomerConfiguration(args)
       } catch (error: PaymentSheetException) {
-        initPromise.resolve(createError(ErrorType.Failed.toString(), error))
+        promise.resolve(createError(ErrorType.Failed.toString(), error))
         return
       }
 
@@ -380,8 +379,10 @@ class PaymentSheetManager(
     mapToTermsDisplay(args)?.let { configurationBuilder.termsDisplay(it) }
 
     paymentSheetConfiguration = configurationBuilder.build()
-    if (arguments.getBooleanOr("customFlow", false)) {
+    if (args.getBooleanOr("customFlow", false)) {
       configureFlowController(promise)
+    } else {
+      promise.resolve(Arguments.createMap())
     }
   }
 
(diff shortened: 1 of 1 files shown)


example/src/screens/PaymentSheetDeferredIntentMultiStepScreen.tsx
```diff
@@ -186,7 +186,9 @@ export default function PaymentSheetDeferredIntentMultiStepScreen() {
       Alert.alert('Success', 'The payment was confirmed successfully!');
       setPaymentSheetEnabled(false);
     }
+    setPaymentMethodOption(null);
     setLoading(false);
+    await initialisePaymentSheet();
```
tjclawson-stripe, line 191: Should I leave this and write and E2E test that attempts multiple payments?

android/src/main/java/com/reactnativestripesdk/PaymentSheetManager.kt
```diff
@@ -268,6 +235,118 @@ class PaymentSheetManager(
...
+            @SuppressLint("RestrictedApi")
+            PaymentSheet
+              .Builder(paymentResultCallback)
+              .confirmCustomPaymentMethodCallback(this)
+              .build(activity, signal)
+          }
+        initPromise.resolve(Arguments.createMap())
+      }
+    }
+    configure(arguments, initPromise)
```
cttsai-stripe, line 288: `initPromise` is resolved as success on line 286 (non-customFlow, first creation), and then `configure()` runs here. If `configure()` hits a validation error (bad appearance, empty merchantDisplayName, invalid customer config), it tries to resolve `initPromise` again — but `PromiseImpl` nulls its callbacks after the first `resolve()`, so the error is silently dropped. JS sees success even when config is invalid.

Suggest removing the `initPromise.resolve(Arguments.createMap())` on line 286, and letting `configure()` handle promise resolution for both customFlow and non-customFlow paths (see comment on line 383).

android/src/main/java/com/reactnativestripesdk/PaymentSheetManager.kt
```diff
@@ -268,6 +235,118 @@ class PaymentSheetManager(
...
+    paymentSheet = null
+  }
+
+  fun configure(
+    args: ReadableMap,
+    promise: Promise,
+  ) {
+    val merchantDisplayName = args.getString("merchantDisplayName").orEmpty()
+    if (merchantDisplayName.isEmpty()) {
+      initPromise.resolve(
```
cttsai-stripe: `configure()` accepts a `promise` parameter but all error paths resolve `initPromise` (the constructor field) instead. This works on the first call since `onCreate()` passes `initPromise` as the `promise` arg. But on the reuse path, `StripeSdkModule` calls `configure(params, newPromise)` with a fresh bridge promise — errors go to the stale `initPromise` (already resolved, so silently dropped) and `newPromise` hangs forever.

Same issue on lines 325, 333, 341.

```suggestion
      promise.resolve(
```

android/src/main/java/com/reactnativestripesdk/PaymentSheetManager.kt
```diff
@@ -283,72 +362,26 @@ class PaymentSheetManager(
...
-      mapToPaymentMethodLayout(arguments.getString("paymentMethodLayout")),
+      mapToPaymentMethodLayout(args.getString("paymentMethodLayout")),
     )
 
-    mapToTermsDisplay(arguments)?.let { configurationBuilder.termsDisplay(it) }
+    mapToTermsDisplay(args)?.let { configurationBuilder.termsDisplay(it) }
 
     paymentSheetConfiguration = configurationBuilder.build()
-
     if (arguments.getBooleanOr("customFlow", false)) {
```
cttsai-stripe: Two issues here:

1. **`arguments` vs `args`**: This reads from `arguments` (the constructor field — first call's args) instead of `args` (the method parameter). So if a second call changes `customFlow`, it checks stale state.

2. **Non-customFlow never resolves `promise`**: Only the customFlow branch calls `configureFlowController(promise)`. For non-customFlow, the method falls off the end without resolving `promise`. On the reuse path, the JS `await initPaymentSheet()` hangs forever.

Fix for both:
```suggestion
    if (args.getBooleanOr("customFlow", false)) {
      configureFlowController(promise)
    } else {
      promise.resolve(Arguments.createMap())
    }
```

android/src/main/java/com/reactnativestripesdk/PaymentSheetManager.kt
tjclawson-stripe: Great catch, thanks!

android/src/main/java/com/reactnativestripesdk/PaymentSheetManager.kt
tjclawson-stripe: Updated!

android/src/main/java/com/reactnativestripesdk/PaymentSheetManager.kt
tjclawson-stripe: Updated!

cttsai-stripe (approved):

PR #2348, opened 2026-03-17
StripeCryptoOnramp: CardAndBankAccount / GooglePay Update

Twigz:
## Summary
1. Updates to the latest Android SDK
2. Adds CardAndBankAccount option for both iOS/Android
3. Fixes an issue with payment display data for settlement speed.
4. Adds GooglePay as a payment option on Android.
5. Updates example app to make use of these new options.

## Motivation
Implementing GooglePay to match the existence of ApplePay on iOS.

## Testing

- [x] I tested this manually
- [ ] I added automated tests

## Documentation
Breaking: Removed `PaymentOptionData` type, as it was duplicative of `PaymentMethodDisplayData`. Going forward use the `PaymentMethodDisplayData` type in place of that.

Select one: 

### DIFF
--- a/src/types/Onramp.ts
+++ b/src/types/Onramp.ts
@@ -259,7 +259,7 @@ export type PaymentMethodDisplayData = {
   /** Details about the underlying payment method, e.g., "Visa Credit •••• 4242". */
   sublabel?: string;
   /** The type of payment method */
-  type: string;
+  type: 'Card' | 'BankAccount' | 'ApplePay' | 'GooglePay';
 };
 
 /**


example/src/screens/Onramp/CryptoOnrampFlow.tsx
```diff
@@ -52,6 +51,7 @@ import {
   OnrampResponseStatusSection,
 } from './sections';
 import { colors } from '../../colors';
+import { PaymentMethodDisplayData } from '../../../../src/types/Onramp';
```
jeanregisser: Can we import from the package instead?

src/types/index.ts
```diff
@@ -293,9 +293,3 @@ export type CollectBankAccountForSetupResult =
       setupIntent?: undefined;
       error: StripeError<CollectBankAccountError>;
     };
-
-export type PaymentOptionData = {
```
jeanregisser, line 297: This looks like a breaking change. Is this ok?
Are there existing integrators using it?

src/types/Onramp.ts
```diff
@@ -228,6 +258,8 @@ export type PaymentMethodDisplayData = {
   label: string;
   /** Details about the underlying payment method, e.g., "Visa Credit •••• 4242". */
   sublabel?: string;
+  /** The type of payment method */
+  type: string;
```
jeanregisser: Could we use something like this instead?


```suggestion
  type: 'Card' | 'BankAccount' | 'ApplePay' | 'GooglePay';
```

ios/StripeSdkImpl.swift
```diff
@@ -1545,11 +1545,13 @@ public class StripeSdkImpl: NSObject, UIAdaptivePresentationControllerDelegate {
             paymentMethodType = .card
         case "BankAccount":
             paymentMethodType = .bankAccount
+        case "CardAndBankAccount":
+            paymentMethodType = .cardAndBankAccount
         case "PlatformPay":
             guard let applePayParams = platformPayParams["applePay"] as? NSDictionary else {
                 resolve(Errors.createError(ErrorType.Failed, "You must provide the `applePay` parameter."))
                 return
-            }
```
jeanregisser, line 1552: Nothing to do now, just noting there may be an opportunity for CI enforced formatting in swift files.

android/src/onramp/java/com/reactnativestripesdk/OnrampSdkModule.kt
```diff
@@ -596,6 +620,12 @@ class OnrampSdkModule(
       displayData.putString("label", paymentDetails.label)
       displayData.putString("sublabel", paymentDetails.sublabel)
 
+      if (token.hasKey("card")) {
+        displayData.putString("type", "Card")
+      } else if (token.hasKey("us_bank_account")) {
+        displayData.putString("type", "BankAccount")
+      }
```
jeanregisser, line 627: Should we see `GooglePay` here as well?
And is it fine to have `type` not set if no case match?

android/src/onramp/java/com/reactnativestripesdk/OnrampSdkModule.kt
```diff
@@ -433,6 +433,30 @@ class OnrampSdkModule(
...
+              ?: run {
+                promise.resolve(
+                  createFailedError(
+                    IllegalArgumentException("Missing googlePay params in platformPayParams"),
+                  ),
+                )
+                return
+              }
+          val currencyCode = googlePayParams.getString("currencyCode") ?: ""
+          val amount = googlePayParams.getDouble("amount").toLong()
```
jeanregisser, line 449: What unit is `amount` in, minor units (cents) or major units (dollars)? Is there a JSDoc comment on the type to make it explicit?

example/src/screens/Onramp/CryptoOnrampFlow.tsx
```diff
@@ -464,6 +465,10 @@ export default function CryptoOnrampFlow() {
     handleCollectPaymentMethod({ type: 'BankAccount' });
   }, [handleCollectPaymentMethod]);
 
+  const handleCollectCardAndBankAccountPayment = useCallback(async () => {
```
jeanregisser, line 468: Not for this PR, but I notice the example app uses `useCallback` extensively for handlers that may not need memoization. Might be worth simplifying in a future pass,  example code tends to get copy-pasted, and unnecessary `useCallback` is a common misconception.

android/src/onramp/java/com/reactnativestripesdk/OnrampSdkModule.kt
Twigz: It is fine in this particular instance yes. Only Card and BankAccount are valid here.

android/src/onramp/java/com/reactnativestripesdk/OnrampSdkModule.kt
Twigz: For google pay the amount is in the lowest denomination of that currency, so in the case of US, that would cents. It is described as such on the `GooglePayBaseParams` type documentation of the property..

src/types/index.ts
Twigz: It is technically a breaking change, not 100% sure if integrators are using it. It's an unnecessary duplicate of `PaymentMethodDisplayData` that shouldn't have been added, so we're using this opportunity to rectify that. We should make it clear in the release that any references can be replaced with that.

jeanregisser: ok I'll check with @mats-stripe,
in the meantime, could we make the breaking change explicit in the PR title and/or description?

android/src/onramp/java/com/reactnativestripesdk/OnrampSdkModule.kt
jeanregisser: ok, and should we throw an error if something else shows up? instead of having `type` not set?
just wanna avoid silent errors.
Or is it really impossible?

Twigz: If it's neither of those, previously in the function an illegal argument exception would be created and sent via the promise. I could move this particular section up a bit, which would be less duplicative of the check, but it does separate the creation of the display data a bit more. I'm fine with either.

PR #2351, opened 2026-03-19
Add PMME API

tjclawson-stripe:
## Summary

- Add PaymentMethodMessagingElement
- Add PaymentMethodMessagingElementFont
- Add PaymentMethodMessagingElementScreen
- Add height measurement fix for iOS
- Update iOS and android parses to use PaymentMethodMessagingElementFont
- Add e2e test

## Motivation

- Release PMME
- Changed to PaymentMethodMessagingElementFont as there is only one font size for PMME so using a size scale factor isn't intuitive

## Testing

- [ ] I tested this manually
- [X] I added automated tests

## Documentation

Select one: 


src/components/PaymentMethodMessagingElement.tsx
```diff
@@ -0,0 +1,128 @@
...
+      // listen for height changes
+      const sub = addListener(
+        'paymentMethodMessagingElementDidUpdateHeight',
+        ({ height: h }) => {
+          LayoutAnimation.configureNext(LayoutAnimation.Presets.easeInEaseOut);
+          setHeight(h);
+        }
+      );
+      return () => sub.remove();
+    });
```
porter-stripe, line 70: I think we're missing the dependency array, I believe this should be `}, []);.`, but maybe not an issue if it's meant to be empty.

src/components/PaymentMethodMessagingElement.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+    useEffect(() => {
+      // listen for load complete
+      const sub = addListener(
```
porter-stripe, line 74: This listener also calls onStateChange, but the native onStateChange direct event handler at line 66 does the same. Will users receive duplicate callbacks? Can we remove one of them?

src/components/PaymentMethodMessagingElement.tsx
```diff
@@ -0,0 +1,128 @@
...
+ *
+ * @example
+ * ```tsx
+ * <PaymentMethodMessagingElement
+ *    appearance={{
+ *      style: Style.Dark,
+ *      textColor: '#FFFFFF',
+ *      infoIconColor: '#CCCCCC',
+ *      font: {
+ *        family: 14,
```
porter-stripe: 14? Shouldn't this be a string?

ios/PaymentMethodMessagingElementConfig.swift
```diff
@@ -48,6 +51,18 @@ internal class PaymentMethodMessagingElementConfig {
         return appearance
     }
 
+    private class func buildFont(params: NSDictionary) throws -> PaymentSheet.Appearance.Font {
```
porter-stripe: This isn't called from anywhere, is that intentional?

ios/PaymentMethodMessagingElementView.swift
```diff
@@ -76,7 +92,10 @@ public class PaymentMethodMessagingElementContainerView: UIView, UIGestureRecogn
         let messagingElementView = messagingElement.view
         addSubview(messagingElementView)
         messagingElementView.translatesAutoresizingMaskIntoConstraints = false
+        let height = self.messagingInstance?.view.systemLayoutSizeFitting(CGSize(width: self.messagingInstance?.view.bounds.width ?? 0, height: UIView.layoutFittingCompressedSize.height)).height
+        self.previousHeight = height ?? 0
 
+        StripeSdkImpl.shared.emitter?.emitPaymentMethodMessagingElementDidUpdateHeight(["height": height ?? 0])
```
porter-stripe, line 98: This emits height immediately, but layoutSubviews (line 73) will also fire right after since addSubview triggers a layout pass. Can we remove one of the height emissions? Maybe just rely on layoutSubViews?

src/index.tsx
```diff
@@ -36,6 +36,10 @@ export * from './types/EmbeddedPaymentElement';
 export * from './types/PaymentSheet';
 export * from './types/ConfirmationToken';
 
+export * from './types/components/PaymentMethodMessagingElementComponent';
+export type { Props as PaymentMethodMessagingProps } from './components/PaymentMethodMessagingElement';
```
porter-stripe: nit: For consistency with the component name, consider PaymentMethodMessagingElementProps.

porter-stripe (commented):
Can we update jest/mock.js to provide a mock for PaymentMethodMessagingElement?

src/index.tsx
tjclawson-stripe: Updated

ios/PaymentMethodMessagingElementView.swift
tjclawson-stripe: This is required to fix a bug where if the view reloads but the result doesn't change, the `see plans` CTA will be cut off.

src/components/PaymentMethodMessagingElement.tsx
tjclawson-stripe: Updated

src/components/PaymentMethodMessagingElement.tsx
tjclawson-stripe: Removed native onStateChange as it wasn't actually being called

src/components/PaymentMethodMessagingElement.tsx
tjclawson-stripe: I don't think we need a dependency here because we're calling setHeight directly

ios/PaymentMethodMessagingElementConfig.swift
tjclawson-stripe: Removed

tjclawson-stripe:
> Can we update jest/mock.js to provide a mock for PaymentMethodMessagingElement?

@porter-stripe added `PaymentMethodMessagingElement: () => 'PaymentMethodMessagingElementView'` to `mock.js`. I think that's all that's required based on CardField/CardForm, which are also just views

porter-stripe (approved):

PR #2358, opened 2026-03-25
update iOS dependency to 25.9.0

gbirch-stripe:
## Summary
- update iOS native SDK dependency from 25.7.0 to 25.9.0
- update CryptoOnRamp SDK wrapper

## Motivation
- include latest iOS bug fixes

## Testing
N/A

## Documentation

Select one: 

### DIFF
--- a/ios/StripeSdkImpl.swift
+++ b/ios/StripeSdkImpl.swift
@@ -8,7 +8,7 @@ import StripeFinancialConnections
 import StripePaymentsUI
 import UIKit
 #if canImport(StripeCryptoOnramp)
-@_spi(STP) import StripeCryptoOnramp
+@_spi(CryptoOnrampAlpha) import StripeCryptoOnramp
 
 @_spi(STP)
 @_spi(EmbeddedPaymentElementPrivateBeta)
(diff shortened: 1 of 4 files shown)


ios/StripeSdkImpl.swift
```diff
@@ -1347,6 +1347,13 @@ public class StripeSdkImpl: NSObject, UIAdaptivePresentationControllerDelegate {
                 try await coordinator.authenticateUserWithToken(linkAuthTokenClientSecret)
                 resolve([:])  // Return empty object on success
             } catch {
+                if let onrampError = error as? CryptoOnrampCoordinator.Error,
+                   case let .seamlessSignInTokenInvalid(reason) = onrampError {
+                    let errorResult = Errors.createError(ErrorType.Failed, reason ?? onrampError.localizedDescription)
+                    resolve(["error": errorResult["error"]!])
+                    return
+                }
+
```
Twigz, line 1356: What's the reason for the special casing for this particular error case here?

gbirch-stripe: If you look at https://github.com/stripe/stripe-react-native/blob/b613850fcd1047a02a9174cf6a28640431d2eb30/ios/Errors.swift#L94, you can see that we will maket he error into an NSError with a localizedDescription but with the `reason` stripped out and no longer accessible for RN users, and so we need to add handling if we want to pass this info on

Twigz: Ah, I see, and the previous cases didn't have reasons which is why that didn't crop up until now. Makes sense!

amk-stripe (approved):
LGTM -- can you please add a PR description?

PR #2361, opened 2026-03-25
Allow usage of PaymentSheet and FlowController

tjclawson-stripe:
## Summary

- Refactor `PaymentSheetManager` to be able to instantiate both FlowController and PaymentSheet
- Keep track of whether PS or FC was last configured, present whichever was last configured

## Motivation

Allow usage of both FC and PS in the same app

https://github.com/user-attachments/assets/18312e71-422a-46ee-b5b4-2c7f52d84ab6

## Testing

- [X] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # CHANGELOG
 
+## 0.61.0 - 2026-03-25
+**Changes**
+- Updated Stripe iOS SDK from 25.7.+ to 25.9.+.
+- Updated Stripe Android SDK from 23.0.+ to 23.1.+.
+
+**Fixes**
+* [Fixed] Enum type mismatch that caused iOS build failures on Xcode 26+. ([#2357](https://github.com/stripe/stripe-react-native/issues/2357))
+
 ## 0.60.0 - 2026-03-23
 **Changes**
 - Updated Stripe Android SDK from 22.8.+ to 23.0.+. See our [migration guide](https://github.com/stripe/stripe-android/blob/master/MIGRATING.md#migrating-from-versions--2300) for more information.
--- a/android/src/main/java/com/reactnativestripesdk/pushprovisioning/AddToWalletButtonView.kt
+++ b/android/src/main/java/com/reactnativestripesdk/pushprovisioning/AddToWalletButtonView.kt
@@ -2,6 +2,7 @@ package com.reactnativestripesdk.pushprovisioning
 
 import android.annotation.SuppressLint
 import android.content.res.ColorStateList
+import android.graphics.Bitmap
 import android.graphics.drawable.Drawable
 import android.graphics.drawable.RippleDrawable
 import android.view.MotionEvent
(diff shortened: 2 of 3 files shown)


android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
```diff
@@ -258,7 +258,9 @@ class StripeSdkModule(
     promise: Promise,
   ) {
     if (paymentSheetManager != null) {
-      paymentSheetManager?.configure(params, promise)
+      UiThreadUtil.runOnUiThread {
```
tjclawson-stripe, line 261: This has to be called on UiThread, see StripeSdkModule.registerStripeUIManager which calls uiManager.create() which invokes the onCreate fun on the UI Thread

android/src/main/java/com/reactnativestripesdk/PaymentSheetManager.kt
```diff
@@ -380,15 +186,211 @@ class PaymentSheetManager(
...
+    return CreateIntentWithConfirmationTokenCallback { confirmationToken ->
+      val stripeSdkModule: StripeSdkModule? = context.getNativeModule(StripeSdkModule::class.java)
+      val params =
+        Arguments.createMap().apply {
+          putMap("confirmationToken", mapFromConfirmationToken(confirmationToken))
+        }
+
+      stripeSdkModule?.eventEmitter?.emitOnConfirmationTokenHandlerCallback(params)
+
+      val resultFromJavascript = paymentSheetConfirmationTokenCreationCallback.await()
```
samer-stripe, line 274: Does this timeout after some time?

android/src/main/java/com/reactnativestripesdk/PaymentSheetManager.kt
```diff
@@ -86,209 +86,15 @@ class PaymentSheetManager(
...
-              .confirmCustomPaymentMethodCallback(this)
-              .build(activity, signal)
-          }
-      }
-    }
     configure(arguments, initPromise)
   }
 
   override fun onDestroy() {
+    super.onDestroy()
```
samer-stripe, line 97: OOC, why do we need to call the `super` call now?

tjclawson-stripe: super calls `signal.unregister()` for the signal passed for PaymentSheet which I believe needs to be called

android/src/main/java/com/reactnativestripesdk/PaymentSheetManager.kt
tjclawson-stripe: I don't believe so, it would hang indefinitely if the intent/confirmation callback did

samer-stripe: Probably not important since the same thing can happen with the native implementation but was curious.

PR #2371, opened 2026-03-30
StripeCryptoOnramp: ContentOnPrimary Fix

Twigz:
## Summary
Fixes an issue where `contentOnPrimary` was not a valid appearance customization for iOS on CryptoOnramp. This makes use of the already existing `contentOnPrimary` flag in RN that can be set, but has no function on iOS.

## Motivation
Matching customization options from Android.

## Testing

- [x] I tested this manually
- [ ] I added automated tests

1. Launch the app.
2. Login.
3. Collect Payment. Make sure the text on the button matches the `contentOnPrimary` setting.
4. Change the `contentOnPrimary` color, make sure it's reflected in the app.
5. Test on light and dark mode.

Simulator Screenshot - iPhone 16 Pro Max - 2026-03-30 at 11 27 45

## Documentation

Select one: 

### DIFF
--- a/ios/Mappers.swift
+++ b/ios/Mappers.swift
@@ -1214,7 +1214,7 @@ class Mappers {
             darkContentOnPrimary ?? lightContentOnPrimary
         }
 
-        let colors: LinkAppearance.Colors? = if primary != nil || selectedBorder != nil {
+        let colors: LinkAppearance.Colors? = if primary != nil || selectedBorder != nil || contentOnPrimary != nil {
             .init(primary: primary, contentOnPrimary: contentOnPrimary, selectedBorder: selectedBorder)
         } else {
             nil


ios/Mappers.swift
```diff
@@ -1201,8 +1206,16 @@ class Mappers {
             darkSelectedBorder ?? lightSelectedBorder
         }
 
-        let colors: LinkAppearance.Colors? = if primary != nil || selectedBorder != nil {
-            .init(primary: primary, selectedBorder: selectedBorder)
+        let contentOnPrimary: UIColor? = if let darkContentOnPrimary, let lightContentOnPrimary {
+            UIColor { trait in
+                trait.userInterfaceStyle == .dark ? darkContentOnPrimary : lightContentOnPrimary
+            }
```
mats-stripe, line 1212: Will this update dynamically when the device theme changes? In the past I've used this API: https://developer.apple.com/documentation/uikit/uicolor/init(dynamicprovider:) but since the pattern in your code is used for other colors, we don't need to change anything here 👍

Twigz: When I tested it, it seemed to update just fine!

mats-stripe: Oh, perfect!

Issue #2352, opened 2026-03-20
Add support for multibanco


remonh87:
**Is your feature request related to a problem? Please describe.**
Both the iOS and Android sdks of Stripe support the multibanco payment method but the react native sdk doesn't. Some users of the Flutter Stripe library request us to implement it but we need the binding methods from the react native sdk first.

**Describe the solution you'd like**
Add support for multibanco

**Describe alternatives you've considered**
- 

**Additional context**
[Offical multibanco docs](https://docs.stripe.com/payments/multibanco/accept-a-payment?payment-ui=mobile&platform=android)

porter-stripe:
We just merged in support for this payment method, it will be available in the next release.

PR #2392, opened 2026-04-13
Add multibanco support

porter-stripe:
## Summary
- Adds Multibanco support via bindings

https://github.com/user-attachments/assets/e303f379-9bcd-4b56-bd2a-ccbdf3ff23b5

## Motivation
- https://github.com/stripe/stripe-react-native/issues/2352

## Testing
- New tests
- Manual

### DIFF
--- a/example/src/screens/MultibancoPaymentScreen.tsx
+++ b/example/src/screens/MultibancoPaymentScreen.tsx
@@ -1,0 +1,163 @@
+import { PaymentIntent, useConfirmPayment } from '@stripe/stripe-react-native';
+import React, { useMemo, useState } from 'react';
+import { Alert, StyleSheet, Text, TextInput, View } from 'react-native';
+import Button from '../components/Button';
+import PaymentScreen from '../components/PaymentScreen';
+import { API_URL } from '../Config';
+import { colors } from '../colors';
+
+type MultibancoVoucherDetails = {
+  voucherURL?: string;
+  entity?: string;
+  reference?: string;
+  expiresAt?: number;
+};
+
+export default function MultibancoPaymentScreen() {
+  const [email, setEmail] = useState('');
+  const [voucherDetails, setVoucherDetails] =
+    useState<MultibancoVoucherDetails | null>(null);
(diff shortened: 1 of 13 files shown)


example/src/screens/MultibancoPaymentScreen.tsx
```diff
@@ -0,0 +1,163 @@
+import { PaymentIntent, useConfirmPayment } from '@stripe/stripe-react-native';
```
yuki-stripe, line 1: do we really want to add an example screen for every LPM?

ios/Mappers.swift
```diff
@@ -475,6 +477,14 @@ class Mappers {
                     "type": "konbiniVoucher",
                     "voucherURL": it.konbiniDisplayDetails?.hostedVoucherURL.absoluteString ?? NSNull(),
                 ]
+            case .multibancoDisplayDetails:
```
yuki-stripe, line 480: huh it seems like all of this code is to generate a dictionary from a STPPaymentIntent object. But the  the STPPaymentIntent object was built from JSON. 

So afaict we've gone from `json response` -> `STPPaymentIntent` -> manually creating the `json response` but with camel case. We could probably just return the paymentIntent.allResponseFields, camel cased? Also Stripe.js uses snake case, we should consider aligning. I can't find any reason why we've decided on camel case here.

Another idea - we don't even really need to do all this work to call into the native SDKs at all. We could  delete _all_ of this iOS and Android code and implement these simple API calls once, directly in JS. We'd just need to handle the small set of extra params/headers the native SDKs send like `payment_user_agent` and things from `STPTelemetryClient`.

example/src/screens/MultibancoPaymentScreen.tsx
porter-stripe: I'm on the fence, I primarily added it for my own testing and it fits the pattern of the repo. However, I can see the side that it's a large artifact for this feature, but its kinda hard to test multibanco w/o it. lmk what you think.

yuki-stripe (approved):

PR #2395, opened 2026-04-15
CryptoOnramp SDK: Cleaner Separation on iOS for Excluding Onramp

mliberatore:
## Summary
Removes stub implementations for Onramp methods in favor of keeping `StripeOnrampSdk.h/mm` specific to the Onramp subspec.

## Motivation
[This comment from Jean](https://github.com/stripe/stripe-react-native/pull/2380#pullrequestreview-4078249625). Specifically:

> One thing worth noting: iOS behaves differently from Android in the "without Onramp" case. On Android the module is absent entirely so `NativeOnrampSdk` is `null` and `requireOnrampModule()` throws. On iOS, `StripeOnrampSdk.mm` is part of the base pod's `ios/**/*.mm` glob so the module is always compiled in — the error comes from native stubs in `StripeSdkImpl.swift` (`#if canImport(StripeCryptoOnramp)`) instead. Both end up in the right place, but it's inconsistent. To align iOS with Android you'd move `StripeOnrampSdk.mm` into the Onramp subspec, update the `requireOnrampModule()` message to cover both platforms, and delete the `#else` stubs. Out of scope here but worth a follow-up.

## Testing

- [x] I tested this manually
- [ ] I added automated tests

1. Removed the line that adds Onramp from the example app Podfile.
2. Deleted the corresponding Podfile.lock and Pods directory.
3. Ran `yarn bootstrap` and `yarn run-example-ios`.
4. Tried tapping the Onramp options on the homescreen and saw the relevant alert message stating that Onramp was unavailable.
5. Re-added the removed line from 1.
6. Repeated steps 2+3.
7. Tested a full onramp login -> checkout flow successfully.

## Documentation

Select one: 


example/src/screens/HomeScreen.tsx
```diff
@@ -105,6 +105,23 @@ export default function HomeScreen() {
       });
   }, [configure]);
 
+  const handleNavigateToOnramp = useCallback(
```
mliberatore, line 109: Rather than adding a `.catch` to every call from the example flows to Onramp APIs, I instead call `requireOnrampModule()` here and catch the error, display an alert.

jeanregisser (approved):
Great cleanup ✨

iOS and Android now handle the "Onramp not included" case consistently. 

Found a build cycle in Expo 54 (New Arch) caused by `private_header_files` being left at the top-level spec after `source_files` moved into `Core` - pushed a fix directly to the branch.

Tested on iOS across all combinations:

| Project | Arch | Onramp | Result |
|---|---|---|---|
| Expo 55 (RN 0.83) | New | Without | Error shown ✅ |
| Expo 55 (RN 0.83) | New | With | Flow works ✅ |
| Expo 54 (RN 0.81) | Old | Without | Error shown ✅ |
| Expo 54 (RN 0.81) | Old | With | Flow works ✅ |
| Expo 54 (RN 0.81) | New | Without | Error shown ✅ |
| Expo 54 (RN 0.81) | New | With | Flow works ✅ |

mliberatore:
Thanks @jeanregisser!

PR #2404, opened 2026-04-20
StripeCryptoOnramp: Improve Mapping Tests

Twigz:
## Summary
Improves mapping tests across iOS/Android for Crypto Onramp.

## Motivation
Improving test coverage to prevent future mapping issues as the API evolves.

## Testing

- [ ] I tested this manually
- [x] I added automated tests

## Documentation

Select one: 


ios/Tests/OnrampMappersTests.swift
```diff
@@ -0,0 +1,511 @@
...
+            traitCollection: UITraitCollection(userInterfaceStyle: .light)
+        )
+        assertColor(
+            appearance.colors?.selectedBorder,
+            matches: "#FF00FF",
+            traitCollection: UITraitCollection(userInterfaceStyle: .dark)
+        )
+    }
+
+    func test_mapToLinkAppearance_partialPrimaryButtonIsIgnored() {
```
jeanregisser, line 68: This documents a cross-platform asymmetry - Android allows partial `primaryButton` configs (just `height` or just `cornerRadius` independently), while iOS silently drops the whole config if either field is missing. Worth addressing separately?

jeanregisser (approved):
Nice! ✨

ios/Tests/OnrampMappersTests.swift
Twigz: Probably, yeah. I'll talk with Michael about it.

mliberatore: Some of this has changed, of course, but it looks like this discrepancy dates back to the original mobile API review, where Android's `PrimaryButton` included optionals, and iOS's `PrimaryButton` did not. https://docs.google.com/document/d/1fiRBeegAqMbaR7OUDUy0EA4idvHEDh_FcBE2DTz2vyo/edit?tab=t.0

We can make a change on iOS to match Android, but that should happen first in the iOS repository, and then adjusted in RN to match the behavior (and tests) in Android.

My recommendation would be to remove this specific iOS test here or merge as-is, then I’ll update the iOS repo, and we'll come back around to this after a native release is made to update the iOS mapping logic and tests.

PR #2418, opened 2026-04-27
fix: wire up ref correctly for PMME

tjclawson-stripe:
## Summary

Accept ref from forwardRef callback and pass to native element

## Motivation

The component used forwardRef but ignored the forwarded ref parameter. Instead, it created its own internal useRef (viewRef) and passed that to the native element. This meant any consumer passing a ref to <PaymentMethodMessagingElement ref={myRef} /> would get nothing — the ref was silently dropped

## Testing

- [X] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/src/components/PaymentMethodMessagingElement.tsx
+++ b/src/components/PaymentMethodMessagingElement.tsx
@@ -1,17 +1,11 @@
-import React, { forwardRef, useEffect, useRef, useState } from 'react';
-import {
-  AccessibilityProps,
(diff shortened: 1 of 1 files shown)


huextrat:
So I'm closing https://github.com/stripe/stripe-react-native/pull/2418? Does this repo no longer accept PR from the community?

porter-stripe (approved):

PR #2425, opened 2026-04-29
Add Checkout SDK iOS bridge

porter-stripe:
## Summary
- Adds the iOS bridge for the Checkout SDK
- Will need an update later after the flattened params for a few API methods gets deployed next week.
- Also removes UPI from the iOS bridge, the latest version of the SDK removed UPI support.

## Motivation
- CheckoutSessions on React Native

## Testing
- Adds some unit tests


example/ios/Podfile.lock
```diff
@@ -1820,14 +1820,14 @@ PODS:
...
     - Yoga
-  - Stripe (25.11.0):
-    - StripeApplePay (= 25.11.0)
-    - StripeCore (= 25.11.0)
-    - StripeIssuing (= 25.11.0)
-    - StripePayments (= 25.11.0)
-    - StripePaymentsUI (= 25.11.0)
-    - StripeUICore (= 25.11.0)
-  - stripe-react-native (0.63.0):
+  - Stripe (25.12.0):
```
jaynewstrom-stripe, line 1823: Can you separate the update PR from the new functionality?

porter-stripe: done!

ios/Tests/CheckoutBridgeTests.swift
```diff
@@ -0,0 +1,120 @@
...
+        XCTAssertEqual(shippingFields["country"] as? String, "GB")
+        XCTAssertEqual(shippingFields["line1"] as? String, "221B Baker Street")
+    }
+
+    func test_mapFromCheckoutState_mapsLoadingStatus() {
+        let mapped = Mappers.mapFromCheckoutState(.loading(MockCheckoutSession()))
+        XCTAssertEqual(mapped["status"] as? String, "loading")
+    }
+
+    private struct MockCheckoutSession: Checkout.Session {
```
jaynewstrom-stripe, line 82: I wouldn't expect anyone outside of our SDK to be able to create a checkout session. Otherwise it'll be near impossible to evolve over time. Can we create this with json instead?

ios/Mappers+Checkout.swift
```diff
@@ -0,0 +1,193 @@
...
+        if let phone = addressUpdate.phone {
+            result["phone"] = phone
+        }
+
+        return result
+    }
+
+    private class func mapFromCheckoutAddress(_ address: Checkout.Address) -> NSDictionary {
+        let result = NSMutableDictionary()
+        result["country"] = address.country
```
jaynewstrom-stripe, line 76: I'm a little conflicted here. In the Android SDK right now, we don't allow access to any fields inside the address object. 

On one hand, that feels right -- the only use case is to pass it to us, it's never consumed. On the other hand, we talked about exposing it back to merchants via the CheckoutSession object -- and the only purpose of that would be for merchants to read it.

ios/StripeSdkImpl+Checkout.swift
```diff
@@ -0,0 +1,308 @@
...
+            }
+
+            do {
+                let checkout = try await Checkout(
+                    clientSecret: clientSecret,
+                    configuration: checkoutConfiguration
+                )
+                let sessionKey = UUID().uuidString
+
+                self.checkoutInstances[sessionKey] = checkout
```
jaynewstrom-stripe, line 38: What does this do? When is it removed?

ios/StripeSdkImpl+Checkout.swift
```diff
@@ -0,0 +1,308 @@
...
+        rejecter reject: @escaping RCTPromiseRejectBlock,
+        operation: @escaping (Checkout) async throws -> Void
+    ) {
+        Task { @MainActor [weak self] in
+            guard let self else {
+                reject(ErrorType.Failed, "Stripe SDK is unavailable", nil)
+                return
+            }
+
+            guard let checkout = self.checkoutInstances[sessionKey] else {
```
jaynewstrom-stripe, line 225: This is probably just a react native thing, and it's probably fine to tell me this is just how it's going to be.

But i'm worried we're going to keep checkout instances around ~forever. But we have PII in there, and I don't think it's cool to just live it in memory forever. When the user is done with it, it should be cleared.

jaynewstrom-stripe (approved):

PR #2441, opened 2026-05-05
Updates Android SDK to 23.8.0 + CryptoOnramp SDK EU Support

mliberatore:
## Summary

This PR, based on #2457, updates the Android SDK, and adds the iOS + Android bindings for CryptoOnramp SDK EU support. These EU additions are exercised in the example app flow changes.

## Motivation

Keeping up-to-date with the latest native SDK releases.

## Testing

- [x] I tested this manually
- [ ] I added automated tests

Tested using the example app. It is expected in testmode that the declaration screen is showing a dummy response (HTML) currently.

| New KYC Fields | Missing Identifiers Returned | Successful Identifier Submission | Declaration Screen | Declaration Acceptance |
| --- | --- | --- | --- | --- |
|  |  |  |  |  |

## Documentation

Select one: 


e2e-tests/checkout-adaptive-pricing.yml
```diff
@@ -87,7 +87,7 @@ appId: ${APP_ID}
     optional: true
 
 - tapOn:
-    text: "Card number"
+    text: "Card number. Supported cards include.*"
```
mliberatore, line 90: Looks like this was a recently added test, and it's failing on `master` for the same reason as those fixed in https://github.com/stripe/stripe-react-native/pull/2457/changes#r3235956032, so fixing here.

src/types/Onramp.ts
```diff
@@ -199,8 +199,117 @@ export type KycInfo = {
...
+      error: StripeError<OnrampError>;
+    };
+
+/**
+ * Result of presenting the CRS/CARF declaration.
+ */
+export type CRSCARFDeclarationResult =
+  | {
+      /** The customer accepted the declaration. */
+      status: 'Confirmed';
```
jeanregisser, line 304: nit: can we use lowercase status? Or is Capitalization a convention we're already using?

```suggestion
      status: 'confirmed';
```

ios/StripeSdkImpl.swift
```diff
@@ -1437,6 +1437,89 @@ public class StripeSdkImpl: NSObject, UIAdaptivePresentationControllerDelegate {
...
+        Task {
+            do {
+                let complianceIdentifiers = try identifierDictionaries.map(Mappers.mapToComplianceIdentifier)
+                let result = try await coordinator.submitIdentifiers(complianceIdentifiers)
+                resolve(Mappers.mapFromSubmitIdentifiersResult(result))
+            } catch {
+                if let identifierError = error as? Mappers.ComplianceIdentifierError,
+                   case let .invalidField(field) = identifierError {
+                    let errorResult = Errors.createError(ErrorType.Unknown, "Invalid format for field: \(field)")
+                    resolve(["error": errorResult["error"]!])
```
jeanregisser, line 1485: Nothing to do now, but this is also where we'll want to apply the richer error proposal that we've been discussing.

example/src/screens/Onramp/CryptoOnrampFlow.tsx
```diff
@@ -1,10 +1,7 @@
-import {
-  Onramp,
-  PlatformPay,
-  useOnramp,
-  useStripe,
-} from '@stripe/stripe-react-native';
+import { PlatformPay, useStripe } from '@stripe/stripe-react-native';
 import type { Address } from '@stripe/stripe-react-native';
+import { useOnramp } from '@stripe/stripe-react-native/src/hooks/useOnramp';
+import { Onramp } from '@stripe/stripe-react-native/src/types';
```
jeanregisser: The imports for `useOnramp` and `Onramp` changed from the public API path (`@stripe/stripe-react-native`) to internal source paths. 
Was this intentional? It means the example no longer demonstrates the import pattern that external consumers would use.

example/src/screens/Onramp/utils.ts
```diff
@@ -1,4 +1,4 @@
-import { Onramp } from '@stripe/stripe-react-native';
+import { Onramp } from '@stripe/stripe-react-native/src/types';
```
jeanregisser: ditto about the import path change

jeanregisser (dismissed):
Awesome! 🇪🇺 🚀

src/types/Onramp.ts
mliberatore: A convention we're already using across Onramp (we have `Consented`, `Denied`, `Confirmed`, `UpdateAddress` statuses). I’m not opposed to us shifting to lowercase across the board, but we should plan that accordingly in isolation and discuss client impact if we'd like to do that.

example/src/screens/Onramp/CryptoOnrampFlow.tsx
mliberatore: These were unneeded, so reverted per your point.

finn-stripe (approved):

PR #2444, opened 2026-05-06
chore: Rewrite publish script in Ruby

tjclawson-stripe:
## Summary

- Rewrite publish script in Ruby
- Add ability to continue from step

## Motivation

- Align with stripe-android and stripe-ios release scripts

## Testing

- [X] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 


scripts/publish.rb
```diff
@@ -0,0 +1,227 @@
+#!/usr/bin/env ruby
```
porter-stripe, line 1: I think we need to update package.json to update the yarn release command.

scripts/publish.rb
```diff
@@ -0,0 +1,227 @@
...
+    execute_or_fail("git push -u origin #{@branch_name} --follow-tags")
+
+    puts ""
+    puts "Tag pushed! Please open a PR for your version bump: https://github.com/stripe/stripe-react-native/pulls"
+    puts ""
+  end
+end
+
+def create_github_release
+  return if @is_dry_run
```
porter-stripe: Could we add this
```
if @is_dry_run
  puts "[dry-run] Would create GitHub release for current version from CHANGELOG.md"
  return
end
```

scripts/publish.rb
```diff
@@ -0,0 +1,227 @@
+#!/usr/bin/env ruby
+
+require 'optparse'
+require 'open3'
+require 'json'
```
porter-stripe, line 5: do we use this anywhere?

scripts/publish.rb
```diff
@@ -0,0 +1,227 @@
...
+  execute_or_fail("yarn install --frozen-lockfile")
+end
+
+def run_tests
+  puts "Running tests"
+  execute_or_fail("yarn run test")
+end
+
+def bump_version
+  @branch_name = "release-branch-#{rand(10000..99999)}"
```
porter-stripe, line 84: I think --continue-from won't work past bump_version. @branch_name is only set here, so if someone retries from step 5 or 6 it'll be nil and the git commands below fail.

scripts/publish.rb
```diff
@@ -0,0 +1,227 @@
...
+  if step_index > 1
+    steps = steps.drop(step_index - 1)
+    rputs "Continuing from step #{step_index}: #{steps.first[:name]}"
+  end
+
+  steps.each_with_index do |step, i|
+    current_step = step_index + i
+    rputs "# #{step[:name]} (step #{current_step}/#{step_count})"
+    step[:action].call
+  rescue => e
```
porter-stripe, line 185: Nit: e is unused. Might be worth printing e.message too.

porter-stripe (commented):
few comments

scripts/publish.rb
tjclawson-stripe: open3 is used in git_status, optparse is used on line 192, and json is used in publish_to_npm

scripts/publish.rb
tjclawson-stripe: Added

scripts/publish.rb
tjclawson-stripe: Updated

scripts/publish.rb
tjclawson-stripe: Updated to detect current branch instead of using branch_name var in publish_to_npm

scripts/publish.rb
tjclawson-stripe: Updated to print error message

porter-stripe (approved):

PR #2456, opened 2026-05-12
Currency Selector Android Layer

porter-stripe:
## Summary
- Adds the Android layer of the AP currency selector

## Motivation
- CheckoutSessions

## Testing
- Will come in a future PR

## Documentation
N/A


android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
```diff
@@ -1593,9 +1603,10 @@ class StripeSdkModule(
       activity.startActivity(chooser)
 
       // Schedule cleanup
+      @Suppress("MagicNumber")
       android.os.Handler(android.os.Looper.getMainLooper()).postDelayed({
         file.delete()
-      }, FILE_CLEANUP_DELAY_MS)
+      }, 3000)
```
tjclawson-stripe: Why can't we use the const val here?

tjclawson-stripe (approved):

PR #2462, opened 2026-05-13
Add propose script

tjclawson-stripe:
## Summary

Add propose script to automatically generate proposal PR
- propose.rb takes version option (major, minor, patch) and update package.json accordingly
- propose.rb will *not* create a git tag
- publish.rb no longer bumps version but creates and pushes git tag
- npm publish now runs on master branch

## Motivation

- Prevent accidentally publishing wrong version
- Ensures package.json version is correct and merged to master
- PCI compliance

## Testing

- [X] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/scripts/propose.rb
+++ b/scripts/propose.rb
@@ -1,28 +1,14 @@
 #!/usr/bin/env ruby
 
 require 'optparse'
-require 'open3'
-require 'json'
 require 'date'
 require 'tmpdir'
 require 'shellwords'
+require_relative 'helpers'
 
 @release_type = nil
 
 VALID_RELEASE_TYPES = %w[patch minor major].freeze
-
-def rputs(string)
-  puts "\e[31m#{string}\e[0m"
-end
-
-def execute_or_fail(command)
-  puts "Executing: #{command}"
-  system(command) or abort "Failed to execute: #{command}"
-end
-
-def current_version
-  JSON.parse(File.read("package.json"))["version"]
-end
 
 def next_version
   parts = current_version.split(".").map(&:to_i)
@@ -120,6 +106,8 @@ unless VALID_RELEASE_TYPES.include?(@release_type)
 end
 
 Dir.chdir(`git rev-parse --show-toplevel`.strip)
+
+preflight_checks
 
 version = next_version
 puts "Proposing #{version} (currently #{current_version})"
--- a/scripts/propose.rb
+++ b/scripts/propose.rb
@@ -1,0 +1,121 @@
+#!/usr/bin/env ruby
+
+require 'optparse'
+require 'open3'
+require 'json'
+require 'date'
+require 'tmpdir'
+require 'shellwords'
+
+@release_type = nil
+
+VALID_RELEASE_TYPES = %w[patch minor major].freeze
+
+def rputs(string)
+  puts "\e[31m#{string}\e[0m"
+end
+
+def execute_or_fail(command)
+  puts "Executing: #{command}"
+  system(command) or abort "Failed to execute: #{command}"
+end
+
+def current_version
+  JSON.parse(File.read("package.json"))["version"]
+end
+
+def next_version
+  parts = current_version.split(".").map(&:to_i)
+  case @release_type
+  when "major"
+    parts[0] += 1
+    parts[1] = 0
+    parts[2] = 0
+  when "minor"
+    parts[1] += 1
+    parts[2] = 0
+  when "patch"
+    parts[2] += 1
+  end
+  parts.join(".")
+end
+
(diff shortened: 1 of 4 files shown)


scripts/propose.rb
```diff
@@ -0,0 +1,126 @@
...
+  end
+
+  File.write("CHANGELOG.md", changelog.sub("## Unreleased", header))
+end
+
+def bump_version(version)
+  execute_or_fail("yarn version --no-git-tag-version --new-version #{version}")
+end
+
+def create_proposal_pr(version)
```
porter-stripe, line 44: Can we add a check that the users tree is not dirty? If a user has uncommitted changes, git checkout -b will carry them into the new branch, and they'll get committed along with the version bump

scripts/propose.rb
```diff
@@ -0,0 +1,126 @@
...
+  branch = "release/propose-#{version}"
+  execute_or_fail("git checkout -b #{branch}")
+
+  bump_version(version)
+  update_changelog(version)
+
+  execute_or_fail("git add package.json CHANGELOG.md")
```
porter-stripe, line 51: Probably fine, but are we sure no files will ever need to be included?

scripts/propose.rb
```diff
@@ -0,0 +1,126 @@
...
+    Propose #{version}
+
+    - [x] Ensure the CHANGELOG is up to date with all relevant commits since the last release
+    - [x] Add the version number for this release & the date to the CHANGELOG, underneath "## Unreleased"
+      - e.g. "## 1.2.3 - 2022-02-14"
+    - [x] Update the README if necessary (this is only required when there are breaking changes in the release, such as dropping support for an iOS || Android version)
+  BODY
+
+  puts ""
+  pr_url = `hub pull-request --base master --head #{branch} -F #{pr_message_file.shellescape}`.strip
```
porter-stripe, line 66: I think hub is deprecated in favor of GitHub CLI (gh)

scripts/publish.rb
```diff
@@ -80,40 +81,34 @@ def run_tests
...
-  @branch_name = "release-branch-#{rand(10000..99999)}"
-  execute_or_fail("git checkout -b #{@branch_name}")
-
-  puts "Bumping package.json #{@release_type} version and tagging commit"
-  execute_or_fail("yarn version --#{@release_type}")
-end
-
-def current_branch
-  @branch_name || `git rev-parse --abbrev-ref HEAD`.strip
+def create_git_tag
```
porter-stripe, line 19: If someone re-runs publish.rb after a partial failure, git tag will fail because the tag already exists. Can we handle this gracefully?

scripts/publish.rb
```diff
@@ -147,7 +142,7 @@ def create_github_release
     return
   end
 
-  current_version = version_and_date.split(" -").first.strip
+  current_ver = version_and_date.split(" -").first.strip
```
porter-stripe: nit can we keep this name current_version.

scripts/publish.rb
```diff
@@ -29,6 +26,10 @@ def git_status
   stdout
 end
 
+def current_version
+  JSON.parse(File.read("package.json"))["version"]
+end
+
 def ensure_on_master
```
porter-stripe: is this ever called?

porter-stripe (commented):
few comments

scripts/propose.rb
tjclawson-stripe: Will address in follow up https://jira.corp.stripe.com/browse/MOBILESDK-4534

scripts/propose.rb
tjclawson-stripe: Updated

tjclawson-stripe: Added helper.rb to re-use preflight checks for both scripts

scripts/publish.rb
tjclawson-stripe: Removed to just call the current_version helper method instead

scripts/publish.rb
tjclawson-stripe: Called in preflight_checks which has been moved to helpers.rb

scripts/propose.rb
tjclawson-stripe: I can't think of any reason we'd add other files, but if we need to in the future it's probably best to be explicit instead of `add .`

scripts/publish.rb
tjclawson-stripe: They should be able to re-run from the step that failed, meaning this will be skipped if e.g. publish is what failed

porter-stripe (approved):

PR #2475, opened 2026-05-20
Makes LinkAppearance.PrimaryButtonConfiguration Mapping Consistent With Android [Includes iOS SDK 25.16.0 Update]

mliberatore:
## Summary

Depends on https://github.com/stripe/stripe-ios/pull/6510 and a subsequent iOS SDK update.

Previously, unspecified values on `init` made `PrimaryButtonConfiguration`'s properties default to `.zero`. This led to a mismatch in mapping behavior since the properties couldn’t be `nil`, and individually apply corner radius or height, falling back to Link  defaults for the other.

## Motivation
This comment: https://github.com/stripe/stripe-react-native/pull/2404#discussion_r3134337573

## Testing

- [x] I tested this manually
- [x] I added automated tests

Ensured we were able to set each value independently while relying on a default for the other. I used intentionally obvious values to ensure they were being applied, and differed from sensible defaults.

| Corner radius set only | Height set only | Both set | Neither set (both defaults) |
| --- | --- | --- | --- |
|  |  |  |  |
## Documentation

Select one: 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## X.Y.Z - changes pending release
+**Fixes**
+* [Fixed] Matched iOS Onramp `LinkAppearance.PrimaryButtonConfiguration` mapping to Android so unspecified custom height and corner radius values use Link defaults.
+
 ## 0.66.0 - 2026-05-20
 **Changes**
 * Updated Stripe iOS SDK from 25.14.0 to 25.15.0.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## X.Y.Z - changes pending release
 **Fixes**
-* Updated Stripe iOS SDK from 25.14.0 to 25.16.0.
+* Updated Stripe iOS SDK from 25.15.0 to 25.16.0.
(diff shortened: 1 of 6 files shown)


e2e-tests/ios-only/checkout-adaptive-pricing.yml
```diff
@@ -42,7 +42,7 @@ appId: ${APP_ID}
     centerElement: true
     timeout: 30000
 - tapOn:
-    text: ".*\\$.*"
+    id: "currency_option_usd"
```
mliberatore, line 45: A design change in the iOS SDK causes this button to not be found, which causes the test to fail. It changed from listing the price with "$" to now read "USD", so adapting the test to use the accessibility identifier instead.

The PR that changed the design: https://github.com/stripe/stripe-ios/pull/6460/changes#diff-24f7f8c93c56eba450aeebf7b002a6b571ef3c39d754417e596f117da43bd64f

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## X.Y.Z - changes pending release
+**Fixes**
+* Updated Stripe iOS SDK from 25.14.0 to 25.16.0.
```
jeanregisser:
```suggestion
* Updated Stripe iOS SDK from 25.15.0 to 25.16.0.
```

jeanregisser (dismissed):
🚀

jeanregisser (approved):

PR #2485, opened 2026-06-08
iOS 25.17.0, Android: 23.10.+ SDK Updates + CryptoOnramp EU Updates and Better Error Reporting

mliberatore:
## Summary
- Updates to the latest native SDK code changes.
- Adapts to the latest iOS and Android CryptoOnramp SDK changes, including EU-related changes and richer errors for attestation-related failures.
- Renames the CRS/CARF declarations presentation API wrapper per request from the Crypto team. 

## Motivation
To keep the RN SDK up-to-date with the native SDK versions

## Testing

- [ ] I tested this manually
- [x] I added automated tests

## Documentation

Select one: 


porter-stripe:
@mliberatore new SDKs versions are live! This is blocking some work on our end, can we get this merged?

mliberatore:
@porter-stripe still working on it at the moment. It appears there's a failure when running `yarn test-ios e2e-tests/ios-only/checkout-adaptive-pricing.yml` with PaymentSheet not presenting when it's expected to, unrelated to the CryptoOnramp changes that we're adapting. Any help would be appreciated! 
CleanShot 2026-06-09 at 11 47 19

android/src/onramp/java/com/reactnativestripesdk/OnrampErrors.kt
```diff
@@ -0,0 +1,144 @@
...
+  val apiException =
+    error as? CryptoOnrampApiException
+      ?: return createNonApiOnrampError(code, error)
+
+  val stripeError = apiException.stripeError
+  val onrampErrorType =
+    when (apiException) {
+      is AppAttestationException -> "AppAttestationError"
+      is UncategorizedApiErrorException -> "UncategorizedApiError"
+      else -> null
```
jeanregisser: Should this fall back to `"UncategorizedApiError"` instead of `null`?

iOS maps any unknown `StripeCryptoOnrampAPIError` subtype to `"UncategorizedApiError"`, and the RN type uses `onrampErrorType` as the discriminator for API rich errors.

If Android receives a future `CryptoOnrampApiException` subtype, sending `null` here makes the runtime shape inconsistent with both iOS and the TS type, even though the rest of the API diagnostics are present.

android/src/test/java/com/reactnativestripesdk/OnrampErrorsTest.kt
```diff
@@ -0,0 +1,165 @@
...
+        appPackageName = appPackageName,
+        mode = mode,
+        apiErrorCode = apiErrorCode,
+        apiErrorType = apiErrorType,
+        apiErrorMessage = apiErrorMessage,
+        apiUserMessage = apiUserMessage,
+        docUrl = docUrl,
+        underlyingError = cause,
+      )
+    val constructor =
```
jeanregisser, line 151: nit: Since these native error classes are not directly constructible here, the reflection helper makes sense.

Could we add a short comment explaining that? Otherwise a future native SDK constructor change will fail this test with `NoSuchMethodException`, and it will not be obvious why reflection is being used instead of normal construction.

jeanregisser (commented):
Thanks for pushing this through. I left a few focused comments; the main one is the Android configure path still using the generic mapper, which looks like it can drop the new rich error fields.

jeanregisser:
One more Android configure-path note that GitHub would not let me anchor directly on the unchanged line:

In `android/src/onramp/java/com/reactnativestripesdk/OnrampSdkModule.kt`, `OnrampConfigurationResult.Failed` still resolves through the generic mapper:

```kotlin
promise.resolve(createError(ErrorType.Failed.toString(), configureResult.error))
```

That drops the new Onramp rich fields (`onrampErrorType`, `developerMessage`, `operation`, `sdkVersions`, API context, etc.) for configure failures, while the other Onramp failure paths use `createOnrampFailedError(...)`.

Should this be:

```kotlin
promise.resolve(createOnrampFailedError(configureResult.error))
```

This matters for attestation/configure failures.

jeanregisser (approved):
Looks good to me now 🚢. Thanks for the quick Android error handling tweaks.

One future follow-up: the RN error type direction here feels like a good incremental step, but probably not the final shape we’ll want long-term. At some point we should make CryptoOnrampError mirror the native hierarchy more directly: a base SDK-level rich error, plus an API-error subtype for API context fields. This PR gets the API-error typing in place while keeping existing StripeError consumers working, so I think it’s a good place to land and iterate from.

PR #2495, opened 2026-06-16
CryptoOnramp SDK: Splits Rich Error Types (SDK vs. API)

mliberatore:
## Summary
To more closely match the native SDKs, we're splitting the `OnrampApiError` type to have a base `OnrampSdkError` that covers non-API errors.

This also adds the non-API attestation failure that occurs on `create` from the native SDKs, so we remove the temporary shim from #2486.

## Motivation

https://github.com/stripe/stripe-react-native/pull/2485#pullrequestreview-4462529728

## Testing

- [ ] I tested this manually
- [x] I added automated tests

## Documentation

Select one: 


jeanregisser (dismissed):
Thanks! 🚀

mliberatore:
Moved back to draft as we address the changes that occurred in https://github.com/stripe/stripe-ios/pull/6588 and
https://github.com/stripe/stripe-android/pull/13240

src/types/Onramp.ts
```diff
@@ -268,23 +268,43 @@ export type ComplianceIdentifierRequirements = {
...
+  | OnrampApiErrorType
+  | 'AppAttestationUnavailableError';
 
-export type OnrampApiError = StripeError<OnrampError> & {
-  onrampErrorType: string;
+/**
+ * Base rich error shape returned by native Crypto Onramp SDK errors.
+ */
+export type OnrampSdkError = StripeError<OnrampErrorStatus> & {
+  onrampErrorType: OnrampErrorType;
```
finn-stripe, line 289: In `OnrampErrors.kt` it looks like `onrampErrorType` is null or `AppAttestationUnavailableError`, but here seems like it just needs to be one of the three values (`AppAttestationError`, `UncategorizedApiError`, `AppAttestationUnavailableError`). So it seems like you could have `onrampErrorType` as null with `developerMessage` and `userMessage` still being set which doesn't seem to fit any of the shapes of `CryptoOnrampError` below.

Maybe I am misunderstanding something with how these types / the logic in `OnrampErrors.kt` though?

finn-stripe (commented):
one question, might just be due to my lack of understanding of Android. Once that is resolved I can approve!

src/types/Onramp.ts
Twigz: You're roughly understanding correctly, however it was done this way intentionally, as when new errors are added; we must add the new types as well. Alternatively we could use an unrecognized type, but given that this particular error set is for SDK errors, and not API errors, we'll always have a known error type, we just have to add new ones as they exist.

We'll think more on how to automatically support new errors though, as right now it's a manual process when SDK updates happen.

finn-stripe (approved):

PR #2502, opened 2026-06-23
Add kycRecipientAccountId internal prop to ConnectAccountOnboarding

matebek-stripe:
## Summary

Expose an internal-only prop `kycRecipientAccountId` on the [AccountOnboarding](https://docs.stripe.com/connect/supported-embedded-components/account-onboarding?platform=react-native) component that allows us to onboard recipients for money-movement actions initiated between accounts.

This property shouldn't be available for external customers, so I've marked it with `@internal` as I've observed it being used to annotate other properties in the codebase.

## Testing

- [ ] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/etc/stripe-react-native.api.md
+++ b/etc/stripe-react-native.api.md
@@ -4229,9 +4229,9 @@ interface WeChatPayParams_2 {
 // Warnings were encountered during analysis:
 //
 // src/components/CustomerSheet.tsx:374:27 - (ae-forgotten-export) The symbol "Component" needs to be exported by the entry point index.d.ts
-// src/connect/Components.tsx:74:3 - (ae-forgotten-export) The symbol "StepChange" needs to be exported by the entry point index.d.ts
-// src/connect/Components.tsx:78:3 - (ae-forgotten-export) The symbol "CollectionOptions" needs to be exported by the entry point index.d.ts
-// src/connect/Components.tsx:243:3 - (ae-forgotten-export) The symbol "PaymentsListDefaultFilters" needs to be exported by the entry point index.d.ts
+// src/connect/Components.tsx:75:3 - (ae-forgotten-export) The symbol "StepChange" needs to be exported by the entry point index.d.ts
+// src/connect/Components.tsx:79:3 - (ae-forgotten-export) The symbol "CollectionOptions" needs to be exported by the entry point index.d.ts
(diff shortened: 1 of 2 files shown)


maragues-stripe (dismissed):
Approving to unblock this.

Security wise, `@internal` / lack of docs / any RN-side guard should not be treated as a security boundary, since external integrators can still see or patch the package. The real control needs to be backend authorization: the connected account/platform context must be authorized to act on the supplied recipient account. From the code I checked, the relevant V2 operations go through IAM/resource checks

One follow-up I'd recommend: consider adding a small Dashboard-app-only guard for this prop as API hygiene, with a comment making clear it is not an auth boundary. That would reduce accidental external usage, but security should continue to rely on backend enforcement. Something like

```typescript
const STRIPE_DASHBOARD_APP_IDS = new Set([
    'com.stripe.dashboard',
    'com.stripe.android.dashboard',
    'com.stripe.dashboard.rn',
    'com.stripe.android.dashboard.rn',
  ]);

  const sanitizedComponentProps = useMemo(() => {
    if (
      component === 'account-onboarding' &&
      componentProps?.setKycRecipientAccountId &&
      !STRIPE_DASHBOARD_APP_IDS.has(String(overrides?.applicationId ?? ''))
    ) {
      const {setKycRecipientAccountId, ...rest} = componentProps;
      return rest;
    }

    return componentProps;
  }, [component, componentProps, overrides?.applicationId]);  
```

Then replace uses of componentProps with sanitizedComponentProps for the WebView bridge

etc/stripe-react-native.api.md
```diff
@@ -1232,6 +1232,7 @@ export function ConnectAccountOnboarding(input: {
     fullTermsOfServiceUrl?: string;
     privacyPolicyUrl?: string;
     collectionOptions?: CollectionOptions;
+    kycRecipientAccountId?: string;
```
jorgea-stripe: How are we preventing connect platforms from using this?

jorgea-stripe (commented):
r? @matebek-stripe

maragues-stripe (dismissed):
Thanks!

nit: add a brief comment above the any cast explaining that this is an intentionally hidden internal prop and not an auth boundary

maragues-stripe (approved):
💯

PR #2533, opened 2026-07-14
LinkControllerPreview APIs

mats-stripe:
## Testing

- [ ] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/LinkControllerManager.kt
+++ b/android/src/main/java/com/reactnativestripesdk/LinkControllerManager.kt
@@ -2,6 +2,7 @@ package com.reactnativestripesdk
 
 import androidx.activity.ComponentActivity
 import androidx.lifecycle.SavedStateHandle
+import androidx.lifecycle.lifecycleScope
 import com.facebook.react.bridge.Arguments
 import com.facebook.react.bridge.Promise
 import com.facebook.react.bridge.ReactApplicationContext
@@ -11,7 +12,6 @@ import com.reactnativestripesdk.utils.createError
 import com.reactnativestripesdk.utils.mapFromPaymentMethod
 import com.stripe.android.link.LinkController
 import com.stripe.android.link.LinkControllerPreview
-import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
@@ -78,7 +78,7 @@ internal class LinkControllerManager(
 
         val controller = linkController!!
 
-        CoroutineScope(Dispatchers.Main).launch {
+        activity.lifecycleScope.launch {
             val result = controller.configure(config)
             if (result.isSuccess) {
                 // Create once: presenter may register ActivityResultLaunchers, which must happen before onStart().
@@ -137,8 +137,9 @@ internal class LinkControllerManager(
                 response.putMap("paymentMethod", mapFromPaymentMethod(result.paymentMethod))
 
                 val preview = controller.paymentMethodPreview.value
-                if (preview != null) {
-                    CoroutineScope(Dispatchers.Main).launch {
+                val currentScope = (context.currentActivity as? ComponentActivity)?.lifecycleScope
+                if (preview != null && currentScope != null) {
+                    currentScope.launch {
                         val iconBase64 = withContext(Dispatchers.IO) {
                             try {
                                 convertDrawableToBase64(preview.icon)
--- a/android/src/main/java/com/reactnativestripesdk/LinkControllerManager.kt
+++ b/android/src/main/java/com/reactnativestripesdk/LinkControllerManager.kt
@@ -1,0 +1,163 @@
+package com.reactnativestripesdk
+
+import androidx.activity.ComponentActivity
+import androidx.lifecycle.SavedStateHandle
+import com.facebook.react.bridge.Arguments
+import com.facebook.react.bridge.Promise
+import com.facebook.react.bridge.ReactApplicationContext
+import com.facebook.react.bridge.ReadableMap
+import com.reactnativestripesdk.utils.ErrorType
+import com.reactnativestripesdk.utils.createError
+import com.reactnativestripesdk.utils.mapFromPaymentMethod
+import com.stripe.android.link.LinkController
+import com.stripe.android.link.LinkControllerPreview
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+
+internal const val LINK_CONTROLLER_NOT_INITIALIZED_ERROR =
+    "LinkController has not been initialized. Call initLinkController first."
+
+/**
+ * Manages [LinkController] initialization and presentation for the React Native bridge.
+ *
+ * @PrivatePreview
+ */
+@OptIn(LinkControllerPreview::class)
+internal class LinkControllerManager(
+    private val context: ReactApplicationContext,
+    private val publishableKey: String,
+    private val stripeAccountId: String?,
+) {
+    private var linkController: LinkController? = null
+
+    fun configure(params: ReadableMap, promise: Promise) {
+        val email = params.getString("email")
+        val merchantDisplayName = params.getString("merchantDisplayName")
+
+        if (email == null || merchantDisplayName == null) {
+            promise.resolve(createError(ErrorType.Failed.toString(), "email and merchantDisplayName are required."))
+            return
+        }
+
+        val activity = context.currentActivity as? ComponentActivity
+        if (activity == null) {
+            promise.resolve(createError(ErrorType.Failed.toString(), "Activity is not available. Retry this method."))
+            return
+        }
+
+        val phoneNumber = params.getString("phoneNumber")
+        val allowLogout = if (params.hasKey("allowLogout")) params.getBoolean("allowLogout") else true
+        val setupIntentClientSecret = params.getString("setupIntentClientSecret")
+
+        val supportedTypes: List<LinkController.PaymentMethodType>? =
+            params.getArray("supportedPaymentMethodTypes")?.let { arr ->
+                (0 until arr.size()).mapNotNull { i ->
+                    when (arr.getString(i)) {
+                        "card" -> LinkController.PaymentMethodType.Card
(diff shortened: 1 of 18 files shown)


android/src/main/java/com/reactnativestripesdk/LinkControllerManager.kt
```diff
@@ -0,0 +1,163 @@
...
+
+        val supportedTypes: List<LinkController.PaymentMethodType>? =
+            params.getArray("supportedPaymentMethodTypes")?.let { arr ->
+                (0 until arr.size()).mapNotNull { i ->
+                    when (arr.getString(i)) {
+                        "card" -> LinkController.PaymentMethodType.Card
+                        "bankAccount" -> LinkController.PaymentMethodType.BankAccount
+                        else -> null
+                    }
+                }.ifEmpty { null }
```
tillh-stripe: This line makes Android pass `null` for an empty list while iOS will pass `[]`. They resolve to the same, but just pointing it out.

android/src/main/java/com/reactnativestripesdk/LinkControllerManager.kt
```diff
@@ -0,0 +1,163 @@
...
+    private var linkController: LinkController? = null
+
+    fun configure(params: ReadableMap, promise: Promise) {
+        val email = params.getString("email")
+        val merchantDisplayName = params.getString("merchantDisplayName")
+
+        if (email == null || merchantDisplayName == null) {
+            promise.resolve(createError(ErrorType.Failed.toString(), "email and merchantDisplayName are required."))
+            return
+        }
```
tillh-stripe: This same logic is missing on iOS, no?

tillh-stripe: Or rather: It's happening on presentation there, not on configuration.

android/src/main/java/com/reactnativestripesdk/LinkControllerManager.kt
```diff
@@ -0,0 +1,163 @@
...
+
+        val controller = linkController
+        if (controller == null) {
+            promise.resolve(createError(ErrorType.Failed.toString(), LINK_CONTROLLER_NOT_INITIALIZED_ERROR))
+            return
+        }
+
+        val presenter = controller.createPresenter(
```
lng-stripe: Should we be creating and caching the presenter in `configure()` instead of re-creating in `present()` each time?

android/src/main/java/com/reactnativestripesdk/LinkControllerManager.kt
```diff
@@ -0,0 +1,163 @@
...
+internal const val LINK_CONTROLLER_NOT_INITIALIZED_ERROR =
+    "LinkController has not been initialized. Call initLinkController first."
+
+/**
+ * Manages [LinkController] initialization and presentation for the React Native bridge.
+ *
+ * @PrivatePreview
+ */
+@OptIn(LinkControllerPreview::class)
+internal class LinkControllerManager(
```
lng-stripe, line 28: Should this extend `StripeUIManager` like `CustomerSheetManager` does? That base class seems to address some Android lifecycle edge cases.

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
```diff
@@ -2000,6 +2001,33 @@ class StripeSdkModule(
...
     }
 
+  // LinkController - Private Preview
+
+  @ReactMethod
+  override fun initLinkController(
+    params: ReadableMap,
+    promise: Promise,
+  ) {
+    if (linkControllerManager == null) {
```
lng-stripe: [🤖]  **`linkControllerManager` survives `initStripe()` re-init**

`initialise()` updates `publishableKey` but never nulls `linkControllerManager`. The `if (linkControllerManager == null)` guard is false on subsequent calls, so Link API calls silently use the old key. `customerSheetManager` is explicitly replaced — same pattern is needed here.

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
```diff
@@ -2000,6 +2001,33 @@ class StripeSdkModule(
       }
     }
 
+  // LinkController - Private Preview
+
+  @ReactMethod
+  override fun initLinkController(
```
lng-stripe, line 2009: [🤖] `linkControllerManager` not cleaned up in `invalidate()`

Not registered in `stripeUIManagers` (unlike `customerSheetManager`), so `invalidate()` skips it. After bridge teardown, in-flight coroutines hold a stale context and crash when they call `promise.resolve()`.

android/src/main/java/com/reactnativestripesdk/LinkControllerManager.kt
```diff
@@ -0,0 +1,163 @@
...
+        if (linkController == null) {
+            linkController = LinkController.Builder(
+                application = activity.application,
+                savedStateHandle = SavedStateHandle()
+            ).build()
+        }
+
+        val controller = linkController!!
+
+        CoroutineScope(Dispatchers.Main).launch {
```
lng-stripe: [🤖] Unmanaged `CoroutineScope` in `configure()` and the icon-download lambda

Both use `CoroutineScope(Dispatchers.Main).launch` with no lifecycle owner. If the Activity is destroyed mid-flight, the resumed coroutine crashes. Fix: use `activity.lifecycleScope.launch` — `activity` is already in scope.

android/src/main/java/com/reactnativestripesdk/LinkControllerManager.kt
```diff
@@ -0,0 +1,163 @@
...
+            activity = activity,
+            presentCallback = { result ->
+                when (result) {
+                    is LinkController.PresentResult.Completed -> {
+                        val response = Arguments.createMap()
+                        response.putMap("paymentMethod", mapFromPaymentMethod(result.paymentMethod))
+
+                        val preview = controller.paymentMethodPreview.value
+                        if (preview != null) {
+                            CoroutineScope(Dispatchers.Main).launch {
```
lng-stripe: See other comment re: scope

lng-stripe (dismissed):
Could you add this flow to the example app?

In addition to the Android issues raised by agents that I was able to verify, there were iOS issues that I'm less familiar with. Posting below in case useful:

---

**5. CONFIRMED — Empty `[]` passed to `LinkConfiguration.supportedPaymentMethodTypes` on iOS** (`StripeSdkImpl+LinkController.swift:24`)
If all input strings are unrecognized, `compactMap` returns `[]` (not `nil`). `LinkConfiguration` interprets `[]` as "allow zero payment methods", showing an empty sheet. Android uses `.ifEmpty { null }` — iOS needs the same guard.

**6. CONFIRMED — Malformed data URI `"data:image/png;base64,"` when `pngData()` returns nil** (`StripeSdkImpl+LinkController.swift:101`)
Vector/PDF-backed `UIImage`s have no PNG representation. The fallback `?? ""` produces a broken URI silently sent to JS, rendering a blank icon. Should omit the `icon` key when `pngData()` is nil, matching Android's behavior.

**7. CONFIRMED — iOS skips `email` / `merchantDisplayName` validation at init time** (`StripeSdkImpl+LinkController.swift:18`)
Android rejects immediately if either is nil. iOS stores nil silently, resolves init with `{}`, then fails later at `presentLinkController` with a confusing error. Validation belongs at `initLinkController`.

mats-stripe:
>5. CONFIRMED — Empty [] passed to LinkConfiguration.supportedPaymentMethodTypes on iOS (StripeSdkImpl+LinkController.swift:24)
If all input strings are unrecognized, compactMap returns [] (not nil). LinkConfiguration interprets [] as "allow zero payment methods", showing an empty sheet. Android uses .ifEmpty { null } — iOS needs the same guard.

Both `[]` and `nil` are interpreted as "show all payment method types". We can make the same guard on iOS, but there will be no behavior change

>6. CONFIRMED — Malformed data URI "data:image/png;base64," when pngData() returns nil (StripeSdkImpl+LinkController.swift:101)
Vector/PDF-backed UIImages have no PNG representation. The fallback ?? "" produces a broken URI silently sent to JS, rendering a blank icon. Should omit the icon key when pngData() is nil, matching Android's behavior.

This one's worth looking into!

>7. CONFIRMED — iOS skips email / merchantDisplayName validation at init time (StripeSdkImpl+LinkController.swift:18)
Android rejects immediately if either is nil. iOS stores nil silently, resolves init with {}, then fails later at presentLinkController with a confusing error. Validation belongs at initLinkController.

"Android rejects immediately if either is nil." is not the correct behavior, these parameters should be optional. The merchant doesn't need to pass in a email if they don't have one. We should make sure that behavior is working as expected.

android/src/main/java/com/reactnativestripesdk/LinkControllerManager.kt
lng-stripe: Turns out no, it shouldn't.

mats-stripe:
>>CONFIRMED — Malformed data URI "data:image/png;base64," when pngData() returns nil (StripeSdkImpl+LinkController.swift:101)
Vector/PDF-backed UIImages have no PNG representation. The fallback ?? "" produces a broken URI silently sent to JS, rendering a blank icon. Should omit the icon key when pngData() is nil, matching Android's behavior.

>This one's worth looking into!

Following up on this one: I don't think there's an issue here, the RN demo app on iOS is able to display the correct icon: 

Simulator Screenshot - iPhone 16 Pro Max - 2026-07-14 at 18 50 23

android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt
mats-stripe: Done in https://github.com/stripe/stripe-react-native/pull/2533/commits/6aef0b3f7b4259f5b1a1d70affc26819569a6546

android/src/main/java/com/reactnativestripesdk/LinkControllerManager.kt
mats-stripe: Both of these fields are optional on iOS. Email is also optional on Android, so I made email an optional field here: https://github.com/stripe/stripe-react-native/pull/2533/commits/d60aa9e2542c523c52f968b23e161e737eadb00b.  merchantDisplayName is non optional on Android, so I kept it non-optional at the RN level.

tillh-stripe (approved):

PR #2534, opened 2026-07-14
Update README with React Native compatibility info

RN-Akash:
Added React Native compatibility table for Stripe SDK versions.


RN-Akash (commented):
Update README with React Native compatibility info

PR #2543, opened 2026-07-21
Bump to latest Stripe SDKs & update `useLinkController`

mats-stripe:
## Summary

* Updated Stripe iOS SDK from 26.3.0 to 26.4.1.
* Updated Stripe Android SDK from 23.12.0 to 23.13.1.
* Update the (private preview) `useLinkController` hook to the latest API, which splits out the setup intent confirmation step into its own API, and adds a field in the LinkConfiguration for specifying the payment method types used. 

## Motivation

Latest SDK version contains new APIs and bug fixes.

## Testing

Manually tested

## Documentation

Select one: 


lng-stripe (dismissed):
Seems we'll need to fix Android analytics first:
 * broke [here](https://github.com/stripe/stripe-android/pull/13500/changes#diff-13c8ed8f0c72707d6528b0fd6338c57ff05d3121f60bac354a164f105614c8ecL24-L30)
 * makes analytics value always `"crypto_onramp"` ([1](https://github.com/stripe/stripe-android/blob/1ea2bf66b0f991fe004be928353678467da6bd45/paymentsheet/src/main/java/com/stripe/android/paymentsheet/state/DefaultAnalyticsMetadataFactory.kt#L247-L247), [2](https://github.com/stripe/stripe-android/blob/1ea2bf66b0f991fe004be928353678467da6bd45/paymentsheet/src/main/java/com/stripe/android/paymentsheet/state/DefaultAnalyticsMetadataFactory.kt#L83))
 * this analytics map is used everywhere in DefaultEventReporter ([code](https://github.com/stripe/stripe-android/blob/1ea2bf66b0f991fe004be928353678467da6bd45/paymentsheet/src/main/java/com/stripe/android/paymentsheet/analytics/DefaultEventReporter.kt#L599-L599))

android/src/main/java/com/reactnativestripesdk/LinkControllerManager.kt
```diff
@@ -110,6 +109,37 @@ internal class LinkControllerManager(
...
+        val activity = context.currentActivity as? ComponentActivity
+        val scope = activity?.lifecycleScope
+        if (scope == null) {
+            promise.resolve(createError(ErrorType.Failed.toString(), "Activity is not available. Retry this method."))
+            return
+        }
+
+        scope.launch {
+            val confirmResult = presenter.confirmSetupIntent(clientSecret)
+            when (confirmResult) {
```
lng-stripe: This needs fixing. Follow-up: we should run android-lint in CI to catch these issues.

lng-stripe (changes requested):

patches/old-arch-codegen-fix.patch
```diff
@@ -1,101 +1,2 @@
...
-+    callback: (event: UnsafeObject<any>) => void
-+  ): void;
-+  onCustomPaymentMethodConfirmHandlerCallback(
-+    callback: (event: UnsafeObject<any>) => void
-+  ): void;
- }
-
- export default TurboModuleRegistry.getEnforcing<Spec>('StripeSdk');
+--- src/specs/NativeStripeSdkModule.ts.orig
++++ src/specs/NativeStripeSdkModule.ts
```
lng-stripe: This file is now malformed (it's basically empty). See https://github.com/stripe/stripe-react-native/blob/master/CONTRIBUTING.md#how-to-update-the-patch

lng-stripe (changes requested):

patches/old-arch-codegen-fix.patch
mats-stripe: Reverted to master since it's not relevant to this PR, but this file is dead code and should just be deleted in a follow-up.

lng-stripe (approved):
👍

patches/old-arch-codegen-fix.patch
```diff
@@ -1,2 +1,101 @@
---- src/specs/NativeStripeSdkModule.ts.orig
-+++ src/specs/NativeStripeSdkModule.ts
+--- src/specs/NativeStripeSdkModule.ts.orig	2025-07-09 10:31:12.373374861 -0600
```
lng-stripe, line 1: This still doesn't patch successfully, though the issue exists on `master` so we can fix separately.

PR #2570, opened 2026-08-12
StripeCryptoOnramp: Implements Samsung Pay

Twigz:
## Summary
Adds SamsungPay implementation to the SDK and the example app.

## Motivation
Updates the RN library to contain Samsung Pay, as the Android SDK now contains it.

## Testing

- [x] I tested this manually
- [x] I added automated tests

You can follow the instructions in the libs/README.md for testing.

## Documentation

Select one: 

### DIFF
--- a/android/src/onramp/java/com/reactnativestripesdk/OnrampMappers.kt
+++ b/android/src/onramp/java/com/reactnativestripesdk/OnrampMappers.kt
@@ -8,6 +8,8 @@ import com.facebook.react.bridge.Arguments
 import com.facebook.react.bridge.ReadableArray
 import com.facebook.react.bridge.ReadableMap
 import com.facebook.react.bridge.WritableArray
+import com.reactnativestripesdk.utils.getIntegerList
+import com.reactnativestripesdk.utils.mapToPreferredNetworks
 import com.stripe.android.core.model.CountryCode
 import com.stripe.android.crypto.onramp.ExperimentalCryptoOnramp
 import com.stripe.android.link.LinkControllerPreview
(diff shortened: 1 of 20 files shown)


android/src/onramp/java/com/reactnativestripesdk/OnrampSdkModule.kt
```diff
@@ -193,6 +203,17 @@ class OnrampSdkModule(
     }
   }
 
+  @ReactMethod
+  override fun isSamsungPaySupported(promise: Promise) {
+    rnScope.launch {
+      val isSupported =
+        withTimeoutOrNull(15_000L) {
```
jeanregisser: Curious, why 15 secs?

android/src/onramp/java/com/reactnativestripesdk/OnrampSdkModule.kt
```diff
@@ -88,6 +88,7 @@ class OnrampSdkModule(
   private var checkoutPromise: Promise? = null
   private var verifyKycPromise: Promise? = null
   private var userAttestationPromise: Promise? = null
+  private var samsungPayAvailability = CompletableDeferred<Boolean>()
```
jeanregisser: Since this starts as a pending `CompletableDeferred`, calling `isSamsungPaySupported()` before `configureOnramp` runs will block on `await()` for the full 15s timeout before resolving `false`.

Could we initialize it to `CompletableDeferred(false)` instead? `configureOnramp` already reassigns it for both the configured and null cases, so the only behavior change is that the "not configured yet" path returns immediately.

android/src/onramp/java/com/reactnativestripesdk/OnrampMappers.kt
```diff
@@ -51,20 +54,88 @@ internal fun mapConfig(
...
+@SuppressLint("RestrictedApi")
+internal fun mapSamsungPayConfig(params: ReadableMap?): OnrampConfiguration.SamsungPayConfig? {
+  if (params == null) return null
+
+  val serviceId = params.getString("serviceId")?.takeIf { it.isNotBlank() } ?: return null
+  val merchantId = params.getString("merchantId")
+  val merchantName = params.getString("merchantName")
+  val allowedCardBrands = params.getIntegerList("allowedCardBrands")
+
+  return if (allowedCardBrands == null) {
```
jeanregisser: Small edge case: `getIntegerList` returns an empty list (not null) for `allowedCardBrands: []`, so an explicitly empty array takes the `else` branch and produces a config with zero allowed brands - whereas omitting the field falls back to the SDK defaults.

Is that distinction intended, or should `[]` behave the same as leaving it out?

jeanregisser (commented):
Great 🚀

jeanregisser (approved):

Issue #2575, opened 2026-08-21
STPApplePayContext hangs after didAuthorizePayment on iOS 26 + React Native New Architecture — "Payment Not Completed"


owen-appicidea:
### Describe the bug

When a customer pays with Apple Pay via `@stripe/stripe-react-native`'s
`confirmPlatformPayPayment({ applePay: ... })`, authorization succeeds
(Face ID), the sheet dismisses to a spinner, and the payment never settles —
ending with Apple Pay's "Payment Not Completed". Reproducible on iOS 26.x
devices under the React Native New Architecture (Fabric) + Hermes.

Native instrumentation confirms that on iOS,
`confirmPlatformPayPayment({ applePay })` invokes `STPApplePayContext`.

### React Native implementation (failing path)

```ts
// @stripe/stripe-react-native 0.74.0 (bundled Stripe iOS SDK 26.6.0)
import { confirmPlatformPayPayment, PlatformPay } from '@stripe/stripe-react-native'

const result = await confirmPlatformPayPayment(paymentIntentClientSecret, {
  applePay: {
    merchantCapabilities: [
      PlatformPay.ApplePayMerchantCapability.Supports3DS,
      PlatformPay.ApplePayMerchantCapability.SupportsCredit,
      PlatformPay.ApplePayMerchantCapability.SupportsDebit,
    ],
    merchantCountryCode: 'HK',
    currencyCode: 'HKD',
    cartItems: [
      {
        label: 'Order',
        amount: (amount / 100).toString(),
        paymentType: PlatformPay.PaymentType.Immediate,
      },
    ],
  },
})
```

The promise returned by `confirmPlatformPayPayment` never resolves or
rejects after authorization. We confirmed this by wrapping the call in a
`Promise.race` with a 90 s timeout plus an `AppState` watchdog — the
timeout/rejection fires, but Stripe's completion never does.

### Root-cause evidence (instrumentation)

We instrumented `StripeApplePay/StripeApplePay/Source/ApplePayContext/
STPApplePayContext.swift` with `NSLog` + `NotificationCenter` posts and
forwarded them into Sentry. Every failing payment shows this exact sequence:

1. `presentApplePay completion fired`
2. `didAuthorizePayment fired`          ← Face ID authorization accepted
3. `_completePayment started`
4. … nothing after

Confirmed absent in every failing case:

- `PaymentMethod.create completion fired` — never logged
- `handleSuccess` / `handleFailure` — never called
[...]

PR #2577, opened 2026-08-24
Fix PaymentSheet startActivity crash after checkout re-open (Android)

### LINKED ISSUE #2563: startActivity called from non-Activity context on Android
**Describe the bug**
The payment sheet will occasionally get into a "stuck" state where it crashes whenever you try and open it. This is only on Android and only after the payment sheet has already been successfully opened and then closed. It's almost like it sometimes doesn't get clean up right or something. Once in this state, it will always error on open until you restart the app. We only see this on Android, not iOS.

**To Reproduce**
This is not a consistent bug, but every repro when we have gotten follows this pattern. 
1. Open checkout (which will initialize paymentSheet)
2. Open payment sheet
3. Close payment sheet and checkout
4. Re-open checkout
5. Open payment sheet
6. Observe payment sheet closes and logs error

```
// logged from com.stripe.android.paymentsheet.PaymentSheetActivity
startActivity called from non-Activity context; forcing Intent.FLAG_ACTIVITY_NEW_TASK for: Intent { cmp=com.rover.android/com.stripe.android.paymentsheet.PaymentSheetActivity (has extras) }
```
**Expected behavior**
Payment sheet should re-open

**Screenshots**
This is what it looks like when you try to open it the second time. It will behave like this until you close the app completely and restart it. In other words, every attempt will error until restart.

Image

 - Device: Samsung Galaxy F15 (14)
 - OS: Android
 - Version: 16

**Additional context**
React native version: 81
stripe-react-native version: 0.72.0
Stripe android: 23.13.1
Stripe ios: 26.4.1

ayushsingh82:
## Summary
Fixes #2563 — PaymentSheet gets stuck erroring on Android after close/reopen of checkout, with the log:
```
startActivity called from non-Activity context; forcing Intent.FLAG_ACTIVITY_NEW_TASK
```

`PaymentSheetManager` caches `paymentSheet`/`flowController` for its whole lifetime and only rebuilds when the field is `null`. Both are built via `.build(activity, signal)`, which registers `ActivityResultLauncher`s on that specific `Activity` instance. If the underlying `Activity` is recreated while the manager stays alive across a close/reopen of the checkout screen, the cached instance keeps pointing at launchers registered on a dead `Activity`, and presenting falls back to `startActivity` on a non-`Activity` context — matching the reporter's log exactly.

This PR tracks which `Activity` the sheet/controller was built against (`builtWithActivity`). In `configureMode`, if the current `Activity` differs from that, the cached instance is nulled out and rebuilt against the fresh `Activity` before presenting.

## Verification
- Compiles cleanly against the Android SDK (`android-36`) with no new imports.
- Full existing Android unit test suite passes: `./gradlew :stripe_stripe-react-native:testDebugUnitTest` — 346/346 tests, 0 failures.
- **Not yet covered by a dedicated regression test** — the existing `PaymentSheetManagerTest.kt` only exercises pure mapper functions and `handleFlowControllerConfigured`, nothing in the activity-rebuild path itself.
- **Not reproduced against a physical device** — I don't have the reporter's repro environment (Samsung Galaxy F15, Android 16). This is my best hypothesis from tracing `initPaymentSheet`/`initFlowController`/`onDestroy`/`presentWithTimeout`, not a confirmed root cause from a live repro.

Happy to add a Robolectric-based regression test for the activity-swap path, or adjust the approach, if a maintainer can confirm this matches the internal ticket's diagnosis.

## Test plan
- [ ] Maintainer/reporter confirms this resolves the repro in #2563
- [ ] Add regression test for activity-rebuild path (open to guidance on best approach here)

### DIFF
--- a/android/src/main/java/com/reactnativestripesdk/PaymentSheetManager.kt
+++ b/android/src/main/java/com/reactnativestripesdk/PaymentSheetManager.kt
@@ -87,6 +87,10 @@ class PaymentSheetManager(
   private var keepJsAwake: KeepJsAwakeTask? = null
   private var lastConfigureWasCustomFlow: Boolean? = null
 
+  // Activity `paymentSheet`/`flowController` registered their launchers on; if the current
+  // Activity is ever a different instance, those launchers are stale and must be rebuilt.
+  private var builtWithActivity: Activity? = null
+
   @SuppressLint("RestrictedApi")
   override fun onCreate() {
     configure(arguments, initPromise)
@@ -96,6 +100,7 @@ class PaymentSheetManager(
     super.onDestroy()
     flowController = null
     paymentSheet = null
+    builtWithActivity = null
   }
 
   fun configure(
@@ -211,9 +216,13 @@ class PaymentSheetManager(
     args: ReadableMap,
     promise: Promise,
   ) {
+    val currentActivity = context.currentActivity
+    val activityChanged = currentActivity != null && currentActivity !== builtWithActivity
+
     if (args.getBooleanOr("customFlow", false)) {
       lastConfigureWasCustomFlow = true
-      if (flowController == null) {
+      if (flowController == null || activityChanged) {
+        flowController = null
         initFlowController(args, promise)
       }
       configureFlowController(promise)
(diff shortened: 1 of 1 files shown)


ayushsingh82:
cc @wooj-stripe — saw you opened an internal ticket for this on #2563. This PR is my best-effort fix based on tracing `PaymentSheetManager`'s activity-caching logic (details in the PR description), but it's not yet confirmed against a physical repro or your internal ticket's diagnosis. Would appreciate a look whenever you have bandwidth, and happy to adjust if it doesn't match what you're seeing internally.

ayushsingh82:
Following up on this — still reproducible, and it maps to #2575. PR is mergeable; happy to rebase or add a test if that helps triage.

PR #2578, opened 2026-08-24
Add Checkout Session JavaScript APIs

porter-stripe:
## Summary

Adds the React Native API surface for Checkout Sessions from the API review, including controller creation, the useCheckout hook, Checkout Session models, and the Checkout-owned Payment Element. The JavaScript implementations are stubs so the native bridges can land separately. Marks the full surface with `@CheckoutSessionPrivatePreview`.

API Review: https://docs.google.com/document/d/1mi1xmUzONmuX0tUwjDQ1PUBZZNWgaEuWTje2V1maXuE/edit?tab=t.0

## Motivation

## Testing

- [ ] I tested this manually
- [x] I added automated tests
- Full Jest suite, TypeScript, ESLint, Prettier, API Extractor, and package type checks

## Documentation


etc/stripe-react-native.api.md
gbirch-stripe: Can we add the docstrings?

src/types/Checkout.ts
```diff
@@ -0,0 +1,638 @@
...
+
+  /**
+   * Parameters for updating or clearing the customer's shipping address.
+   * @CheckoutSessionPrivatePreview
+   */
+  export interface UpdateShippingAddressParams {
+    /** The recipient's name. */
+    name?: string | null;
+    /** The shipping address. Pass `null` to clear it. */
+    address: Address | null;
```
gbirch-stripe, line 170: At least on iOS we don't accept a nil address to the `updateShippingAddress` call.

src/types/Checkout.ts
```diff
@@ -0,0 +1,638 @@
...
+   */
+  export interface PaymentElementConfiguration {
+    /** Controls how Payment Element asks customers to save payment methods. */
+    savePaymentMethodOptInBehavior?: SavePaymentMethodOptInBehavior;
+    /** Customizes the appearance of Payment Element. */
+    appearance?: PaymentElementAppearance;
+    /** Preferred networks for co-branded cards. */
+    preferredNetworks?: CardBrand[];
+    /** Controls how billing details are collected during checkout. */
+    billingDetailsCollectionConfiguration?: BillingDetailsCollectionConfiguration;
```
gbirch-stripe, line 353: I think this type includes email, which we don't support in CS?

src/types/Checkout.ts
```diff
@@ -0,0 +1,638 @@
...
+   */
+  export interface Session {
+    /** The Checkout Session ID. */
+    id: string;
+    /** The business name configured in Stripe Business Public Details. */
+    businessName?: string;
+    /** Whether the Checkout Session is in live mode. */
+    livemode: boolean;
+    /** Three-letter ISO 4217 currency code in lowercase. */
+    currency: string;
```
gbirch-stripe, line 390: There won't be a currency for setup, and `currency` is an optional on iOS

porter-stripe: We will always have a currency! https://docs.google.com/document/d/1mi1xmUzONmuX0tUwjDQ1PUBZZNWgaEuWTje2V1maXuE/edit?disco=AAAB_Xsaw18

src/types/Checkout.ts
porter-stripe: API review has this as optional: https://docs.google.com/document/d/1mi1xmUzONmuX0tUwjDQ1PUBZZNWgaEuWTje2V1maXuE/edit?tab=t.0

See change log on Aug. 10

src/types/Checkout.ts
porter-stripe: Oh yeah you're right, we should re-define this.

gbirch-stripe (approved):

PR #2579, opened 2026-08-24
Add Checkout controller bridge architecture

porter-stripe:
## Summary

Adds the bridge foundation for Checkout Elements controllers. UI-thread-confined native registries support multiple controllers and clean them up when the React Native module is invalidated. A controller-scoped event router buffers updates during creation and prevents delayed or duplicate events from rolling JavaScript state backward.

## Motivation

## Testing

- [ ] I tested this manually
- [x] I added automated tests
- JavaScript, iOS, and Android unit tests

## Documentation

Select one:


ios/Checkout/CheckoutControllerRegistry.swift
```diff
@@ -0,0 +1,76 @@
...
+/// Element and presenter, observation tokens, and pending operations.
+@MainActor
+protocol CheckoutControllerInstance: AnyObject {
+    /// Releases all resources owned by this instance. Must be idempotent and run on the main thread.
+    func destroy()
+}
+
+/// Stores Checkout controllers behind opaque identifiers used by JavaScript.
+@MainActor
+final class CheckoutControllerRegistry {
```
gbirch-stripe, line 15: Are we intentionally supporting multiple `CheckoutController` instances?

porter-stripe: Yeah, based on our discussions in the meeting today, it may be recommended to create a new one based on the screen. If we don't register them it can get wonky what delegates are called and such. This ensures the right events/API calls get routed to the right instance. Also iOS and Android support this so I don't see why we shouldn't do it for RN.

gbirch-stripe (approved):

PR #2583, opened 2026-08-25
Map Checkout configuration

porter-stripe:
## Summary
- Adds types that can decode the JS bridge type into native configuration objects
- Adds unit tests as well.

## Motivation

## Testing
- [ ] I tested this manually
- [x] I added automated tests

Focused iOS and Android mapping tests and Checkout JS tests pass.

## Documentation


ios/Checkout/CheckoutConfigurationMapper.swift
```diff
@@ -0,0 +1,311 @@
...
+        let paymentElementParams = params["paymentElement"] as? NSDictionary
+        configuration.paymentElement = try mapPaymentElement(
+            paymentElementParams,
+            didSelectPaymentOption: didSelectPaymentOption
+        )
+
+        let applePayParams = paymentElementParams?["applePay"] as? NSDictionary
+        if let applePayParams {
+            configuration.applePayConfiguration = Checkout.ApplePayConfiguration(
+                merchantId: merchantIdentifier ?? "",
```
gbirch-stripe: On native iOS merchantId is non-optional, but here we're treating it as optional and passing an empty string to the native SDK, which doesn't have validation but will fail in this scenario

porter-stripe: Good catch, agreed. Updated!

gbirch-stripe (approved):

PR #2590, opened 2026-09-03
[SPM] Declare the Stripe SPM package

gbirch-stripe:
## Summary
- create scaffolding of full SPM solution
- declare the iOS dependency via React Native's `spm_dependency`
- infrastructure to use non-release branches

## Motivation
SPM resolution support in stripe-react-native

## Testing

- [x] I tested this manually
- [x] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/stripe_spm.rb
+++ b/stripe_spm.rb
@@ -73,13 +73,6 @@
 #     there also means the phase is appended after CocoaPods' own `[CP]`
 #     phases regardless of whether the app project is fresh (first install,
 #     Expo prebuild --clean) or already integrated.
-#   - On very old CocoaPods versions without post_integrate hooks (< 1.10),
-#     the user-project stage falls back to the end of the post_install stage.
-#     That works too — the analyzer, our hook, and the integrator all share
-#     one in-memory instance of the user project, so mutations made before
-#     integration survive it — the phase just ends up ordered before the
-#     `[CP]` phases on a fresh project.
-#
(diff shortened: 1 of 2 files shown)


stripe_spm.rb
```diff
@@ -0,0 +1,186 @@
...
+#     A mutation made any later is silently lost (CocoaPods does not save
+#     the Pods project again after integration).
+#   - The embed phase touches the user's project, and runs in the
+#     post_integrate stage. CocoaPods documents post_integrate as existing
+#     precisely so that hooks "can alter [the user project] after it is
+#     written to the disk", which our helpers do. Running
+#     there also means the phase is appended after CocoaPods' own `[CP]`
+#     phases regardless of whether the app project is fresh (first install,
+#     Expo prebuild --clean) or already integrated.
+#   - On very old CocoaPods versions without post_integrate hooks (< 1.10),
```
davidme-stripe: Do we want to do extra work to support these versions? That's 6+ years old. I'm unclear on whether this is extra work for us, it's tricky to parse.

gbirch-stripe: It looks like those old CocoaPods versions actually shouldn't even be possible to use on React Native 0.75+. I removed the branching for CocoaPods versions and instead replaced them with an assertion that the CocoaPods version is high enough (just in case)

davidme-stripe (approved):

PR #2592, opened 2026-09-03
[SPM] Guard Pods project UUID counter

gbirch-stripe:
## Summary
Guard against the UUID collision bug

## Motivation
There is a bug that causes a corrupted `Pods.xcodeproj`. This bug is fixed in later React Native versions but we need to provide it ourselves in earlier ones

## Testing

- [ ] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/stripe_spm.rb
+++ b/stripe_spm.rb
@@ -257,6 +257,79 @@ module StripeSPM
       end
     end
 
+    # Guards against a UUID-collision bug that corrupts Pods.xcodeproj when a
+    # post_install hook creates new project objects. This is provided by
+    # React Native on later React Native versions but we need to provide it
+    # ourselves for earlier ones.
+    #
+    # Background, verified against CocoaPods 1.16.2: `Pod::Project` overrides
+    # `generate_available_uuid_list` to mint deterministic UUIDs — a 6-char
+    # project prefix, a 7-hex-digit counter, and a trailing '0' — with, per
+    # its own comment, *no collision check* against existing objects, "as the
+    # Pods project is regenerated each time, and thus all UUIDs will have
+    # come from this method". Whenever that freshly-generated assumption
+    # breaks (an incremental-installation project loaded from cache, UUID
+    # postprocessing that sweeps the counters), the counter restarts near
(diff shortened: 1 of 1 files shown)


stripe_spm.rb
```diff
@@ -257,6 +257,79 @@ def apply_user_project(installer)
...
     end
 
+    # Guards against a UUID-collision bug that corrupts Pods.xcodeproj when a
+    # post_install hook creates new project objects. This is provided by
+    # React Native on later React Native versions but we need to provide it
+    # ourselves for earlier ones.
+    #
+    # Background, verified against CocoaPods 1.16.2: `Pod::Project` overrides
+    # `generate_available_uuid_list` to mint deterministic UUIDs — a 6-char
+    # project prefix, a 7-hex-digit counter, and a trailing '0' — with, per
```
davidme-stripe, line 321: What the heck

stripe_spm.rb
```diff
@@ -476,7 +549,21 @@ def stripe_spm_activate!(spec, version:)
       alias_method :stripe_spm_original_run_podfile_post_install_hooks, :run_podfile_post_install_hooks
 
       define_method(:run_podfile_post_install_hooks) do
-        # Run the regular hooks first: react_native_post_install (called from
+        # The UUID guard must run before the regular hooks: it is defending
+        # against object creation *inside* react_native_post_install.
+        # We fail softly only since a future CocoaPods release could change
+        # the behavior.
+        begin
```
davidme-stripe, line 623: Maybe we should turn this off if StripeDisableSPM is on? I'm a little worried about yet another patch, though I do see the issue...

gbirch-stripe: Done! Agreed that it's not worth risking breaking the fallback dependency resolution. The fallback flag should truly revert to the old behavior IMO

stripe_spm.rb
gbirch-stripe: I know it's so crazy!

davidme-stripe (dismissed):

tjclawson-stripe (approved):

PR #2594, opened 2026-09-03
[SPM] Prepare the example for SPM builds

gbirch-stripe:
## Summary
Add infrastructure to the example iOS app to support SPM resolution:
- conditionally use prebulit RN core
- conditionally fix DevSupport dependencies

## Motivation
- old arch dynamic linking doesn't work with prebuilt RN, so we need to disable it in this case
- the DevSupport pod under-declares its dependencies and so we have to manually correct it

## Testing

- [ ] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/example/ios/Podfile
+++ b/example/ios/Podfile
@@ -6,16 +6,82 @@ require "#{ws_dir}/node_modules/react-native-test-app/test_app.rb"
 
 workspace 'example.xcworkspace'
 
-# Use prebuilt RN.
-ENV['RCT_USE_RN_DEP'] = '1'
-ENV['RCT_USE_PREBUILT_RNCORE'] = '1'
+# Keep the example app on the existing CocoaPods resolution path while the
+# SPM integration is staged. The activation change will make SPM the default
+# and retain CocoaPods as an explicit fallback.
+disable_spm = true
+$StripeDisableSPM = true
+
+# SPM resolution requires dynamic frameworks: the Stripe Swift package
+# products use automatic linkage, which only reaches the final app correctly
+# when the pod that consumes them is a dynamic framework (stripe_spm.rb
+# enforces this with an actionable error). The CocoaPods fallback keeps
+# React Native's default static libraries.
+use_frameworks! :linkage => :dynamic unless disable_spm
 
 new_arch_enabled = ENV['NEW_ARCH_ENABLED'].nil? ? true : ENV['NEW_ARCH_ENABLED'] == 'true'
+
+# The prebuilt core only wires framework linkage through the new-architecture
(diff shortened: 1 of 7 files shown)


example/ios/Podfile
```diff
@@ -6,16 +6,78 @@ require "#{ws_dir}/node_modules/react-native-test-app/test_app.rb"
...
+# TODO: turn off
+disable_spm = true
+$StripeDisableSPM = true
+
+# SPM resolution requires dynamic frameworks.
+use_frameworks! :linkage => :dynamic unless disable_spm
 
 new_arch_enabled = ENV['NEW_ARCH_ENABLED'].nil? ? true : ENV['NEW_ARCH_ENABLED'] == 'true'
 
+# The prebuilt core only wires framework linkage through the new-architecture
```
davidme-stripe, line 27: Not sure I understand this. I don't think we even still want to support the old architecture at this point, it's effectively deprecated afaik.

example/ios/Podfile
```diff
@@ -6,16 +6,78 @@ require "#{ws_dir}/node_modules/react-native-test-app/test_app.rb"
...
+
 options = {
   :bridgeless_enabled => new_arch_enabled,
   :fabric_enabled =>  new_arch_enabled,
   :hermes_enabled => true,
+  # react-native-test-app runs this lambda from its own post_install hook.
+  :post_install => lambda do |installer|
+    # Dynamic-frameworks builds (SPM mode) make every pod framework resolve
+    # its own symbols at link time, and react-native-test-app's DevSupport
+    # pod under-declares its dependencies for that. Which `-framework` flags
```
davidme-stripe: Is it possible to declare the dependencies properly somehow instead of patching them in? I'm confused as to why we'd need to add React/ReactNativeDependencies/hermes ourselves here just to get Stripe to work.

example/ios/Podfile
gbirch-stripe: Let's keep following up on this discussion via Slack, but tl;dr this is just for CI to cover old arch via the example, which we should have it continue to do until we properly drop support, at which point we can clean this up

example/ios/Podfile
gbirch-stripe: Let's keep following up about this on Slack, but tl;dr the proper fix in RNTA's podfile. I added a patch to this PR that fixes their podfile. If we want to even more properly fix this we should open up a PR on RNTA

davidme-stripe (approved):

davidme-stripe (dismissed):

tjclawson-stripe (approved):

PR #2595, opened 2026-09-03
[SPM] Validate SPM dependency integration

gbirch-stripe:
## Summary
Validate this fresh Expo project scenario in CI.

## Motivation
This specific scenario is one that ran into problems during development, so we are testing to guard against it.

## Testing

- [ ] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/scripts/update-expo-plugins.js
+++ b/scripts/update-expo-plugins.js
@@ -6,9 +6,12 @@ const path = require('path');
 const args = process.argv.slice(2);
 const appJsonPath = path.resolve(args[args.indexOf('--app-json') + 1]);
 const expoVersion = args[args.indexOf('--expo-version') + 1];
+// When set, opts the app out of Swift Package Manager resolution through the
+// plugin's disableSPM option, exercising the CocoaPods fallback path instead.
+const disableSPM = args.includes('--disable-spm');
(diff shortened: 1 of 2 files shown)


scripts/update-expo-plugins.js
```diff
@@ -19,18 +22,30 @@ expoConfig.plugins.push([
...
+const buildProperties = {};
+if (needsKotlinVersion) {
+  buildProperties.android = {
+    kotlinVersion: '2.0.21',
+  };
+}
+if (!disableSPM) {
+  // Every Expo SDK this script targets ships React Native >= 0.75, so the
+  // Stripe iOS SDK resolves through Swift Package Manager by default — which
+  // requires dynamic frameworks (Expo's default is static; `pod install`
```
davidme-stripe, line 38: I see — if this script is only for our test app, and Expo's default is static linking, do we then need to force every Expo-using app to switch from static linking to dynamic?

gbirch-stripe: All apps using our SDK, both expo and non-expo, need to switch to dynamic linking to support BPM resolution.

davidme-stripe (dismissed):

tjclawson-stripe (approved):

PR #2608, opened 2026-09-06
[SPM] Define the SPM framework embedding script

gbirch-stripe:
## Summary
Add the script that embeds the Stripe frameworks in the user's project.

## Motivation
React Native's `spm_dependency` helper doesn't correctly guarantee that frameworks with are embedded, causing runtime crashes.

## Testing

- [ ] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/stripe_spm.rb
+++ b/stripe_spm.rb
@@ -131,6 +131,64 @@ module StripeSPM
   ONRAMP_PRODUCT = 'StripeCryptoOnramp'.freeze
   ONRAMP_SUBSPEC = "#{POD_NAME}/Onramp".freeze
 
+  # Shown in Xcode's build-phases UI; also the key used to find/replace
+  # the phase on later installs.
+  EMBED_PHASE_NAME = '[stripe-react-native] Embed SPM Frameworks'.freeze
+
+  # Embeds SPM-built dynamic frameworks into the app bundle.
+  #
+  # Why this is needed: stripe-ios's package products are "automatic" linkage
+  # libraries, and in some configurations Xcode chooses to build them as real
+  # dynamic frameworks (under BUILT_PRODUCTS_DIR/PackageFrameworks for regular
+  # builds, and under OBJROOT/UninstalledProducts/<platform> for Archive
+  # builds, which never populate PackageFrameworks). Xcode only auto-embeds
+  # package frameworks for targets that link them *directly*; frameworks
+  # linked by a CocoaPods pod target are invisible to both Xcode's embedding
+  # and CocoaPods' "[CP] Embed Pods Frameworks" phase. Without this script the
+  # app builds fine and then crashes at launch with
+  # "dyld: Library not loaded: @rpath/Stripe....framework".
+  #
+  # Script details:
+  #   - Filters to Stripe*.framework so we never touch others.
+  #   - Uses file(1) to skip statically linked frameworks: those are already
+  #     linked into their consumers, and embedding a static framework in the
+  #     bundle can fail App Store validation.
+  #   - Strips Headers/PrivateHeaders/Modules, which don't belong in a shipped
+  #     app bundle.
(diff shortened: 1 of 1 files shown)


stripe_spm.rb
```diff
@@ -131,6 +131,64 @@ module StripeSPM
...
+      for FRAMEWORK in "$SEARCH_DIR"/Stripe*.framework; do
+        [ -d "$FRAMEWORK" ] || continue
+        NAME="$(basename "$FRAMEWORK" .framework)"
+        BINARY="$FRAMEWORK/$NAME"
+        [ -f "$BINARY" ] || continue
+        file -b "$BINARY" | grep -q "dynamically linked" || continue
+        if [ ! -d "$DEST/$NAME.framework" ]; then
+          rsync -a --exclude Headers --exclude PrivateHeaders --exclude Modules "$FRAMEWORK" "$DEST/"
+          if [ -n "${EXPANDED_CODE_SIGN_IDENTITY:-}" ] && [ "${CODE_SIGNING_ALLOWED:-NO}" = "YES" ]; then
+            codesign --force --sign "$EXPANDED_CODE_SIGN_IDENTITY" --preserve-metadata=identifier,entitlements "$DEST/$NAME.framework"
```
davidme-stripe, line 178: This is pretty spooky, and it's kind of hard to review because it's separated from the PR that actually calls the `EMBED_SCRIPT` script.

I'm a little concerned that this might be brittle, even though we're checking the codesign environment variables: The user may not want these exact codesign options happening, they may handle their codesigning elsewhere, etc.

Is there a paved road for embedding these SPM frameworks without these hacks? I sort of thought this would be a solved problem for all React Native packages...

gbirch-stripe: I dug into this further, and unfortunately I think this is still the best path. The good news is that this script only does something if Xcode builds Stripe products as dynamic frameworks. Ours are set to "automatic", and my research indicates that we can actually expect them to usually get built as static, in which case this script doesn't do anything. This xcodebuild behavior isn't guaranteed, though, and so I think we should still keep this script for full coverage.
As for code signing, this is actually the same thing that CocoaPods does during its `Embed Pods Frameworks` build phase. For users who handle code signing elsewhere, we respect `CODE_SIGNING_ALLOWED=NO` and the user can still re-sign later.
Unfortunately I'm not able to find a more "paved road" approach.

davidme-stripe: Okay, if Cocoapods is doing it, then I guess it's fine considering the circumstances. 😬

davidme-stripe (dismissed):

stripe_spm.rb
```diff
@@ -124,6 +124,64 @@ module StripeSPM
...
+    mkdir -p "$DEST"
+    for SEARCH_DIR in "${BUILT_PRODUCTS_DIR}/PackageFrameworks" "${OBJROOT}/UninstalledProducts/${PLATFORM_NAME}"; do
+      [ -d "$SEARCH_DIR" ] || continue
+      for FRAMEWORK in "$SEARCH_DIR"/Stripe*.framework; do
+        [ -d "$FRAMEWORK" ] || continue
+        NAME="$(basename "$FRAMEWORK" .framework)"
+        BINARY="$FRAMEWORK/$NAME"
+        [ -f "$BINARY" ] || continue
+        file -b "$BINARY" | grep -q "dynamically linked" || continue
+        if [ ! -d "$DEST/$NAME.framework" ]; then
```
davidme-stripe: What would this do if the file already exists? If something changes in the Stripe pod (like someone updates to a newer version), would this still copy the new framework? Or would it be stuck with an outdated version?

gbirch-stripe: Made this unconditional. The time cost is ~1 second and there's no reliable safe way I could find to make it conditional

tjclawson-stripe (approved):

PR #2610, opened 2026-09-06
[SPM] Define SPM integration workflows

gbirch-stripe:
## Summary
1. add an old arch variant to the unit tests job that only runs in the test-project pipeline
2. add a build-only job for the CocoaPods fallback
3. add test for fresh Expo projects
4. limit E2E build to arm64 architecture only

## Motivation
Referencing the numbered items above:
1. This scenario now has different behavior with SPM resolution, and so we should add coverage of it, but not on every PR since it is expensive.
2. Ensure the health of the CocoaPods fallback
3. This was a scenario that had bugs during development, so we should guard against them
4. This is the only architecture actually needed for tests, and limiting CI to this reduces CI times. I manually verified that CI still runs find without this change (it's just much faster with it)

## Testing

- [ ] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -137,14 +137,10 @@ workflows:
   # via e2e-build-ios-old-arch / e2e-tests-ios-old-arch.
   #
   # Unlike unit-test-ios, this job embeds a prebuilt JS bundle
-  # (_build_js_bundle, as the e2e jobs do). CI has no Metro server, and with
-  # neither Metro nor an embedded bundle the Debug test app's old-arch launch
-  # path never becomes ready for XCTest — the run fails with "test runner
-  # hung before establishing connection" after the app's failed
-  # http://localhost:8081/status probe. react-native-test-app prefers Metro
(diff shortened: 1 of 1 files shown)


bitrise.yml
```diff
@@ -120,6 +120,60 @@ workflows:
           inputs:
             - notify_user_groups: none
 
+  # Old-architecture variant of unit-test-ios: the same native unit tests
```
davidme-stripe, line 123: I think we can drop old architecture at this point. Recent RN versions don't support it.

gbirch-stripe: See comment on other PR + Slack discussion, but IMO we should keep old arch in CI for now and then fully drop it in 1.0, at which point we can just delete all of this CI stuff to support it.

davidme-stripe (approved):

porter-stripe (approved):

PR #2628, opened 2026-09-10
StripeCryptoOnramp: OnrampPresenter Activity Recreation Fix

Twigz:
## Summary

Recreates the Onramp presenter when its host Activity changes and runs presentation calls on the main thread.
While this was focused on fixing this for Authentication (where the bug was reported), it could theoretically affect anywhere the presenter was being used.

## Motivation

The RN module can survive Activity recreation while retaining a presenter tied to the destroyed Activity. Calling authorize then attempts to use its unregistered NativeLinkActivityContract launcher.

Tracks the presenter's owning Activity, clears it on destruction, and recreates it for the current host while preserving the existing coordinator.

## Testing

- [x] I tested this manually
- [ ] I added automated tests

Tested on a Pixel 7a by recreating the Activity after configuring Onramp, then calling authorize. Confirmed the module survived while the Activity changed. Reproduced the unregistered-launcher exception on master and verified authorization works with this fix.

## Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## Unreleased
+
+**Fixes**
+
+* [Fixed] Android: Fixed Crypto Onramp presentation failures after Activity recreation by recreating the presenter for the current Activity.
+
 ## 0.76.0 - 2026-09-01
 **Features**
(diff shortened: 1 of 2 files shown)


jeanregisser (approved):
🚀

PR #2632, opened 2026-09-11
[Old arch X] Remove Android old arch support

gbirch-stripe:
## Summary
- Removes all of the old arch Android sources
- Enforces new arch in build.gradle

## Motivation
Remove old arch support

## Testing

- [ ] I tested this manually
- [ ] I added automated tests

## Documentation

Select one: 


android/build.gradle
```diff
@@ -36,12 +36,8 @@ def getExtOrIntegerDefault(name) {
   return rootProject.ext.has(name) ? rootProject.ext.get(name) : (project.properties["StripeSdk_" + name]).toInteger()
 }
 
-def isNewArchitectureEnabled() {
-  // To opt-in for the New Architecture, you can either:
-  // - Set `newArchEnabled` to true inside the `gradle.properties` file
-  // - Invoke gradle with `-newArchEnabled=true`
-  // - Set an environment variable `ORG_GRADLE_PROJECT_newArchEnabled=true`
-  return project.hasProperty("newArchEnabled") && project.newArchEnabled == "true"
+if (project.findProperty("newArchEnabled")?.toString() == "false") {
```
davidme-stripe: Is this still needed? I thought it was always on by default now and the setting was removed. Maybe that's just for Expo 🤔

gbirch-stripe: New arch is on by default starting in 0.76 and the only option starting in 0.82, so it's definitely still possible for users to have it on. Once the SPM stuff goes through we'll have an RN 0.75 minimum (right now we don't have one), but either way there's no guarantee that merchants have new arch.

davidme-stripe: I dug in a bit more, and I think this was entirely removed in 0.82. Here's 0.81: https://github.com/react/react-native/blob/v0.81.5/packages/gradle-plugin/react-native-gradle-plugin/src/main/kotlin/com/facebook/react/utils/ProjectUtils.kt#L31 vs 0.82: https://github.com/react/react-native/blob/v0.82.0/packages/gradle-plugin/react-native-gradle-plugin/src/main/kotlin/com/facebook/react/utils/ProjectUtils.kt#L32

So they might not have it defined in their gradle.properties, and we'd incorrectly block them. Though if the property is missing (instead of explicitly `false`) here, it would actually resolve to `"null"`, which would not equal `"false"`, so they wouldn't hit this exception anyway?

gbirch-stripe: Did some more research. In 0.82+ this check in its original form did work correctly because RN forces it to be true ([source](https://github.com/react/react-native/blob/v0.82.0/packages/gradle-plugin/react-native-gradle-plugin/src/main/kotlin/com/facebook/react/ReactRootProjectPlugin.kt#L21-L42)). I updated the logic though to be better aligned with RN's own, though, to reduce edge cases

davidme-stripe (approved):

PR #2642, opened 2026-09-14
Bridge iOS Checkout controller lifecycle

porter-stripe:
## Summary

Creates and destroys iOS Checkout controllers and forwards native session and payment selection events. The native layer keeps controllers by ID and cancels pending creation on destroy.

## Motivation

Checkout Sessions

## Testing

- [ ] I tested this manually
- [x] I added automated tests

## Documentation


ios/Checkout/CheckoutControllerRegistry.swift
```diff
@@ -1,76 +0,0 @@
-import Foundation
```
porter-stripe, line 1: From an old implementation and being replaced.

tjclawson-stripe (approved):

